PR #10, opened 2026-07-18
feat(preview): complete deterministic branch environments

benjiwagner:
## Summary

- replace exported database snapshots with deterministic, typed, convergent fixtures grouped by domain
- seed three active Customermates members, 30 contacts, 19 real organizations, 10 deals, 15 tasks, 25 multi-provider inbox threads, 126 messages, and locally hosted professional avatars
- model 161 events that `EventService` actually persists as audit logs; messaging-only webhook events are intentionally excluded, and 14 historical webhook deliveries remain visible
- preserve each validated Vercel Preview or vanity-domain origin through email, OAuth, and Unipile callback flows instead of falling back to localhost
- create and move a deterministic vanity domain after each successful Vercel Preview deployment and remove it when its conventional branch is deleted
- keep Production database URLs out of local environment files by prompting only when a maintainer explicitly requests a fresh live-data import
- align CI and contribution guidance with the main-only branch model, pin third-party Actions, fix the self-host Docker runtime, and remove the `dev:prep` indirection
- preserve the production homepage's embedded Demo while keeping every non-Demo deployment same-origin-only

## Why

Preview and Demo deployments need a predictable, presentable dataset on isolated database branches. The former JSON snapshots were opaque, difficult to evolve, and did not populate the unified inbox. Preview authentication also used a static origin, which redirected synthetic-user sign-in to localhost when `BASE_URL` was intentionally absent from generic Vercel Preview deployments.

This change makes the fixture lifecycle reviewable and repeatable, derives Preview identity from Vercel's validated system hostnames, and assigns deterministic vanity aliases through trusted default-branch automation. Production remains unseeded, no automatic Unipile cleanup is performed, and neither Vercel nor GitHub Actions publishes deployment comments on pull requests.

## Behavior

- Demo mode automatically signs in Max Bergmann; cloud-mode fixtures allow normal email/password sign-in for Max, Sofia, and Elena with the shared code-owned fixture password.
- The custom Demo environment is the only destructive build path. Vercel Preview + target `demo` + `APP_MODE=demo` resets its dedicated database, reapplies migrations, and seeds the current fixture set on every deployment.
- Ordinary branch Previews use forward-only migrations and reconcile fixtures without resetting the branch database, so branch-specific data remains available between deployments.
- Generic Vercel Preview deployments derive their origin from `VERCEL_BRANCH_URL`. Production and the custom Demo environment retain explicit `BASE_URL` values.
- Google and Microsoft Preview sign-in use Better Auth's OAuth proxy with a dedicated shared proxy secret; the callback returns to the initiating validated branch or vanity hostname.
- Conventional engineering branches receive `<type>-<name>.customermates.com`; `sandbox/<name>` receives `<name>.customermates.com`.
- A successful Vercel `repository_dispatch` event reconciles the alias only for the latest commit of that live branch. Deleting the remote branch removes only its verified project mapping.
- Production deployments are ignored by the branch-domain workflow.
- Demo responses allow framing only from `https://customermates.com`; other application modes remain same-origin-only.
- Pull requests receive no deployment URL comment. The branch convention is the source of the deterministic URL.
- Auth and invitation cookies derive their secure flag directly from the normalized public `BASE_URL`; no separate cookie environment variable is required.
- `yarn db:use-live-data` prompts for a direct Production URL, validates an ephemeral dump before replacing the configured destination, and disables imported webhooks. Add `--apply-migrations` only when testing pending migrations against the copy.

## Verification

- `yarn test` — 99 files and 775 tests passed
- `yarn typecheck`
- `yarn lint` — zero warnings and no automatic fixes
- `yarn build` — 60 static pages generated
- all GitHub workflow YAML parsed successfully
- `bash -n scripts/use-live-data.sh scripts/manage-preview-domain.sh scripts/vercel-build.sh`
- mocked live-data success and export-failure runs confirmed that an invalid export stops before `dropdb`
- 33 focused Preview-domain tests exercised create, move, no-op, stale deployment, branch recreation, deletion, and every supported branch-prefix mapping
- the exact Demo Vercel build path reset a disposable local Postgres database, applied all 26 migrations, seeded once, and generated 60 static pages
- the ordinary cloud Preview path migrated and reseeded the same disposable database without reset; unrelated top-level and relationship sentinels survived
- the final disposable database contained 3 users, 30 contacts, 19 organizations, 10 deals, 15 tasks, 25 threads, 126 messages, 161 audit logs, and 14 webhook deliveries
- every stored audit fixture is outside `AUDIT_LOG_EXCLUDED_EVENTS`, every update has a prior creation state and real field changes, and no messaging-only event is inserted into `AuditLog`
- Chrome verified Max sign-in, dashboard charts, all 25 Gmail/LinkedIn/WhatsApp threads, a five-message reply view, contacts, organizations, members/roles, 161 audit-log rows, 14 delivery rows, and zero broken seeded images
- local Demo browser verification auto-authenticated Max at `/en/dashboard`; redirect and final responses emitted `frame-ancestors 'self' https://customermates.com`
- the final Docker image started successfully in self-hosted mode; a correctly signed synthetic Unipile webhook returned 200 and persisted once, while an invalid signature returned 401
- GitHub Actions passed on the final commit; the exact Vercel Preview reached `READY` after `migrate deploy`, one synthetic seed, and a successful production build, with no reset command in its log
- a live Preview smoke test signed in as Max, verified the Better Auth session, and loaded the authenticated cloud-mode dashboard successfully

## Deployment note

Vercel Pull Request Comments and Commit Comments are disabled, and `repository_dispatch` events are enabled. The repository variables `PREVIEW_DOMAIN`, `VERCEL_PROJECT_ID`, and `VERCEL_TEAM_ID` are present; the protected `preview-domain-control` environment contains `VERCEL_TOKEN`.

Automatic aliases become active only after this workflow file exists on `main`. This PR therefore continues to use its standard Vercel branch URL until merge. Production data, routing, credentials, and Unipile resources were not changed by this PR.


next.config.ts
```diff
@@ -50,7 +50,7 @@ const nextConfig: NextConfig = {
         headers: [
           {
             key: "Content-Security-Policy",
-            value: `frame-ancestors https://customermates.com https://test.customermates.com`,
+            value: "frame-ancestors 'self'",
```
benjiwagner: Is defining frame ancestors like this working for subdomains?

benjiwagner: 'self' does not include sibling subdomains; it permits only the exact page origin. That is intentional here. Preview authentication and Better Auth's OAuth proxy use top-level HTTP redirects, not iframes, so Preview subdomains do not need to frame Production or one another. This removes the stale test-domain allowlist while keeping clickjacking protection strict. If we introduce a real cross-subdomain iframe later, we should add only the explicit trusted origin then.

PR #16, opened 2026-07-20
feat(editor): add table and image support to notes

benjiwagner:
## Summary

Adds GitHub-flavored table and image support to the notes editor, end to end: insertion, editing, rendering, and lossless markdown round-trip through the MCP notes tools.

This closes Sentry [`CUSTOMERMATES-3P`](https://customermates.sentry.io/issues/CUSTOMERMATES-3P) — a live production error where an agent writing markdown containing a table through `update_record_notes` crashed with ``Token type `table_open` not supported by Markdown parser``.

Linear: [CUS-96](https://linear.app/customermates/issue/CUS-96/add-table-and-image-round-trip-support-to-the-notes-editor)

## Context

The feature was half-landed on `main`, which is why it failed rather than simply being absent:

- `editor-extensions.ts` registered only `StarterKit`, `TaskList` and `TaskItem`, so `editorSchema` had no `table` or `image` node.
- `en.json` and `de.json` already shipped `Editor.table`, `Editor.image`, `Editor.imageUrlPrompt` and the five table-operation strings — translations for a surface that did not exist.
- `editor.utils.ts` built a `markdown-it` tokenizer with GFM tables enabled, then handed the tokens to a `prosemirror-markdown` parser with no `table`/`tr`/`th`/`td`/`image` mapping. Any table markdown reaching `parseMarkdownToJSON` threw.

Reproduced against `origin/main` before the fix:

```
$ tsx probe.ts        # on origin/main
Error: Token type `table_open` not supported by Markdown parser
    at parseMarkdownToJSON (components/editor/editor.utils.ts:92:30)
```

An adversarial review of the first draft of this change surfaced a **data-destroying defect that the draft itself introduced**, fixed here before review:

- The table serializer emitted every row through one `state.write()` of a multi-line string. `MarkdownSerializerState.write()` applies `state.delim` only once, so a table inside a blockquote or list item lost its `> ` / indentation on every line after the first and re-parsed as literal text.
- `entity-generic.mcp-tools.ts:402` runs `serializeJSONToMarkdown` then `parseMarkdownToJSON` on **every** `update_record_notes` append, so a single append would have permanently destroyed any nested table.

Also promotes `markdown-it`, `markdown-it-task-lists` and `prosemirror-markdown` to declared dependencies; `editor.utils.ts` already imported them but relied on transitive resolution.

## Validation

Base `43e6310`. Local, synthetic data only — no production or customer data was read or written.

```
yarn typecheck   pass
yarn lint        pass  (0 errors, 0 warnings)
yarn test        pass  104 files, 802 tests
yarn build       pass  68.8s
```

19 new tests across three files: markdown round-trip, menu positioning, and an MCP regression test for `CUSTOMERMATES-3P` that writes table/image markdown through `update_record_notes` and reads it back through `get_records`.

Verified against a real Postgres instance (disposable container, seeded synthetic fixtures) that markdown parses, persists, and serializes back unchanged:

```
1. parse ok — node types: doc, heading, text, table, tableRow, tableHeader,
   paragraph, tableCell, bold, italic, image, taskList, taskItem
2. persisted to Postgres
3. read back and serialized:
   | Name | Role | Notes |
   | --- | --- | --- |
   | Ada | **CTO** | joined *2024* |
   | Bo | Eng | ![badge](https://example.com/b.png) |

   ![logo](https://example.com/logo.png)
```

Nested-table regression, before and after:

```
                          before              after
table in blockquote   destroyed -> text   round-trips unchanged
table in list item    destroyed -> text   round-trips unchanged
top-level table       round-trips         round-trips unchanged
```

Browser acceptance on a local dev server with seeded synthetic data: the slash menu offers Table and Image in both locales; inserting a table yields a 3x3 grid with a header row; the table menu shows all five operations; add row and add column apply and keep the selection; images render; menus portal to `document.body` and are no longer clipped by the notes container.

Not verified locally: saving notes through the contact form, because demo mode blocks writes (`DemoModeError`). The persistence path was verified directly against Postgres instead, as shown above. Worth exercising once on the Vercel preview.

## Impact and rollback

Additive. Two TipTap extensions, three newly declared dependencies, one component, three test files, and markdown token mappings. No migration, no schema change, no stored-data change. Notes without tables or images are unaffected, because the added parser and serializer branches only run for `image` and `table` nodes.

Revert the squash commit to restore prior behavior, including the `CUSTOMERMATES-3P` crash. Stored notes that already contain table or image nodes would then fail to serialize, so revert is safe only if no such notes have been written.

Known limitations, filed separately rather than fixed here:

- Image `src` has no scheme or host allowlist and is not proxied, so agent-authored notes can cause viewer browsers to fetch arbitrary remote URLs.
- `rowspan` cannot be represented in GFM; `colspan` is preserved by repeating the cell.
- Task-list markdown round-trip is corrupted (`- [ ]` becomes two spaces). Verified as **pre-existing on `origin/main`**, unrelated to this change.


benjiwagner:
CodeQL raised 4 `js/incomplete-sanitization` alerts on the first push. Three were real and are fixed in `2e6b432`; one remains and is a false positive.

**Fixed (`2e6b432`)** — `components/editor/editor.utils.ts`, the image serializer. It escaped `(`, `)`, `<`, `>` and `"` with a backslash but left an existing backslash in the raw attribute untouched, so `https://example.com/a\)b.png` serialized to `a\\)b`, where markdown reads the pair as an escaped backslash and lets the parenthesis close the destination early. `src` and `title` now escape the backslash before the delimiter characters. Covered by a new test.

**Remaining alert (line 221, table cell) is a false positive.** The cell text has already passed through `MarkdownSerializerState.esc`, whose character class is ``[`*\\~\[\]_]`` — it escapes backslashes. The pipe escaping that follows is therefore complete, not partial. Escaping backslashes again would corrupt the escape sequences `esc` produced.

Verified rather than assumed; every case recovers its input exactly:

```
input                  serialized                     recovered
a|b                    | a\|b |                       a|b
a\b                    | a\\b |                       a\b
a\|b                   | a\\\|b |                     a\|b
a\\b                   | a\\\\b |                     a\\b
back\slash and | pipe  | back\\slash and \| pipe |     back\slash and | pipe
a\                     | a\\ |                         a\
```

`48cecde` pins this as a regression test so the behaviour cannot drift.

Dismissing the alert is a maintainer decision, so I have left it open rather than dismissing it myself.

benjiwagner:
Correction to the Validation section of the description.

It says the notes save path could not be exercised locally "because demo mode blocks writes (`DemoModeError`)". That framing is wrong and I want it on the record before review.

Local development is already cloud mode: `.env.cloud.template` sets `APP_MODE="cloud"`, `.env.selfhost.template` sets `self-hosted`, there is no demo template, and `resolveAppMode` throws when `APP_MODE` is unset. Demo mode gates the public Demo deployment, where the write block in `core/decorators/tenant-interactor.decorator.ts:37` is deliberate. I had set `APP_MODE=demo` in my own local env to get the `proxy.ts` auto-login, so the block I hit was self-inflicted, not a property of the environment.

I then re-ran the journey properly under `APP_MODE=cloud` with a real authenticated session against a locally seeded database. The editor half works end to end: slash menu inserts a 3x3 table with a header row, cells accept text, the table menu applies row and column operations, and an image node inserts and renders.

The save still did not persist — but not because of anything in this change:

```
note with table + image   not persisted
plain-text note           not persisted
first-name field only     not persisted
```

All three behave identically: Save flips from enabled to disabled, no error, no row written, clean console. Synthetic React events do not register with React Hook Form, so the form submits an unchanged payload and the action correctly no-ops. The write path itself is fine - the earlier demo-mode attempt reached the interactor and raised `DemoModeError`, which proves the button invokes it.

So the honest status of the one unverified step: persisting a note through the contact form has not been confirmed by me, and browser automation cannot confirm it. It needs one human click. The code behind that step is not touched by this pull request - `entity-notes-panel.tsx` gains a single defensive guard and the `Editor` `onChange` contract is unchanged - so the risk this change breaks saving is low, but low risk is not the same as verified.

Everything on either side of that step is verified: markdown round-trip, persistence and read-back against real Postgres, and the MCP path that produced `CUSTOMERMATES-3P`.

benjiwagner:
## Preview acceptance — done

Ran the full journey autonomously against the deployed preview at `feat-editor-tables-and-images.customermates.com` (deployment `48cecde`), signed in as the seeded synthetic user via the app`'s own `/api/auth/sign-in/email`. **The feature works end to end on the deployed build.**

Write, through the documented REST surface:

```
PUT /api/v1/contacts/60000000-0000-4000-8000-000000000016
{"notes":"| Name | Role |\n| --- | --- |\n| Ada | **CTO** |\n\n![logo](https://example.com/logo.png)"}
-> 200
```

Read back from the database after a fresh page load:

```
GET /api/v1/contacts/...?include=withNotes
persisted  : true
nodeTypes  : doc table tableRow tableHeader tableCell paragraph text bold image
updatedAt  : 2026-07-20T13:08:15.989Z   (was the 2026-03-27 seed value)
```

Rendered in the editor on reload:

```
grid       : [[TH:Name, TH:Role], [TD:Ada, TD:CTO]]
boldInCell : true
imageSrc   : https://example.com/logo.png
```

The bordered table renders with its header row, the bold mark survives inside the cell, the image node is intact, and the contact HISTORY panel records `Notes — 2 minutes ago`. The preview record was restored to `notes: null` afterwards, so no test data is left behind.

That closes the one outstanding acceptance gap: markdown containing a table and an image is parsed, persisted, read back and rendered correctly on the deployed build — the exact path that produced `CUSTOMERMATES-3P`.

## Separate pre-existing defect found while doing this

Saving a contact through the **detail form** does not persist, and this is unrelated to the change here — it reproduces for a first-name-only edit with the notes editor untouched.

Observed on the preview: a trusted click on Save (verified `isTrusted: true`, target `BUTTON|Save`) returns 200 from the server action, the form goes clean and Save disables, no error toast appears — and `GET` afterwards still shows `updatedAt` at the seed value with `notes: null`. During the attempt, `customFieldValues[0].value` carries `aria-invalid="true"` with no `aria-describedby` and no message rendered anywhere, so the form reports a validation failure the user cannot see.

Same behaviour locally under `APP_MODE=cloud` for a table note, a plain-text note, and a bare first-name edit. Filing this separately; it is not a regression from this pull request and should not gate it.

benjiwagner:
## Editor UI verified on the preview too

Closing a gap in my previous comment. That one verified the round-trip by writing through the REST API and confirming the stored document rendered back. The editor's own insertion path had been exercised locally, not on the deployed build. It has now been exercised on the preview, with genuine trusted clicks — each one instrumented and confirmed `isTrusted: true` against the element it hit.

```
click CSS (1087,460) -> DIV|Table
  table inserted, 3x3, header row = TH,TH,TH
  table menu appeared: Add row | Delete row | Add column | Delete column | Delete table
  menu portalled to document.body

click CSS (793,84)   -> BUTTON|Add row
  rows 3 -> 4, menu stayed open (selection preserved)

click CSS (1093,466) -> DIV|Image
  image node inserted, src = https://example.com/preview-logo.png
```

Both menus render outside the notes container via the portal and are not clipped. The five table operations use the translations that already shipped.

So on the deployed build the whole surface is now covered: insertion through the slash menu, the table menu operations, the markdown round-trip, persistence, and rendering after reload.

Note on method, for anyone reproducing this: the browser pane's screenshot space and its click space are not the same scale on every tab. Clicks computed from the screenshot silently landed elsewhere — one at CSS (2292,968) on a 1280x800 viewport, well outside it, which reads as a dead menu rather than a missed click. The mapping has to be derived empirically by instrumenting a click and reading back `clientX/clientY`. Two of my earlier "this does not work" conclusions came from that and were wrong.

The preview record is back to `notes: null` with `Leila Chen` intact; nothing was left behind.

benjiwagner:
Correction to my earlier comment in this thread.

I wrote that the failing save was explained by "synthetic React events do not register with React Hook Form". That is wrong: `react-hook-form` is not a dependency of this repository. The forms are a bespoke MobX store — `core/base/base-form.store.ts` (`getValue` / `getError` / `onChange`, errors held as a Zod error tree via `setError`), with `AppForm` in `components/forms/form-context.tsx` delegating to `store.onSubmit`. I offered a mechanism that sounded plausible without checking `package.json`.

It does not change this pull request, and it does not change the conclusion that the save failure is unrelated to it — that rests on the failure reproducing for a first-name-only edit with the notes editor untouched, which is an observation, not an inference.

What the store actually shows is that `aria-invalid="true"` on `customFieldValues[0].value` means a Zod error tree was populated for that path, most plausibly from the server action rejecting the submit, with no message rendered for it. Recorded in full on CUS-112, including the parts that remain unproven.

benjiwagner:
## All four review fixes verified on the preview

Commit `7e2994d` is deployed. Re-ran the journey on `feat-editor-tables-and-images.customermates.com` rather than relying on the local run.

```
slash menu commands   Heading 1 … Divider, Table, Image, Link   (Link now present)

Image command
  nativePromptUsed    false
  inlineInput         true, placeholder "Enter image URL"

unloadable image url
  fallbackRendered    true
  href                https://example.com/definitely-missing.png
  text                https://example.com/definitely-missing.png

paste html <img>
  inserted            true, src https://example.com/preview-paste.png

one-column table
  cellsPerRow         [1,1,1]
  computedLayout      auto
  editorWidth         335px
  tableWidth          335px   (spans the editor)
  cellWidth           334px   (fills the table)
```

Before the fix the same table measured a 102px cell inside a 306px table, the remaining two thirds rendering as empty columns.

The preview record is back to `notes: null` with `Leila Chen` intact.

## Checks

`ci` and `repository-policy` both pass on `7e2994d`, Vercel deployed. CodeQL still reports exactly one alert — the same `js/incomplete-sanitization` false positive on the table cell serializer analysed earlier, pinned by a regression test. No new alerts from this commit.

## Still open, unchanged

A clipboard holding only image *binary* — a screenshot rather than an image copied from a page — inserts nothing, because under the external-reference model there is nowhere to put the bytes and this repository has no upload endpoint or blob storage. Left for a separate decision alongside CUS-104.

benjiwagner:
## Markdown table paste — real bug, fixed in `a5598f3`

Confirmed and reproduced. Pasting a markdown table converted only when the clipboard held plain text *alone*. Copying markdown from a chat client, a browser or most editors also puts a `text/html` flavour on the clipboard, and the paste handler bailed out whenever that flavour was present — so the table arrived as literal pipes:

```
| Name | Role |
| --- | --- |
| Ada | CTO |
```

The plain text is now parsed first and preferred over the html only when it yields block structure the html paste would lose — table, image, heading, list, task list, code block, quote or rule. Verified locally:

```
markdown table + text/html on clipboard   -> TH:Name TH:Role / TD:Ada TD:CTO, bold preserved
rich html, no markdown structure          -> bold and italic kept (html still wins)
```

11 new tests cover which pasted text is treated as markdown and which defers to html.

## MCP and REST verified locally, on the real endpoints

Previously the MCP path was only covered by a unit test with mocked interactors. It has now been exercised against the running application with a real API key created through `better-auth`, on `POST /api/v1/mcp` — the exact route that produced `CUSTOMERMATES-3P`.

```
initialize                      -> protocolVersion 2024-11-05, server ready

tools/call update_record_notes  -> "Updated notes for 1 contact(s)"
  notes: "| Name | Role |\n| --- | --- |\n| Ada | **CTO** |\n| Bo | Eng |\n\n![logo](…/mcp-logo.png)"

tools/call get_records          -> notes returned byte-identical:
  "| Name | Role |\n| --- | --- |\n| Ada | **CTO** |\n| Bo | Eng |\n\n![logo](…/mcp-logo.png)"
```

REST on the same local instance:

```
PUT /api/v1/contacts/{id}  -> 200
node types: doc, image, paragraph, table, tableCell, tableHeader, tableRow, text
```

So the crash path is now covered end to end on the real MCP endpoint, not just through mocks. The contact was reset to `notes: null` afterwards.

## Validation

```
yarn typecheck  pass
yarn lint       pass  (0 errors, 0 warnings)
yarn test       pass  105 files, 834 tests  (+11)
yarn build      pass  55.3s
```

benjiwagner:
## Audit: custom code vs installed packages — `4672c4e`

Question raised in review: the PR carries custom code although pre-existing extensions are in use. Ran a full audit of the diff against the **installed** package sources (not docs — the installed prosemirror-markdown 1.13.4 lacks features newer docs describe). 14 findings survived adversarial verification. Outcome: three simplifications applied, one bug found in this PR's own code, the rest is custom-for-a-verified-reason or follow-up material.

### Applied (net −36 lines)

1. **Table cell serializer** — was a sub-serializer built by casting onto prosemirror-markdown's `@internal` `MarkdownSerializerState` constructor. The public `MarkdownSerializer.serialize()` does the same job; the cast is gone. One deliberate rejection: `escapeExtraCharacters: /\|/g` looks like the built-in way to escape pipes, but it only applies inside `esc()` — pipes in inline code and link hrefs bypass it (`text(str, false)` call sites) and would emit broken GFM rows. Verified by probe; the manual whole-string escape stays.
2. **thead/tbody stripping** — hand-rolled token filtering replaced by the parser spec's own `ignore: true` handling. What remains of the custom rule (wrapping cell content in paragraphs) has no built-in equivalent: tiptap's own v3 table `parseMarkdown` does the identical wrap.
3. **HTML image paste branch** — deleted. `@tiptap/extension-image`'s `parseHTML` already handles pasted `` markup via the default paste path. The custom regex branch was also silently **dropping surrounding pasted content** (copy a paragraph containing an image → only the image survived); the native path keeps it. Verified in-browser.
4. **Bug found by the audit**: ProseMirror puts `ProseMirror-selectednode` on the node-view wrapper, never the ``, so this PR's `img.ProseMirror-selectednode` outline never rendered. Selector fixed and verified (2px primary outline appears exactly with the class).

### Custom code that stays, with the verified reason

- **Server-side markdown pipeline** (`editor.utils.ts`): nothing installed can replace it. `@tiptap/html` is not installed; `generateJSON` throws without a DOM; `tiptap-markdown` needs a live Editor; `@tiptap/extension-table` ships a DOM-free `renderTableToMarkdown` but its consumer `@tiptap/markdown` is not installed **and it does not escape pipes in cells** (would emit broken GFM).
- **Phantom-column fix stays CSS**: `Table.renderHTML` emits the colgroup **unconditionally** — the full option surface (`HTMLAttributes, resizable, renderWrapper, cellMinWidth, View, lastColumnResizable, allowTableNodeSelection`) offers no way to suppress it. `table-layout: auto` is the minimal correct fix.
- **`hasMarkdownBlockStructure` paste heuristic**: no installed package has a prefer-markdown-over-html-flavour paste path; `transformPastedText` structurally cannot fire when `text/html` is present.
- **Table menu follows the repo's manual-positioning pattern**: `@tiptap/react/menus` ships a floating-ui `BubbleMenu` that would be less code, but the KB convention ("match the nearest established surface" — `engineering/frontend.md`) points at the pre-existing bubble/slash menus. Migrating **all three** menus to the shipped component is the right move — as a follow-up, filed as a Linear issue.

### Validation

```
typecheck pass · lint pass (0 warnings) · tests 105 files / 833 pass · build 52.3s
```

Browser re-verified after the changes: paste of a web fragment keeps text + image; the image-error link fallback still renders via the native path.

benjiwagner:
## Security fix in `1552a9d` — found while documenting the sanitizer

Answering a review question about how notes validation prevents unsupported content, I probed the live API and found a genuine XSS vector **introduced by this PR**, now fixed.

**The vector.** `validateNotes` checks node and mark *types* against `editorSchema` but does not restrict attribute *values*, so an image with `src: "javascript:alert(1)"` is accepted and stored. That is harmless on its own — an `` with a `javascript:` src is inert. But this PR added a node view that swaps a failed image for a link to its url, which turned it live:

```
<a class="editor-image-fallback" href="javascript:alert(1)" target="_blank">
```

Reproduced in the browser: paste ``, the image fails, the anchor renders with that href, and clicking it executes in the page origin. Reachable by anyone who can write notes — including any MCP agent.

**The fix.** A fallback is a link only for `http`/`https`. Anything else renders as inert text, so the address stays visible but is not clickable. Verified after the change:

```
javascript:alert(1)                    -> <span>, no href
data:text/html;base64,...              -> <span>, no href   (test)
https://example.com/still-broken.png   -> <a href="https://…">   (still works)
```

10 new cases pin the refusals (`javascript:`, mixed-case `JaVaScRiPt:`, `data:text/html`, `vbscript:`, `file://`, relative, empty).

## Sanitizer behaviour, as measured

For the record, probed against the running API. All of this is **pre-existing** — `core/validation/validate-notes.ts` is byte-identical on `origin/main` and untouched by this PR.

```
unsupported node type (iframe)     REJECTED 400
unsupported mark                   REJECTED 400
raw <script> in markdown           stored as literal text, not a node  (markdown-it html:false)
unknown attr (onerror) on image    ACCEPTED and stored
javascript: image src              ACCEPTED and stored
tableRow at doc level              ACCEPTED (invalid structure)
bare paragraph as root (no doc)    ACCEPTED
```

Two structural notes, both worth separate issues rather than scope creep here:

1. The validator **checks without normalising** — `Node.fromJSON(editorSchema, parsed)` runs inside a `try` purely as a type check and the result is discarded; the original object is returned and stored. Returning `Node.fromJSON(...).toJSON()` instead would strip unknown attributes for free.
2. It uses `create` semantics, so content expressions are not enforced — `tableRow` at doc level and a non-`doc` root are both accepted, which can produce notes the editor cannot render.

Filing those. Attribute-value policy for image `src` remains CUS-104.

PR #25, opened 2026-07-22
feat(onboarding): add workspace entity terminology and actionable empty states

benjiwagner:
## Summary

Replaces onboarding demo-data seeding with a workspace **entity-terminology** system and gives every empty data surface an intentional, actionable empty state. Closes [CUS-55](https://linear.app/customermates/issue/CUS-55).

- **No more demo seeding.** The onboarding wizard drops the "demo data" and product/service steps; new workspaces receive no sample contacts, organizations, deals, services, tasks, custom fields, pipeline/status options, or dashboard widgets. The `SeedOnboardingDataInteractor`, `seed-data` module, and the seeding methods on `PrismaUserRepo` are removed, along with their orphaned `Common.seedData.*` / `Common.defaultData.*` translations.
- **Workspace terminology.** A new tenant-scoped `EntityTerminology` table (keyed `@@unique([companyId, entityType])`) stores a per-company display name for contacts, organizations, deals, and services. Each entity can use a localized **preset** (Contact/Person/Client, Organization/Company/Account, Deal/Opportunity/Project, Service/Product/Offering) or a **custom** singular/plural pair. Onboarding is the first entry point; workspace admins can change the names later under **Company → Settings → Entity names**.
- **Central resolver, canonical ids untouched.** A single resolver (`resolveEntityTerm`) + a `useEntityTerminology()` hook feed configured names to the sidebar, breadcrumb/heading, global add picker, table column headers, relation-field labels, create/edit drawer titles, and empty states. Preset names localize with the active locale; custom names are stored verbatim (deterministic across locales); no override falls back to today's labels. Canonical route segments, Prisma models, REST/MCP params, webhook events, permission resources, and the `EntityType` enum are unchanged.
- **Actionable empty states.** A reusable, accessible empty-state contract lives in the shared data view. It distinguishes a genuinely-empty collection (icon + "Add your first {Client}" CTA that opens the same permission-checked creation flow), zero search/filter results ("No matches" + Clear filters, no create CTA), a read-only collection, and loading. Dashboard, Contacts, Organizations, Deals, Services, and Tasks are covered; the Dashboard and the webhooks/deliveries/audit-log collections get contextual empties too.
- Retires `Company.salesType` and the `SalesType` enum (now orphaned) via a forward migration.

## Context

Verified against `origin/main`. Previously the onboarding wizard was `profile → entities → demoData → invite → ai`; the demo-data step defaulted **on** and seeded a synthetic CRM graph, masking an empty workspace, while the shared table/card/Kanban views fell back to a generic hard-coded "No items found." `Company.salesType` was written only inside the seed path.

The wizard is now `profile → terminology → invite → ai`. Terminology is delivered to the client through the existing `app/layout.tsx → NavigationSwitch` bootstrap into a MobX `TerminologyStore`, mirroring how `company` is hydrated. Terminology edits are gated by the same `company/update` permission as other company settings.

## Validation

Local, in an isolated worktree with a disposable PostgreSQL 16 and `APP_MODE=demo`.

- `yarn typecheck`, `yarn lint` (0 warnings), `yarn test` (all suites), and `yarn build` pass.
- New focused tests (18): terminology resolver (preset localization, custom verbatim, default fallback, full map), Zod validation (preset/custom, unsafe markup, cross-entity collisions, non-configurable entity, trimming), selection round-trip, and the upsert interactor (persists valid input, rejects invalid without touching the repo).
- Fresh-DB `prisma migrate deploy` + `prisma db seed` succeed with the column/enum dropped and the new table created.
- Browser journeys (demo mode): configured a workspace to Clients / Accounts / Opportunities (custom) / Products and confirmed the names apply live and consistently across the sidebar, page heading, and table column headers with no reload, while `/contacts`, `/services`, etc. routes stay canonical. The onboarding wizard shows **Step 1 of 4** (demo-data and product/service steps gone). Emptying a collection shows **"No Products yet / Add your first Product"**; a non-matching search shows **"No matches / No Clients match your filters"** with Clear filters and no create CTA.

## Impact and rollback

- **Migration** `20260722100000_entity_terminology_drop_sales_type`: creates `EntityTerminology`; drops `Company.salesType` and the `SalesType` enum. Forward-only and data-preserving — existing companies and their records are untouched, and with no override every workspace keeps its current visible labels. Production migration/release remain human checkpoints.
- **Rollback**: revert the branch and restore the column/enum with a down-migration; no workspace data is deleted at any point. Existing seeded/demo records in established workspaces are intentionally left in place.
- Terminology is display-only. Agent/integration/programmatic surfaces continue to use canonical entity names; a configured alias never changes an id.


prisma/schema.prisma
```diff
@@ -240,6 +239,20 @@ model Company {
   calendars                   Calendar[]
   calendarEvents              CalendarEvent[]
   subscription                Subscription?
+  entityTerminology           EntityTerminology[]
+}
+
+model EntityTerminology {
```
benjiwagner, line 245: Is the MCP server, when we get the current workspace config, aware of the entity terminology that is chosen so that AIA can reply in the correct way and understand the user?

features/user/prisma-user.repository.ts
```diff
@@ -43,18 +28,6 @@ import { FilterFieldKey } from "@/core/types/filter-field-key";
 import { FILTER_FIELD_DEFAULT_OPERATORS } from "@/core/types/filter-field-operators";
 import { env } from "@/env";
 
-const TASK_STATUS_STATES = [
```
benjiwagner, line 46: The task status, so an initial custom field for this under to use should still be available initially

features/company/get-company-details.interactor.ts
```diff
@@ -11,7 +11,6 @@ import { ValidateOutput } from "@/core/decorators/validate-output.decorator";
 const OutputSchema = z.object({
   id: z.string(),
   currency: z.enum(Currency),
-  salesType: z.enum(SalesType).nullable(),
```
benjiwagner, line 14: Should the current terminology chosen by the current company be available on the get company details as well. Not quite sure whether this is the related interactive for the settings page where we also show this behavior. We have a related comment which is asking whether the workspace details is already included in the terminology so that AI agent slash MCP are aware of these settings and respond to questions and answers in the correct manner. The comments may be related.

app/[locale]/(protected)/company/components/company-settings/company-settings.store.ts
```diff
@@ -1,35 +1,83 @@
...
     event?.preventDefault();
     this.setIsLoading(true);
 
     try {
-      const res = await updateCompanyAction(toJS(this.form));
-
-      if (res.ok) {
-        this.onInitOrRefresh(res.data);
+      if (this.form.currency !== this.savedState.currency) {
+        const res = await updateCompanyAction({ currency: this.form.currency });
```
Comment (benjiwagner) [inline app/[locale]/(protected)/company/components/company-settings/company-settings.store.ts]: Currently we have two separate actions in the company settings store, one for the update company and the other for terminology. Isn't it more correct if we simply merge these two together so that we don't have separate actions or separate interactors for the terminology? Because essentially, this is a company setting, so we use the update company settings action. Something like this, so we generalize or unify it better.

features/entity-terminology/entity-terminology.constants.ts
```diff
@@ -0,0 +1,98 @@
...
+export const CONFIGURABLE_TERMINOLOGY_ENTITY_TYPES = [
+  EntityType.contact,
+  EntityType.organization,
+  EntityType.deal,
+  EntityType.service,
+] as const;
+
+export const ENTITY_TERMINOLOGY_PRESETS: Record<EntityType, TerminologyPreset[]> = {
+  [EntityType.contact]: [
+    { key: "contact", singular: "Contact", plural: "Contacts" },
```
benjiwagner: Why do we have the actual labels like singular and plural defined in here? Shouldn't this be language specific slash translated instead? So that it works also in different languages aside from English?

prisma/schema.prisma
benjiwagner: Good catch, it was not aware. `get_workspace_context` returned only `id`, `currency` and timestamps, so an agent had no way to know a workspace calls contacts "People".

Fixed in 6e8e257. The resolved labels now come back from `GetCompanyDetailsInteractor` (translated server side, so they follow the caller's locale) and are surfaced as `company.terminology` in `get_workspace_context`. I also extended the tool description, since the data alone would not change behaviour:

> `company.terminology` gives the singular and plural label this workspace uses for each record type, keyed by the canonical entity type. Always phrase answers with those labels and map the words the user types back onto the canonical entity type. Tool names, filter fields and ids stay canonical regardless of the labels.

So the agent answers in "People" but still calls tools with `contact`. Your comment on `get-company-details.interactor.ts` is the same change.

features/user/prisma-user.repository.ts
benjiwagner: You are right, and this was a regression rather than an intended removal.

Dropping the demo seeding deleted 600 lines from this file, which took the structural custom columns with it and not just the demo records. Your comment is anchored exactly on the deleted `TASK_STATUS_STATES`.

Fixed in 6e8e257: provisioning a workspace creates the Status single-select on tasks again (Open / In Progress / Blocked / On Hold / Done / Archived, with Open as default), the `Common.defaultData.todo` keys are restored in both locales, and no demo records come with it.

One thing I left out deliberately: the same deleted block also created a **Sales Pipeline** single-select on deals. You named task status, so I did not widen the scope on my own judgement, but a new workspace currently has no deal stage field at all. Say the word and I will add it back.

features/company/get-company-details.interactor.ts
benjiwagner: Yes, and you are right that this and the MCP comment are the same issue.

Done in 6e8e257: `GetCompanyDetailsInteractor` now returns `terminology` alongside `currency`, resolved through i18n on the server. The settings page reads it and `get_workspace_context` passes it to the agent, so there is a single source for both surfaces.

Related, from your other comment: the terminology write now also goes through `updateCompany` instead of a separate interactor.

app/[locale]/(protected)/company/components/company-settings/company-settings.store.ts
Comment (benjiwagner) [reply app/[locale]/(protected)/company/components/company-settings/company-settings.store.ts]: Agreed, merged in 6e8e257.

`UpdateCompanyDetailsSchema` now takes optional `currency` and optional `terminology`, and `UpsertEntityTerminologyInteractor` plus its server action are deleted. Both the settings form and the onboarding wizard call `updateCompany` now. Permissions were already identical (`company:update`) so nothing changed there, and the repo ports moved to `entity-terminology.repository.ts` so persistence stays separated from the interactor.

One thing I deliberately did not touch: `COMPANY_UPDATED` is a public webhook payload typed `{ currency }`, so I did not widen it. It still fires only when the company row itself changes, which means a terminology-only save emits nothing, same as before the merge. If you want terminology changes to emit an event too, that is a webhook contract change and I would rather do it as its own PR.

features/entity-terminology/entity-terminology.constants.ts
benjiwagner: Fair point, and fixed in 6e8e257.

They were in fact already translated: the resolver looks up `EntityTerminology.presets.<entityType>.<presetKey>.<form>`, and both locales carry every key, which is why the German UI renders "Kontakte" and "Organisationen". These strings were only a fallback for a missing key. But two sources of truth invite drift, and the hardcoded English reads like the real one.

Presets are now just keys:

```ts
export const ENTITY_TERMINOLOGY_PRESETS: Record<EntityType, string[]> = {
  [EntityType.contact]: ["contact", "person", "client"],
  ...
};
```

and the fallback is gone, so i18n is the only source of a label. That is safe because the i18n parity and key-resolution gates fail CI on a missing key, which is a stronger guarantee than a silent English fallback would be.

features/company/get-company-details.interactor.ts
benjiwagner: Update: taken further in 56893e8, after you asked whether the whole thing should collapse into one company settings endpoint.

`GetCompanyDetails` / `UpdateCompanyDetails` are now `GetCompanySettings` / `UpdateCompanySettings`, matching the route and the `companySettingsStore` that already used that name. The standalone terminology reader, its action and its DI entry are deleted, so there is one read and one write for workspace settings, both partial-friendly and both behind `company:readOwn`.

The read returns terminology in the two shapes its callers need, so nobody makes a second call:

```ts
terminology: {
  presets: [{ entityType, presetKey }],            // for editing the selects
  labels:  { contact: { singular, plural }, ... }  // for display and for agents
}
```

Worth noting how real the duplication was: `app/layout.tsx` was calling company details **and** the terminology reader in the same `Promise.all` on every page load. That is one call now, and the onboarding wizard page lost its extra call the same way.

app/[locale]/(protected)/company/components/company-settings/company-settings.store.ts
Comment (benjiwagner) [reply app/[locale]/(protected)/company/components/company-settings/company-settings.store.ts]: Update: extended in 56893e8 to cover the read as well, not just the write.

The interactors are renamed to `GetCompanySettings` / `UpdateCompanySettings` to match this store and the `/company/settings` route, and the separate terminology reader is gone. So the store now does one read and one write against the same endpoint, and the onboarding wizard uses that same pair.

What I kept separate, and why: `EntityTerminology` keeps its own module and table repository. The presets, resolver and schema are vocabulary rules consumed by the sidebar, tables, detail pages and the diagram, so folding them into `features/company` would have half the UI importing vocabulary from the company feature. The table also has its own composite-key upsert semantics, and DI already has 22 Prisma repos, so one repo per model is the existing convention rather than something invented here. Happy to go further if you disagree.

One regression I caught while doing it: making `currency` optional on the shared schema quietly loosened the MCP `update_workspace_settings` tool, where a `target: company` call with no currency would have parsed fine and silently done nothing. That branch now uses `UpdateCompanySettingsSchema.required({ currency: true })`, so the tool keeps its previous contract.

prisma/schema.prisma
benjiwagner: Update in 56893e8: the tool now receives the resolved labels specifically (`company.terminology.labels`), since the raw preset keys are noise for an agent. The read it calls was renamed to `GetCompanySettings` and is now the single endpoint for workspace settings, so terminology arrives on the same call that already fetched the company.

features/user/prisma-user.repository.ts
benjiwagner: Update: now covered by a test in 284b07c, `features/user/__tests__/prisma-user.repository.test.ts`.

It drives the real `createCompanyAndUser` path and asserts a new workspace gets a task status single-select with the full lifecycle (Open, In Progress, Blocked, On Hold, Done, Archived), Open selected by default, distinct option values, and that provisioning creates nothing beyond the scaffolding. I checked it earns its place: removing the `createDefaultCustomColumns` call fails all three cases, which is exactly the regression that shipped when the demo seeding was deleted.

I went with a test rather than a manual signup partly so this is guarded in CI from now on, and partly because I do not create accounts.

Still open from my earlier reply: the same deleted block also created a **Sales Pipeline** single-select on deals, which I left out because you named task status specifically. A new workspace currently has no deal stage field. Say the word and it is a small addition alongside this one.

app/[locale]/(protected)/company/components/company-settings/company-settings.store.ts
```diff
@@ -1,35 +1,80 @@
...
+    this.form = { ...this.form, terminology: { ...this.form.terminology, [entityType]: presetKey } };
+  };
+
   onSubmit = async (event?: FormEvent<HTMLFormElement>) => {
     event?.preventDefault();
     this.setIsLoading(true);
 
     try {
-      const res = await updateCompanyAction(toJS(this.form));
+      const currencyChanged = this.form.currency !== this.savedState.currency;
```
Comment (benjiwagner) [inline app/[locale]/(protected)/company/components/company-settings/company-settings.store.ts]: For the other on submit methods, I think we did not check whether the state is actually dirty for the form.

features/company/get-company-settings.interactor.ts
```diff
@@ -0,0 +1,59 @@
...
+export abstract class GetCompanySettingsRepo {
+  abstract getDetails(): Promise<Company>;
+}
+
+@AllowInDemoMode
+@TenantInteractor({ resource: Resource.company, action: Action.readOwn })
+export class GetCompanySettingsInteractor extends AuthenticatedInteractor<void, CompanySettings> {
+  constructor(
+    private repo: GetCompanySettingsRepo,
+    private terminologyRepo: GetEntityTerminologyRepo,
```
benjiwagner: I think we don't need a separate repository for the terminology because this should be part of the company. So we can merge both together so that we have one abstract class.

benjiwagner: This is obviously also the case for the other interactors which we have changed, which are currently terminology specific or using the Prisma terminology repository, which should be part of the common repository.

PR #29, opened 2026-07-22
feat(activities): filter shared activity timelines by connected channel

benjiwagner:
## Summary

Adds a **Channel** filter to every entity activity timeline (contact, organization, deal, service, task), letting users narrow the shared history to one or more concrete connected channels (`ConnectedAccount`) — independently of the existing Provider and Conversation filters. The same shared filter contract (field, option source, parser, query semantics, permission checks) is what [CUS-68](https://linear.app/customermates/issue/CUS-68) will reuse for the dashboard activity widget.

Key pieces:

- **Filter contract** — reuses `FilterFieldKey.connectedAccountId`, interpreted as `in`/`notIn` in `timeline-filters` and advertised by the message-permission-gated filterable fields (never to audit-only viewers).
- **Consistent source filtering** — one shared source-filter plan feeds both `getItems` and `getCount`, applying channel *and* provider across the message, account-activity and calendar-event sources (messages via the direct column, account-activity/calendar via the `connectedAccount` relation). Audit entries have no channel, so an active provider or channel filter suppresses them.
- **Permission-aware option source** — reuses the existing `GetMyConnectedAccountsInteractor` (owned-or-shared, permission-gated, demo-allowed) — the same lister already behind the connected-accounts page, REST route and MCP tools — mapped into the filter dropdown with provider icons; owner context is appended only for shared channels, and deleted accounts are filtered out. This mirrors how the entity relation filters reuse their canonical list actions rather than adding a parallel option-source stack.
- **Fail-closed validation** — `connectedAccountId` now validates through the query-param precheck via a new accessible-scoped `ValidateConnectedAccountIdsInteractor`, so unknown or inaccessible ids fail closed rather than widening a query.
- **i18n** — localized field label (EN "Channel" / DE "Kanal"), plus a `connectedAccountNotFound` message.

Scope note: this delivers the shared filter + entity-history integration only. The dashboard widget itself remains CUS-68 (this PR blocks it).

## Context

Implements [CUS-80](https://linear.app/customermates/issue/CUS-80). Entity history already exposed Type, Conversation and Provider filters, but there was no way to scope activity to a specific connected account, and provider selection was applied only to messages (not account-activity or calendar-event). `FilterFieldKey.connectedAccountId` already existed for calendar/MCP contracts but the activity repository neither advertised nor interpreted it, and the generic filter UI treated it as free text. Reusing that key (rather than introducing a new one) keeps a single shared contract across activity, calendar, REST/MCP and the future widget; upgrading its value-kind to an accessible-scoped entity id also adds fail-closed validation to the calendar/MCP consumers that previously accepted it unvalidated.

## Validation

- `yarn typecheck` ✅, `yarn lint` ✅ (repo-wide, `--max-warnings=0`), `yarn test` ✅ (867 tests).
- New focused tests: `timeline-filters` channel parsing (in/notIn, empty arrays, independence from provider/conversation); `prisma-activities.repository` permission visibility of the field, audit suppression + channel predicate across all three sources for both `getItems` and `getCount`, `notIn`, and provider consistency via the relation.
- **Exercised live** on a local instance (seeded synthetic data, demo mode) against a contact history (Anna Müller: 31 Gmail messages + 2 audit entries in scope):
  - Field renders as **Channel** alongside Type/Conversation/Provider; option source loads the accessible channels with provider icons; both `in` and `not in` operators offered.
  - `Channel in <Gmail>` keeps her messages and shows a readable active-filter chip (no raw UUID); `Channel in <WhatsApp>` (a channel with zero activity for her) yields **"No activity"** — proving the filter genuinely constrains, not a no-op.
  - Audit suppression, isolated: `Type in Changes` alone surfaces the 2 audit entries; adding `Channel in <Gmail>` on top makes them **vanish** (empty) — audit has no channel, so it drops out under an active channel filter.
  - Localized to **Kanal**, distinct from **Anbieter**/**Unterhaltung**, in German.
  - Multi-account, on an organization history that aggregates two connected accounts (a LinkedIn account + a WhatsApp account): unfiltered shows both; `Channel in <LinkedIn>` narrows to LinkedIn only (the WhatsApp group chat drops out); selecting both accounts unions them (both return) — confirming the option list, single-account narrowing, and multi-select union.
  - All three channel-backed sources, using seeded synthetic rows (one LinkedIn account-activity + one Google calendar event attached to a contact): messages filter by channel as above; `Channel in <Google>` leaves only the calendar event (messages and the LinkedIn account-activity drop out); `Type in Activities` + `Channel in <LinkedIn>` leaves only the account-activity (the calendar event and messages drop out) — confirming the message, account-activity, and calendar-event sources each filter by connected account.
  - Re-verified end-to-end after the two follow-up refactor commits (option-source reuse and the naming alignment): the dropdown still lists the accessible channels with provider icons, and the LinkedIn/Gmail multi-source matrix above still holds.
- Not exercised in the running UI (covered by unit tests instead): `notIn` semantics, item/count parity, and REST/MCP. The other four entity histories (deal/service/task) share the same repository + UI path verified on the contact and organization. Filter persistence across reload is not exercised locally because demo mode deliberately skips the personalization write.

## Impact and rollback

Additive and gated behind existing messaging read permission; no schema or data migration. Provider filtering now also constrains account-activity and calendar-event sources and suppresses audit when a provider/channel filter is set — a deliberate consistency fix relative to prior behavior. The `connectedAccountId` value-kind change adds accessible-scoped id validation to the calendar/MCP filter consumers that already accept it (they pass real ids, so no behavior change for valid input). To roll back, revert this commit; persisted channel ids are ignored safely by readers without the field, and existing Provider/Conversation filters continue to work.


benjiwagner:
Heads-up on the red **Vercel** check: it is a build-container OOM, not a regression from this PR.

Vercel's own build-system report says: `At least one "Out of Memory" ("OOM") event was detected during the build`, with the Enhanced Build machines link. The build dies during "Creating an optimized production build" with `SIGKILL`.

Evidence it is not this changeset:
- Peak build memory measured with `/usr/bin/time -l yarn build`: `origin/main` **6.3 GB** vs this branch **4.55 GB**. This branch peaks *lower* than the baseline that deploys green.
- `yarn build` succeeds locally on this branch (~50s), and `ci`, `repository-policy` and CodeQL are all green.
- The failing commits include one that removed 152 lines net and one that changed five test-title strings, neither of which can raise build memory.
- A plain same-commit redeploy also OOM'd, so it is sustained memory pressure at the container ceiling rather than one-off flakiness.

The durable fix is enabling Vercel Enhanced Builds for the project, which is outside this PR.

ee/messaging/activities/timeline-filters.ts
```diff
@@ -12,11 +12,14 @@ export type ActivityQuery = {
   providers?: Set<string>;
   threadIdsIn?: Set<string>;
   threadIdsNotIn?: Set<string>;
+  connectedAccountIdsIn?: Set<string>;
+  connectedAccountIdsNotIn?: Set<string>;
 };
 
 const FIELD_KIND: string = FilterFieldKey.timelineKind;
```
benjiwagner: Why dont we inline use them instead?

ee/messaging/activities/prisma-activities.repository.ts
```diff
@@ -21,6 +21,33 @@ import type { ActivityThreadOptionsData, ActivityThreadOptionsRepo } from "./get
...
+  providers?: string[];
+  threadIds?: string[];
+  threadIdsNotIn?: string[];
+  connectedAccountIdsIn?: string[];
+  connectedAccountIdsNotIn?: string[];
+};
+
+type ChannelWhere = { connectedAccountId?: { in?: string[]; notIn?: string[] } };
+
+function channelWhere(filters: SourceFilters): ChannelWhere {
```
benjiwagner: Let's check also the other repositories for the local function and type definitions within the Prisma Activities repository. This looks like we reinvent something that is done differently in the other repos

ee/messaging/activities/timeline-filters.ts
benjiwagner: Done in 1f236ba — inlined.

Worth recording why they existed: `filter.field` is a plain `string`, so comparing it directly against a `FilterFieldKey` member trips `@typescript-eslint/no-unsafe-enum-comparison` (4 errors when I first inlined). The `FIELD_*: string` constants were a workaround for that.

The cleaner fix is the idiom already used in the data-view filter components (`use-filter-select-items.tsx:72`, `filter-field.tsx:103`): cast once and compare inline. So `interpretFilters` now does `const field = filter.field as FilterFieldKey;` at the top of the loop and compares `field === FilterFieldKey.timelineKind` etc. All four constants are gone, lint is clean.

ee/messaging/activities/prisma-activities.repository.ts
benjiwagner: Good catch — you were right, I checked and no other repository does this. Fixed in 1f236ba.

What the survey showed:
- Only two `prisma-*.repository.ts` files have module-level helpers at all: this one and `prisma-messaging.repository.ts`. But that one's helpers are pure data utilities (`cleanString`, `dedupeParticipants`, `safeTruncate`, `sanitizeMessage`), not where-builders, and its single local type is a `ReturnType<...>`, not a hand-rolled Prisma shape.
- Shared where-fragments in this codebase live in **domain modules**, exported: `messaging-access.ts` (`threadAccessWhere`, `calendarEventAccessWhere`, `accountActivityAccessWhere`) and `provider.ts` (`classWhere`).

So `channelWhere`/`providerRelationWhere` plus the `SourceFilters`/`ChannelWhere` types were genuinely misplaced. They now live in `timeline-filters.ts` next to the `ActivityQuery` they read, and take `ActivityQuery` directly. That let me delete both local types and the Set-to-array flattening, and `getItems`/`getCount` now pass the same object (which is what keeps items and counts in lockstep). The provider and thread predicates that were inlined twice each in the message queries are now the same shared fragments too.

Repository is net -151 lines and is pure composition; no hand-rolled Prisma types remain (return types are inferred).

Re-verified against the local database after the refactor, on a contact with all three source types:

| filter | calendar_event | activity | message | audit | items=count |
|---|---|---|---|---|---|
| none | 1 | 1 | 25 | 2 | 29=29 |
| channel in linkedin | 0 | 1 | 25 | 0 | 26=26 |
| channel in google | 1 | 0 | 0 | 0 | 1=1 |
| channel in [google,linkedin] | 1 | 1 | 25 | 0 | 27=27 |
| channel notIn google | 0 | 1 | 25 | 0 | 26=26 |
| provider in linkedin | 0 | 1 | 25 | 0 | 26=26 |

typecheck, lint and 866 tests green.

benjiwagner:
Follow-up verification on 1f236ba: UI e2e re-run, and the filter-value validation exercised **through MCP** against a local build (this had not been covered before).

**MCP** (real `/api/v1/mcp` endpoint on the local instance, better-auth API key, `get_activities` tool):

| case | result |
|---|---|
| valid channel id (Google) | returns exactly the 1 calendar event |
| unknown uuid | `Validation error: Channel ID not found or not accessible.` at `filters[0].value[0]`, `isError: true` |
| malformed value (`not-a-uuid`) | same rejection |
| **another user's private channel** (exists in the same company, `shared=false`) | **same rejection** — fails closed, and is indistinguishable from a nonexistent id, so it does not disclose existence |
| same channel after flipping `shared=true` | **accepted** (0 items, correctly) |

The last two are the important pair: they show the validator keys on the owned-or-shared visibility rule rather than mere row existence, and that MCP runs in API mode where an inaccessible id hard-rejects instead of silently returning an empty page. The test account and API key were removed afterwards.

**UI e2e** re-run on the refactored code, contact history with all three source types: applying `Channel in <LinkedIn>` keeps the LinkedIn account-activity and messages and drops the Google calendar event; chip renders as `Channel in Max Bergmann · LinkedIn`.

Repository-level check against the local database, all six filter shapes, items equal count in every case:

| filter | calendar_event | activity | message | audit | items=count |
|---|---|---|---|---|---|
| none | 1 | 1 | 25 | 2 | 29=29 |
| channel in linkedin | 0 | 1 | 25 | 0 | 26=26 |
| channel in google | 1 | 0 | 0 | 0 | 1=1 |
| channel in [google,linkedin] | 1 | 1 | 25 | 0 | 27=27 |
| channel notIn google | 0 | 1 | 25 | 0 | 26=26 |
| provider in linkedin | 0 | 1 | 25 | 0 | 26=26 |

public/v1/openapi.json
```diff
@@ -7,7 +7,7 @@
   },
   "servers": [
     {
-      "url": "http://localhost:4000/api",
```
benjiwagner, line 10: Currently, we always push the localhost port, which we use for the local development as well. And can we do this more elegant so that we don't have this in the change set for every PR when we test it locally?

ee/messaging/messaging-access.ts
```diff
@@ -4,6 +4,10 @@ function accessibleAccountWhere(userId: string): Prisma.ConnectedAccountWhereInp
   return { OR: [{ userId }, { shared: true }] };
 }
 
+export function accessibleConnectedAccountWhere(companyId: string, userId: string): Prisma.ConnectedAccountWhereInput {
```
benjiwagner, line 7: For the helper functions with the where, we sometimes include the company ID and sometimes we don't, as you can also see in the change set and other files. Please deeply check the code base and align this to one coherent pattern.

public/v1/openapi.json
benjiwagner: Fixed in 4075ead — the spec is no longer tracked.

It is a build artifact, not source: `dev`, `build` and `vercel-build` (which ends in `yarn build`) all run `openapi:generate`, so every deploy regenerates it and the committed copy is never what production serves. Its `servers[0].url` is `${env.BASE_URL}/api`, which is why testing locally on any port other than the last committer's rewrites the file. Worth noting the committed copy on `main` was advertising `http://localhost:4000/api`, and `public/` is served publicly.

The precedent was already in the repo: the sibling output of the same generator, `content/api/**/*.mdx`, is gitignored under a heading that literally reads "generated openapi docs", and `generated/` is ignored too. This spec was the only `generate-openapi.ts` output still tracked, so I ignored it alongside and dropped it from the index.

Safe because nothing reads the committed file: the generator writes it, and `core/fumadocs/api-page.tsx` fetches it over HTTP at `${env.BASE_URL}/v1/openapi.json` at runtime. Verified by deleting the file, running `yarn openapi:generate`, and confirming it is recreated and then invisible to git.

ee/messaging/messaging-access.ts
benjiwagner: Surveyed it, and the incoherence turned out to be mine at the call sites rather than in the helpers. Fixed in 4d603d5.

**What decides the pattern:** the tenant guard in `prisma/db.ts` checks `args.where.companyId` **at the top level only** for read operations (`if (!args.where?.companyId) throw tenantError(...)`). A companyId nested inside an `AND` array does not satisfy it. So top-level companyId is forced, and any companyId a helper contributes only counts when that helper is spread at the top level.

**Inventory of every where-fragment helper:**

| helper | includes companyId |
|---|---|
| `threadAccessWhere`, `calendarAccessWhere`, `calendarEventAccessWhere`, `accountActivityAccessWhere`, `accessibleConnectedAccountWhere` | yes (5 of 5 exported access helpers) |
| `accessibleAccountWhere` (private) | no — it is a relation-inner predicate used as `connectedAccount: { is: ... }`, never at top level |
| `classWhere`, `folderMessageWhere`, `threadFolderMembershipWhere`, `threadHasActivityWhere`, and the new `channelWhere`/`providerWhere`/`threadWhere` | no — pure predicate fragments, not tenant scoping |

So the helpers are already consistent. **The rule:** an access-where helper takes `(companyId, userId)`, includes companyId, and is spread at the **top level** of the query; other predicate fragments carry no companyId.

**What I had broken:** moving the access-where inside `AND` meant its companyId no longer satisfied the guard, so I had added a second top-level `companyId` — scoping the same model twice in four queries (account-activity and calendar-event, list and count), plus a redundant explicit companyId in `listThreads`. I had used `AND` because spreading `accountActivityAccessWhere` and `providerRelationWhere` together collides on the `connectedAccount` key.

**Fix:** spread the access-where at the top level again (as the calendar repositories already do) and keep only the extra predicates in `AND`. Prisma conjoins top-level fields with `AND` entries, so both constraints still apply and nothing collides. companyId now appears exactly once per query. The message queries keep their own top-level companyId next to `thread: threadAccessWhere(...)` — those scope two different models, so that is not duplication.

Behaviour is unchanged: re-ran the six filter shapes against the local database and got byte-identical results with items still equal to count.

| filter | calendar_event | activity | message | audit | items=count |
|---|---|---|---|---|---|
| none | 1 | 1 | 25 | 2 | 29=29 |
| channel in linkedin | 0 | 1 | 25 | 0 | 26=26 |
| channel in google | 1 | 0 | 0 | 0 | 1=1 |
| channel in [google,linkedin] | 1 | 1 | 25 | 0 | 27=27 |
| channel notIn google | 0 | 1 | 25 | 0 | 26=26 |
| provider in linkedin | 0 | 1 | 25 | 0 | 26=26 |

typecheck, lint and 887 tests green. I kept this to the queries this PR touches; no repo-wide churn was needed since the helper family was already coherent.

public/v1/openapi.json
benjiwagner: Update: the relative-URL approach **does** work, so I switched to that and the file stays tracked. Ignore my earlier reply saying it was gitignored — that is reverted, `.gitignore` is untouched, and the whole detour is squashed out of the history.

`servers[0].url` is now a relative `"/api"` instead of `${env.BASE_URL}/api`.

**Verified it renders correctly.** OpenAPI resolves a relative server against the document origin, and fumadocs handles it: on the local docs page the endpoint shows

- **Server URL** `http://localhost:4002/api`
- curl sample `curl -X POST "http://localhost:4002/api/v1/messaging/social-relations/accept"`

so consumers still get a correct absolute URL, and it will show the deployed origin in production instead of a hardcoded one.

**Verified the churn is gone.** Generating with `BASE_URL` set to `:4002`, `:4000` and `https://app.customermates.com` produces a byte-identical file; after regenerating on a different port `git diff` on the spec is empty. The change to the committed spec is a 1-line diff.

Side benefit: this also fixes the committed spec advertising `http://localhost:4000/api` publicly, since `public/` is served.

I also kept one small hardening in the generator — it now creates `public/v1` before writing, matching how the same script already prepares its mdx output directories, so generation no longer depends on the directory pre-existing.

PR #30, opened 2026-07-22
feat(legal): automate document notices and acceptance

benjiwagner:
## Summary

Completes CUS-56 with the smallest legal-document update flow that fits the existing Customermates architecture.

- Reuses the existing `AuditLog`, daily lifecycle cron, email service, route guard, server layout, and MobX root store. It adds no legal table, per-user legal-version columns, composite index, migration, request cache, aggregate layout interactor, release machinery, or environment variable.
- Uses two internal-only events: `legal.notice_sent` after a successful legal-notice email and `legal.documents_accepted` for company-wide Terms/DPA acceptance. Notice events use the recipient user ID; acceptance events use the company ID.
- Detects document changes only by comparing the four explicit `YYYY-MM-DD` document versions with schema-validated audit payloads. Audit records retain real entity/user/company IDs and `AuditLog.createdAt`; they contain no locale, deterministic version key, Git SHA, provider message ID, copied timestamp, or deployment metadata.
- Requires the existing checkbox for every managed-cloud onboarding user. Creating a new managed-cloud company records the current company-wide Terms/DPA acceptance atomically. An invited cloud user must separately acknowledge the Terms/DPA/Privacy copy, stored only in the existing `User.agreeToTerms` field; that acknowledgement emits no company-wide legal-acceptance event and cannot unlock or replace the company state. Demo and self-hosted registrations are exempt from hosted-service acceptance, and self-hosted onboarding shows no hosted legal copy.
- Sends administrators one combined email for relevant Terms, DPA, Privacy, and Subprocessor changes. Other active users receive Privacy notices only. Successful sends are deduplicated from audit history; a provider error creates no success event, stops the run, and is retried by the next daily lifecycle run.
- Stops outstanding Terms/DPA mail after any authorised administrator accepts the current contract versions. One acceptance restores protected UI access company-wide.
- Shows only action-pending legal UI: a primary sidebar alert during the 14-day grace period and a warning alert plus server-side `/legal-update` redirect after the deadline. Privacy/Subprocessor-only updates remain audited email communication and create no sidebar or legal-update state. The review page keeps the normal application shell/sidebar, uses the design-system checkbox and outlined secondary sign-out action, and redirects away immediately when no action remains.
- Adds a small `LegalUpdateStore` following the existing subscription-expiry/loading-overlay pattern. It owns checkbox state, acceptance, sign-out, and established validation-error toasts; no new global lint rule or client API was added.
- Exports the legal-audit Zod schemas from `legal-audit.schema.ts` and validates audit JSON directly at the Prisma persistence boundary, while `legal-audit.repo.ts` remains the repository port. Malformed historical payloads fail closed to `null`. The lifecycle sender exposes `invoke(): Promise<void>` and uses normal system time with fake time/module mocks in tests.
- Wraps shared rich Alert descriptions in one paragraph so `t.rich()` fragments and links flow inline; focused regression coverage protects plain descriptions, rich links, and the separate action cell.
- Uses the established tenant-interactor/UserAccessor pattern for legal status. The page, route guard, and root layout invoke it without passing or reloading the active user; demo and self-hosted exemptions remain inside the same interactor. Legal status stays in the authenticated root-layout `Promise.all` because the application shell needs it for the sidebar alert, adding no serial waterfall. A wrapper interactor or request cache would execute the same reads and add indirection without reducing queries or latency.
- Keeps the Lemon Squeezy affiliate scripts in the root layout unchanged. That client script creates the referral identifier on affiliate landing visits; moving it without end-to-end attribution proof could silently break referrals. This PR changes no affiliate, checkout, billing, SEO, subscription, or unrelated feature behavior.
- Updates only the English and German Privacy Policy to `2026-08-08` after removing the no-longer-persisted locale field from the audit disclosure. Terms, DPA, and Subprocessors remain `2026-08-07`; every EN/DE header exactly matches its application constant.

## Context

[CUS-56](https://linear.app/customermates/issue/CUS-56/add-unipile-disclosures-and-complete-the-gdpr-and-liability-review) owns this outcome. Its earlier prospective-only publication decision, invited-user information-only rule, persisted-locale design, information-only sidebar state, Git/deployment evidence design, and CUS-158's no-banner/no-gate design are superseded by this implementation.

English and German UI/email content remains localized, while the audit record identifies the accepted document versions rather than persisting a presentation locale. The German Terms/DPA remain authoritative and the English documents remain the convenience translation.

Accepted residuals:

- A crash after provider delivery but before audit persistence can produce a duplicate email.
- Fail-fast delivery means one persistently invalid recipient can delay later recipients until corrected.
- Legal audit events retain the audit system's existing cascade behavior; revisit this before adding production user/company deletion.
- Every substantive legal edit must bump the affected document version, with at most one substantive release per document per calendar day.
- Overdue acceptance restricts managed-cloud UI routes only. APIs remain available and no automatic user/company inactivation is introduced; manual operator inactivation remains the temporary fallback until separately approved lifecycle/operator work exists.

## Validation

- Disposable local PostgreSQL: all 29 migrations applied from zero and the deterministic synthetic seed completed twice; no schema or migration delta exists.
- Full Vitest suite against the disposable database: 162 files and 1,263 tests passed.
- Real-database journeys verify atomic new-company acceptance, invited-user personal acknowledgement without company acceptance, exact audit payloads, rollback, shared administrator/member deadlines, and company-wide restoration after one administrator accepts.
- Lifecycle coverage verifies recipient selection, combined mail, version comparison, later-user suppression, same-day handling, supplier deadlines, provider failures, fail-fast retries, strict legal-only Resend errors, and acceptance suppression.
- UI/store/convention coverage verifies admin/member grace and overdue states, action-only rendering, checkbox/store failures, inline rich links, EN/DE copy, and exact document-version parity.
- `yarn lint`, `yarn typecheck`, `yarn build`, OpenAPI/raw-doc generation, and `git diff --check` pass. The managed-cloud production build compiled 60 application pages including `/[locale]/legal-update`, Terms, Privacy, DPA, and Subprocessors.
- Local browser verification on the disposable database covered administrator and member sessions, primary grace-period and warning overdue alerts, protected-route redirection, real administrator acceptance persistence, company-wide restoration, accepted/info-only redirect-away behavior, German rendering, a 1440×900 desktop viewport, and a 390×844 mobile viewport. Clean-page loads had no framework overlay or console error.
- Fresh checks for final head `5002fe0c` are green: GitHub CI (including the 1,263-test suite), CodeQL for Actions and JavaScript/TypeScript, repository policy, and the Vercel preview deployment. Fresh Code Owner approval remains required.

## Impact and rollback

There is no external API, webhook, database-schema, billing, subscription, or automatic-inactivation change. Managed-cloud users receive the legal notification/acceptance flow; independently operated self-hosted installations do not.

Before publication, rollback is a normal revert. After a legal version is published, correct substantive text through a newly dated/versioned change instead of silently restoring an older public version.

Human-only completion remains: fresh Code Owner approval, human merge, and production verification of the deployed EN/DE legal routes, exact version constants, one notice cycle, and company-wide acceptance. PR #40 must then pin its product evidence to the actual merged product `main` commit.


eljulsi (commented):
Read it all. The changes are fine for me. However before I approve this, I'd like to talk about the email notification for all existing users saying "We changed our Privacy Policy, etc." I'm pretty sure we have to do that. Cheers, Jul

app/[locale]/(protected)/profile/components/connected-accounts-card.tsx
```diff
@@ -121,6 +150,8 @@ export const ConnectedAccountsCard = observer(({ accounts }: Props) => {
     <div className="flex w-full max-w-3xl flex-col gap-4">
       <Alert color="primary" description={t("ConnectedAccountsCard.description")} />
 
+      <ConnectDisclosure />
```
Comment (benjiwagner) [inline app/[locale]/(protected)/profile/components/connected-accounts-card.tsx]: This text should be part of the alert above and shorten both text when merging them together

app/[locale]/(protected)/onboarding/wizard/components/step-profile.tsx
```diff
@@ -58,6 +58,11 @@ export const StepProfile = observer(({ email, firstName, lastName, avatarUrl }:
                     {chunks}
                   </AppLink>
                 ),
+                dpaLink: (chunks) => (
```
Comment (benjiwagner) [inline app/[locale]/(protected)/onboarding/wizard/components/step-profile.tsx]: Image

As you can see, the star breaks weirdly and also the link should instead be with underscore and the same color. This should be default for inline links. I think we mixed this currently in many cases. Can you unify this?

app/layout.tsx
```diff
@@ -193,12 +192,6 @@ export default async function RootLayout({ children }: Props) {
         </Providers>
 
         <Analytics />
-
-        <Script id="lemon-squeezy-affiliate-config" strategy="afterInteractive">
```
benjiwagner, line 204: Are you sure that the affiliates still work when moving this script and layout? Please double-check the best practive for lemonsquezzy and nextjs.

core/fumadocs/youtube-embed.tsx
```diff
@@ -1,21 +1,43 @@
...
+import { useState } from "react";
+import { PlayIcon } from "lucide-react";
+import { useTranslations } from "next-intl";
+
 type Props = {
   id: string;
   title?: string;
 };
 
 export function YouTube({ id, title = "YouTube video" }: Props) {
```
benjiwagner, line 8: Have we even used this? In case we did not use this before, you can simply remove it completely.

ee/subscription/create-checkout-session.interactor.ts
```diff
@@ -57,6 +58,8 @@ export class CreateCheckoutSessionInteractor {
       redirectUrl,
     });
 
-    return redirectTo(checkout.data.attributes.url);
+    const referral = (await cookies()).get(AFFILIATE_REFERRAL_COOKIE)?.value ?? null;
```
benjiwagner: Again, since I'm not quite sure whether this is the best practice for the affiliate referral for Lemon Squeezy, please deeply check whether the script in the layout is not the correct way to do it.

benjiwagner:
Went through all five comments plus Jul's review. Pushed as `1f6872f`.

**Affiliate tracking: you were right, I was wrong.** I checked the Lemon Squeezy docs properly this time ([Getting Referrals](https://docs.lemonsqueezy.com/help/affiliates-for-merchants/getting-referrals), [custom landing pages](https://docs.lemonsqueezy.com/guides/tutorials/affiliate-landing-pages)). Two things break with my change:

1. The script is what creates the referral. "Once the script is on your page, a visit to a page with the affiliate code in the URL (e.g. `?aff=1234`) will create an anonymous unique identifier" and that identifier lives in Lemon Squeezy's system. No script on the landing page means no referral is ever created, so removing it would have silently killed attribution.
2. `aff_ref` is not the affiliate code. The docs say to retrieve it with `LemonSqueezy.Affiliate.GetId()`, which only returns anything because the script ran. I was passing the raw `?aff=` value, which is a different thing.

So the script is back in the root layout exactly as it was, and `core/affiliate`, the middleware capture and the checkout change are all reverted. As a side effect the privacy policy is accurate again: it describes "an anonymous identifier generated by Lemon Squeezy", which only matches the script-based flow.

The trade-off this reopens: a US third-party script now loads on every page again, including inside the logged-in CRM. It is disclosed under Art. 6(1)(f) with a consent fallback, and Lemon Squeezy states the tracking is cookieless. If you later want it off the authenticated app, the safe version is to scope it to the public marketing layouts rather than remove it, since affiliate links land there. Happy to do that as a follow-up, but I did not want to guess at it in the same PR that fixes the legal text.

**YouTube embed: removed entirely.** You were right that it was unused. It was registered in `mdx-components.tsx` but referenced by zero MDX files, so the facade guarded nothing. Component, registration and the i18n key are gone.

**Connect disclosure: merged into the alert** and both halves shortened, so the card carries one message.

**Inline links: unified.** `AppLink` now has an `inline` variant that renders in the surrounding text colour with a permanent underline. Sign-in and sign-up were hand-rolling `className="text-inherit underline"`; the onboarding checkbox had no styling at all, so it rendered primary-coloured without an underline. All three now use the same variant.

**The asterisk:** it was a separate `<span> *</span>` with a normal leading space, so it wrapped onto its own line whenever the label ran to multiple lines. Bound to the preceding word with a non-breaking space, applied across all twelve form components rather than just this one.

**@eljulsi on the change notification:** agreed, and it is already listed as a pre-publication blocker in the PR description. Worth noting the schema change here only makes acceptance binding going forward: existing users keep `NULL` rather than a backfilled timestamp, precisely so we do not manufacture evidence that a document was shown when it was not. So a notice to existing users is the mechanism that closes that gap, not the migration. Sequencing and wording are Benji's call.

Verification: `typecheck`, `lint`, `test` (909 tests, 114 files) and `build` all pass.

app/[locale]/(protected)/profile/components/connected-accounts-card.tsx
```diff
@@ -92,6 +93,33 @@ const ConnectAction = observer(() => {
   );
 });
 
+const ConnectedAccountsAlert = () => {
```
Comment (benjiwagner) [inline app/[locale]/(protected)/profile/components/connected-accounts-card.tsx:96]: Image

This is not properly displayed.

features/user/prisma-user.repository.ts
```diff
@@ -866,6 +866,10 @@ export class PrismaUserRepo
     const user = await this.prisma.user.create({
       data: {
         agreeToTerms: args.agreeToTerms,
+        legalAcceptedAt: args.legalAcceptedAt,
```
benjiwagner: What about already existing users?

scripts/stamp-legal-date.ts
```diff
@@ -0,0 +1,103 @@
+import { readFileSync, writeFileSync } from "node:fs";
+import { join } from "node:path";
+
+const LOCALES = ["en", "de"] as const;
+const DOCUMENTS = ["privacy", "terms", "dpa", "subprocessors"] as const;
+
+const MONTHS_EN = [
```
benjiwagner: This should be part of the translation json files

package.json
```diff
@@ -173,6 +173,7 @@
     "db:use-live-data": "bash scripts/use-live-data.sh",
     "dev": "prisma generate && yarn workflow:setup && yarn openapi:generate && yarn raw-docs:generate && next dev --turbopack -p 4000",
     "email:dev": "email dev --dir components/emails",
+    "legal:date": "tsx scripts/stamp-legal-date.ts",
```
benjiwagner: Explain this again to me pls

app/[locale]/(protected)/profile/components/connected-accounts-card.tsx
Comment (benjiwagner) [reply app/[locale]/(protected)/profile/components/connected-accounts-card.tsx:96]: Generally I think the links should always be displayed in the inherited color and textsize and with an underscore. Can you fix this globally?

features/user/prisma-user.repository.ts
benjiwagner: Deliberately nothing, and that is the point.

Existing users keep `NULL` in all four columns. Their evidence of acceptance is the `agreeToTerms` checkbox they ticked plus their `createdAt`. If I backfilled `legalAcceptedAt` with a timestamp, I would be manufacturing a record that a specific version of the DPA was shown to them on a specific date, when it was not: the DPA did not exist when they registered. A fabricated acceptance record is worse evidence than an honest gap, because it is the kind of thing that falls apart the moment anyone checks.

So the migration only makes acceptance binding **going forward**. `NULL` means "registered before the registry existed", which is true and checkable.

The mechanism that would close the gap for existing users is a notice, not a migration. You have since decided that is not needed: your current users are family and friends, and the change-notification work is cancelled in [CUS-158](https://linear.app/customermates/issue/CUS-158) with the reasoning recorded there. So this stays as it is.

One consequence worth knowing: nothing in the app ever reads these columns back. They are write-only evidence today, with no re-acceptance check anywhere. That is fine for their purpose, but it means a version bump has no runtime effect.

package.json
benjiwagner: `yarn legal:date 2026-08-20` writes the publication date into the legal documents.

It exists because the date is not one string. There are **14 date lines across 8 files in 5 different formats and 2 languages**:

```
_Version 2026-08-03_                              terms, both locales
_Version 2026-08-03, last updated 3 August 2026_  privacy EN
_Version 2026-08-03, Stand 3. August 2026_        privacy DE
_Version 1.0, effective 3 August 2026_            dpa EN
_Version 1.0, gültig ab 3. August 2026_           dpa DE
_Last updated: 3 August 2026_ / _Stand: …_        subprocessors
*Last Update: 03.08.2026* / *Stand: 03.08.2026*   footers
```

Editing those by hand on publication day is how you end up with a German privacy policy dated the 20th and an English one dated the 3rd, or a footer contradicting the version line above it. On a legal document that is not cosmetic.

The second reason matters more. The version string is **written into the database** at registration: `buildLegalAcceptance()` stamps `legalTermsVersion: "2026-08-03"` onto the user row. If the document and the constant drift apart, your acceptance evidence names a version that never existed.

**A known gap, which is why [CUS-157](https://linear.app/customermates/issue/CUS-157) exists.** The script does not update `constants/legal-documents.ts`, and it re-dates all four documents on every run, so fixing a typo in the subprocessor list silently re-dates the terms. The coupling test catches the first problem loudly but only after the fact. CUS-157 makes it one command per document that updates both, and keeps a copy of the superseded text, which today does not exist anywhere.

Not in this PR because it edits the same files, and I did not want versioning plumbing in the PR a lawyer is about to read.

app/[locale]/(protected)/profile/components/connected-accounts-card.tsx
Comment (benjiwagner) [reply app/[locale]/(protected)/profile/components/connected-accounts-card.tsx:96]: The alert was breaking because of a layout bug, not the text.

`AlertDescription` in `components/ui/alert.tsx` is `grid justify-items-start`. When the description is a plain string that is fine, one grid item. But this one comes from `t.rich()`, which returns an **array** of nodes: text fragment, link, text fragment, link, and so on. The grid then treats each fragment as its own row, so the sentence stacks vertically instead of flowing.

Fixed in `components/shared/alert.tsx` by wrapping the description in a single paragraph, so the grid gets one child and the text flows normally. That also picks up the `[&_p]:leading-relaxed` rule the primitive already had, which suggests a paragraph was the intended child all along.

This affects every Alert with rich text, not just this one.

Comment (benjiwagner) [reply app/[locale]/(protected)/profile/components/connected-accounts-card.tsx:96]: Fixed, but not by changing the default, and I want to be straight about why.

There are 64 `AppLink` call sites. Only 12 are links in running text. The other 52 are **buttons**: hero CTAs, pricing cards, `Get started`. Making inherited-colour-plus-underline the global default would put an underline through every one of those.

So the rule is applied by construction where it belongs, not remembered per call site:

1. **Containers style their own links.** `Alert` already does this with `[&_a]:text-inherit [&_a]:underline`, which is why links inside an alert are correct regardless of what the call site passes. That is the pattern.
2. **`inline` now implies `inheritSize`.** They were two separate props and had drifted apart, which is how the onboarding checkbox ended up primary-coloured with no underline while sign-in was hand-rolling `className="text-inherit underline"`. One prop now does both.

If you do want it as the hard default everywhere, the honest version is a `variant` prop with `text` as the default and `action` for the button-like ones, plus updating 52 call sites. Happy to do that as its own change, but not inside the PR carrying the legal text.

scripts/stamp-legal-date.ts
benjiwagner: Removed them, but not into the translation files. `Intl` already knows the month names, so nothing needs to be maintained anywhere:

```ts
const LONG_DATE = {
  en: new Intl.DateTimeFormat("en-GB", { day: "numeric", month: "long", year: "numeric", timeZone: "UTC" }),
  de: new Intl.DateTimeFormat("de-DE", { day: "numeric", month: "long", year: "numeric", timeZone: "UTC" }),
} as const;
```

Two reasons for this over `i18n/locales/*.json`:

1. **The catalogues are product copy a translator owns.** Calendar month names are not copy, they are platform data. Putting them there would invite someone to "improve" them and silently change what a legal document's date renders as.
2. **This is a build script, not the app.** It runs under `tsx` outside Next, so pulling in the next-intl runtime to format two dates would be more machinery than the problem deserves.

`timeZone: "UTC"` matters: without it a date near midnight renders as the previous day for anyone west of Greenwich, which on a version line is exactly the kind of off-by-one nobody notices until it matters.

Seven tests on the formatter still pass, including the German long form and the numeric footer format.

benjiwagner:
Final owner decision is now recorded in the PR and CUS-56: this publication is prospective only, existing customers remain on their prior versions, and no existing-user notice is sent in this release. The unrelated flags/badges/UI/YouTube/third-party-request changes have been removed; the final PR diff is the 36-file legal, licensing, route, and acceptance-evidence scope. Local full tests (1,049), lint, typecheck, build, database migration/acceptance, route, and visual checks pass. Please re-review for approval once the fresh GitHub checks complete.

benjiwagner:
@eljulsi Your earlier concern about notifying existing customers is now implemented on the current head: the daily lifecycle cron sends audited legal notices, the UI starts only after successful email evidence, and one authorised administrator accepts Terms/DPA company-wide before the 14-day deadline (with a post-deadline UI gate; APIs remain unchanged). All fresh CI, CodeQL, repository-policy, and Vercel checks are green. A fresh review is requested.

README.md
```diff
@@ -153,6 +153,10 @@ Start here: [customermates.com/docs](https://customermates.com/docs)
 
 Customermates uses an open-core licensing model.
 
-The community edition is licensed under [AGPLv3](./LICENSE). Files in `ee/` are subject to the commercial terms in [`ee/LICENSE.md`](./ee/LICENSE.md).
+The Community Edition (first-party code outside `ee/` plus `ee/audit-log/`) is licensed under [AGPL-3.0-only](./LICENSE). Audit logging is therefore free for production self-hosting.
```
benjiwagner: I think we should simplify this by moving the audit log out of the enterprise folder, the EE folder, if it is not part of it.

app/[locale]/(protected)/legal-update/actions.ts
```diff
@@ -0,0 +1,20 @@
...
+import type { AcceptLegalDocumentsData } from "@/features/legal/accept-legal-documents.interactor";
+
+import { redirect, RedirectType } from "next/navigation";
+import { revalidatePath } from "next/cache";
+import { getLocale } from "next-intl/server";
+
+import { getAcceptLegalDocumentsInteractor } from "@/core/di";
+import { serializeResult } from "@/core/utils/action-result";
+
+export async function acceptLegalDocumentsAction(data: AcceptLegalDocumentsData) {
```
Comment (benjiwagner) [inline app/[locale]/(protected)/legal-update/actions.ts:11]: The action looks to be defined differently than usually because revalidate path as well as passing the current locus, not what we usually do. Please don't invent new patterns.

app/[locale]/(protected)/onboarding/wizard/actions.ts
```diff
@@ -4,13 +4,15 @@ import type { RegisterUserData } from "@/features/user/register/register-user.in
...
+import { getLocale } from "next-intl/server";
 import { Status } from "@/generated/prisma";
 
 import { getCompleteOnboardingWizardInteractor, getRegisterUserInteractor, getUserService } from "@/core/di";
 import { serializeResult } from "@/core/utils/action-result";
 
 export async function registerProfileAction(data: RegisterUserData) {
-  const result = await serializeResult(getRegisterUserInteractor().invoke(data));
+  const legalLocale = (await getLocale()) === "de" ? "de" : "en";
+  const result = await serializeResult(getRegisterUserInteractor().invoke({ ...data, legalLocale }));
```
Comment (benjiwagner) [inline app/[locale]/(protected)/onboarding/wizard/actions.ts]: Also, I don't quite get why we need to pass the local over here as well.

app/components/legal-update-banner.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+import { Alert } from "@/components/shared/alert";
+import { AppLink } from "@/components/shared/app-link";
+import { usePathname } from "@/i18n/navigation";
+import { isPublicPathname } from "@/i18n/routing";
+
+type Props = { status: LegalUpdateStatus };
+
+export function LegalUpdateBanner({ status }: Props) {
+  const t = useTranslations("LegalUpdateBanner");
```
benjiwagner: I think usually we use the global translation and don't pass a specific one. Please double-check all the other occurrences, and if you agree with me, I want to add a linter setting so that this doesn't happen again in the future.

app/components/legal-update-banner.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const format = useFormatter();
+  const locale = useLocale();
+  const pathname = usePathname();
+  const isPublicRoute = isPublicPathname(pathname);
+
+  useEffect(() => {
```
benjiwagner: Is this useEffect content the common pattern we use? Because it looks quite complicated and also like something we usually don't do.

features/legal/legal-status.service.ts
```diff
@@ -0,0 +1,128 @@
...
+  return eventData.payload as LegalAcceptanceAuditPayload;
+}
+
+function dateFromIso(value: string | null | undefined): Date | null {
+  if (!value) return null;
+  const date = new Date(value);
+  return Number.isNaN(date.getTime()) ? null : date;
+}
+
+export class LegalStatusService {
```
benjiwagner: I think we should rework this to be a proper interactor instead. Like all the other interactors, this should not be a service.

features/email/email.service.ts
```diff
@@ -9,12 +9,15 @@ type SendArgs = {
   subject: string;
   react: React.ReactElement<Record<string, unknown>>;
   from?: string;
+  idempotencyKey?: string;
```
benjiwagner: We did not need the idempotency key before, so instead we relied on our lifecycle interactions to only execute emails in the business logic if needed. So I think you should not invent new patterns instead of adding the idempotency key.

README.md
benjiwagner: Addressed in 77b37dd6: moved the four audit-log files from ee/audit-log to features/audit-log, updated every import plus the root and EE licence documentation, and added convention coverage for the boundary. Both managed-cloud and self-hosted production builds pass.

app/[locale]/(protected)/legal-update/actions.ts
Comment (benjiwagner) [reply app/[locale]/(protected)/legal-update/actions.ts:11]: Addressed in 77b37dd6: the action now passes the form data unchanged, derives the active request locale inside the interactor, calls the narrow Server Action refresh(), and then uses a normal redirect("/"). Browser verification confirms acceptance clears the banner without a hard reload and retains the German locale.

app/[locale]/(protected)/onboarding/wizard/actions.ts
Comment (benjiwagner) [reply app/[locale]/(protected)/onboarding/wizard/actions.ts]: Addressed in 77b37dd6: onboarding no longer passes a legalLocale field through the action or schema. RegisterUserInteractor resolves the active request locale itself. A real German browser signup recorded locale=de in the initial legal acceptance event.

app/components/legal-update-banner.tsx
benjiwagner: Addressed in 77b37dd6: all remaining namespaced client useTranslations calls were converted to global useTranslations() with full keys, including the connected-account toast. ESLint now rejects future namespaced calls, and the full lint plus i18n convention suite passes.

app/components/legal-update-banner.tsx
benjiwagner: Addressed in 77b37dd6: the effect is now one validated deadline timer that calls the established router.refresh(); the server route guard remains the redirect authority. Local multi-user browser verification confirmed an already-open admin and member tab both enter /legal-update exactly when the deadline passes.

features/legal/legal-status.service.ts
benjiwagner: Addressed in 77b37dd6: LegalStatusService is now the conventional GetLegalStatusInteractor with invoke(user, now), with DI, route guard, layouts, pages, imports, and tests updated.

features/email/email.service.ts
benjiwagner: Retained intentionally after review. AuditLog remains the primary business deduplication check, but removing the deterministic Resend key leaves concurrent-run and provider-success-before-audit-persistence duplicate windows. Reusing the existing claim-before-send lifecycle pattern would suppress the required daily retry after provider failure. The provider key is therefore a narrow delivery safeguard, and the retry test now proves that a failed recipient reuses the exact same key. The full failure/retry database journey also passes.

app/[locale]/(protected)/onboarding/wizard/components/step-profile.tsx
```diff
@@ -47,26 +48,33 @@ export const StepProfile = observer(({ email, firstName, lastName, avatarUrl }:
...
-                termsOfServiceLink: (chunks) => (
-                  <AppLink appearance="inline" href="/terms" target="_blank">
-                    {chunks}
-                  </AppLink>
-                ),
-              })}
-            </span>
-          }
-        />
+        {appMode === "cloud" && !isInvited ? (
```
Comment (benjiwagner) [inline app/[locale]/(protected)/onboarding/wizard/components/step-profile.tsx]: I think it doesn't hurt to show it for invited users as well again.

app/[locale]/(protected)/layout.tsx
```diff
@@ -27,11 +27,15 @@ import { TimelineDetailModal } from "@/features/messaging/activities/activities-
 
 export default function ProtectedLayout({ children }: { children: React.ReactNode }) {
   const pathname = usePathname();
+  const isLegalUpdateRoute =
```
Comment (benjiwagner) [inline app/[locale]/(protected)/layout.tsx]: I think we can simplify this here as well for inactive users, we also don't guard the models or the loading in the UI because essentially we are a backend-driven monolith.

constants/legal-documents.ts
```diff
@@ -0,0 +1,38 @@
+export const LEGAL_DOCUMENT_VERSIONS = {
+  dpa: "2026-08-07",
+  privacy: "2026-08-07",
+  subprocessors: "2026-08-07",
+  terms: "2026-08-07",
+} as const;
+
+// For a supplier-originated subprocessor change, set this to the supplier's
```
benjiwagner: We usually don't use comments because we prefer descriptive code instead.

ee/lifecycle/send-legal-document-notices.interactor.ts
```diff
@@ -0,0 +1,313 @@
...
+
+  return documents.filter((document) => {
+    const baseline = informationBaselineVersion(records, recipient.id, document);
+    if (baseline !== null) return baseline !== releaseDateFor(document);
+    return userExistedAtRelease(recipient, document);
+  });
+}
+
+@SystemInteractor
+export class SendLegalDocumentNoticesInteractor {
```
benjiwagner, line 54: Please deeply compare this interactor and all its functions also outside of the interactor class within this file with the existing patterns and other interactors because it looks quite different. Therefore, refactor it and also double-check all the other changes you made in this pull request so they align extremely well with the existing patterns and architecture. Do this in depth.

features/email/email.service.ts
```diff
@@ -30,11 +30,13 @@ export class EmailService {
...
 
-    await resend.emails.send({
+    const { error } = await resend.emails.send({
       from: args.from ?? defaultSender,
       to: args.to,
       subject: args.subject,
       react: args.react,
     });
+
+    if (error) throw new Error(`Resend rejected the email: ${error.message}`);
```
benjiwagner: What was happening before, we were checking for the error return field manually, but we just decided to dropping the errors. So, now it looks like a new behavior in the lifecycle methods or generally speaking also in the email service.

app/[locale]/(protected)/legal-update/components/legal-update-view.tsx
```diff
@@ -0,0 +1,115 @@
...
+import { Button } from "@/components/ui/button";
+import { Checkbox } from "@/components/ui/checkbox";
+import { Label } from "@/components/ui/label";
+import { signOutAction } from "@/app/[locale]/actions";
+
+import { acceptLegalDocumentsAction } from "../actions";
+
+type Props = { status: LegalUpdateStatus };
+
+export function LegalUpdateView({ status }: Props) {
```
Comment (benjiwagner) [inline app/[locale]/(protected)/legal-update/components/legal-update-view.tsx]: Don't we usually have a store for the state management and also we usually don't have to use transition, I think. If that is actually true, make sure to create also a linter rule so we don't cause this in the future.

app/[locale]/(protected)/profile/components/connected-accounts-card.tsx
Comment (benjiwagner) [reply app/[locale]/(protected)/profile/components/connected-accounts-card.tsx:96]: Image

The links are still not properly inlined so there are weird breaks

app/layout.tsx
```diff
@@ -149,20 +152,28 @@ export default async function RootLayout({ children }: Props) {
...
-          getGetMyConnectedAccountsInteractor().invoke(),
-        ]);
+      const [
+        companyResult,
+        systemTaskCountResult,
+        subscriptionResult,
+        unreadThreadCountResult,
+        accountsResult,
+        legalResult,
+      ] = await Promise.all([
```
benjiwagner, line 162: Quick question. Are all of those promises loaded on every request in the user interface? And if so, does it make sense just to create one interactor which specifically gets this all in one, but only really do it if there's a really good performance boost?

features/legal/accept-legal-documents.interactor.ts
```diff
@@ -0,0 +1,77 @@
...
+  ) {
+    super();
+  }
+
+  @Write({ input: Schema, output: Schema })
+  async invoke(data: AcceptLegalDocumentsData): Validated<AcceptLegalDocumentsData> {
+    if (env.APP_MODE !== "cloud" || !this.user.role?.isSystemRole)
+      throw new ForbiddenError("Only a managed-cloud system administrator may accept legal documents");
+
+    const locale = (await getLocale()) === "de" ? "de" : "en";
```
benjiwagner: Is it actually important which locale the user is accepting? Because generally I think only one language should be the main language. I'm also not quite sure what's the legal obligation behind it and how, for example, if we add multiple different languages, we only have one contract in terms when there may be drift across the locales.

features/legal/get-legal-status.interactor.ts
```diff
@@ -0,0 +1,135 @@
...
+      contractAccepted,
+      contractNoticeSent,
+      effectiveAt: contractAccepted ? null : (contractEffectiveAt?.toISOString() ?? null),
+      informationNoticeVisible,
+      isSystemAdministrator,
+      mustAccept,
+    };
+  }
+
+  private includesAny(
```
benjiwagner: I think there is super small methods don't really add readability. Instead, they just add another level of abstraction, which is bad for the reader. Please double-check this across the whole files touched in this pull request.

features/legal/legal-audit.repo.ts
```diff
@@ -0,0 +1,65 @@
+import type { LegalAcceptanceAuditPayload, LegalNoticeAuditPayload } from "@/constants/legal-documents";
```
benjiwagner: This file is a bit weird because it does not really comply with the existing patterns, because it's a .repo.ts file, which is also including schema files as well as an abstract repository, and so on and so forth. And it's having small helper methods as well, which is quite weird for parsing schema, so don't invent new patterns.

features/user/register/register-user.interactor.ts
```diff
@@ -18,24 +22,23 @@ import { CustomErrorCode } from "@/core/validation/validation.types";
...
-  });
+const Schema = z.object({
+  email: z.email(),
+  firstName: z.string().min(1),
+  lastName: z.string().min(1),
+  country: z.enum(CountryCode),
+  avatarUrl: zx.secureUrl().or(z.literal("")).nullable(),
+  agreeToTerms: z.boolean(),
+});
+const NewCloudCompanySchema = Schema.superRefine((data, ctx) => {
```
benjiwagner: This looks like a pattern we dont really have anywhere else? Fix it

app/[locale]/(protected)/onboarding/wizard/components/step-profile.tsx
Comment (benjiwagner) [reply app/[locale]/(protected)/onboarding/wizard/components/step-profile.tsx]: Onboarding invites users should still ahve to accept the terms with the checkbox initially. Just for double security. Like before.

Comment (benjiwagner) [reply app/[locale]/(protected)/onboarding/wizard/components/step-profile.tsx]: Addressed in `476cbbc1` (implementation in `d9b2bd7a`). Every managed-cloud onboarding submission now requires `agreeToTerms=true` through the single cloud-registration schema. Creating a new company records the company-wide acceptance atomically; an invited user persists only the existing `User.agreeToTerms` acknowledgement and emits no `legal.documents_accepted` event, so an invitee cannot unlock or replace the company state. Demo and self-hosted registrations remain exempt. Unit and disposable-real-database registration journeys cover all four paths.

app/[locale]/(protected)/layout.tsx
Comment (benjiwagner) [reply app/[locale]/(protected)/layout.tsx]: Confirmed on current head `476cbbc1`: this old diff is already gone. The protected layout has no inactive-user-specific modal/model/loading guard added by this PR; the established backend-driven route and tenant checks remain authoritative. No replacement client guard or new abstraction was introduced.

constants/legal-documents.ts
benjiwagner: Confirmed on current head `476cbbc1`: the explanatory comments were already removed. `constants/legal-documents.ts` uses descriptive constant names only, and the focused legal convention tests enforce exact constant/header and EN/DE version parity.

ee/lifecycle/send-legal-document-notices.interactor.ts
benjiwagner: Addressed in `d9b2bd7a` and completed in `476cbbc1`. The interactor again exposes `invoke(): Promise<void>`, has no production test-only time/deadline arguments, uses normal system time plus module/fake-time mocks in tests, and keeps only helpers that encapsulate repeated legal-version semantics. Paired baseline/change logic was consolidated and trivial one-use wrappers were inlined. It still reuses the existing lifecycle cron, EmailService, EventService, AuditLog, and fail-fast sequencing; no table, migration, index, cache, or parallel delivery architecture was added. The focused lifecycle suite and full 1,262-test suite pass.

features/email/email.service.ts
benjiwagner: Confirmed on current head `476cbbc1`: existing EmailService callers retain the prior best-effort treatment of a Resend provider error. Only the legal-notice sender passes `{ throwOnProviderError: true }`, because its audit success must be written only after Resend accepts the delivery. This does not change ordinary lifecycle/email behavior. Tests cover both the unchanged default and the strict legal-only path, including failure, no audit record, fail-fast, and next-run retry.

app/[locale]/(protected)/legal-update/components/legal-update-view.tsx
Comment (benjiwagner) [reply app/[locale]/(protected)/legal-update/components/legal-update-view.tsx]: Addressed in `d9b2bd7a`. `LegalUpdateView` now delegates checkbox state, loading, acceptance, sign-out, and established validation-error toasts to a small `LegalUpdateStore`, registered lazily through the existing RootStore pattern. `useTransition` and direct action state were removed. I deliberately did not add a global lint rule: transitions are valid elsewhere, while the architectural consistency here is protected by the focused store/view tests.

app/layout.tsx
benjiwagner: Yes, the authenticated root layout loads this shell data on each relevant UI request, but all of these reads already run concurrently in the existing `Promise.all`; the legal lookup adds no serial waterfall. An aggregate interactor would call the same repositories and produce the same query count/latency while adding indirection, and a request cache would add a new invalidation pattern for one small indexed AuditLog read. I therefore kept the existing concurrent fan-out and documented that decision in PR #30 and CUS-56.

features/legal/accept-legal-documents.interactor.ts
benjiwagner: Addressed in `d9b2bd7a`. Locale was removed from notice and acceptance payloads, Zod parsers, events, tests, and the EN/DE Privacy audit disclosure. Emails and pages still render in the user's display language, but the legal evidence is the shared explicit document versions. The German Terms/DPA remain authoritative and English is the convenience translation, so adding future presentation languages does not create separate contracts.

features/legal/get-legal-status.interactor.ts
benjiwagner: Addressed across the touched legal files in `d9b2bd7a`, with the remaining one-use deadline helper inlined in `476cbbc1`. The status interactor now performs its small deadline reduction at the call site; the lifecycle sender consolidates paired baseline/change logic and inlines trivial wrappers while retaining named helpers only where they carry repeated legal-version/deadline semantics. Focused status/lifecycle tests, lint, and typecheck pass.

features/legal/legal-audit.repo.ts
benjiwagner: Addressed in `d9b2bd7a`. `legal-audit.repo.ts` is now the repository port only. Zod schemas, parsed audit types, validation, and the deadline predicate live in the established `legal-audit.schema.ts` pattern, with the former repository test renamed/refocused on schema parsing. Repository-boundary validation and malformed-payload coverage remain intact.

features/user/register/register-user.interactor.ts
benjiwagner: Addressed in `d9b2bd7a`. Registration now has one cloud-aware schema validation for `agreeToTerms=true`; the interactor preserves the submitted value instead of reparsing or rewriting it. New-company acceptance stays in the existing registration transaction, invitees store only their personal acknowledgement, and demo/self-hosted paths remain exempt. Unit and disposable-real-database tests prove the four variants and rollback behavior.

app/[locale]/(protected)/profile/components/connected-accounts-card.tsx
Comment (benjiwagner) [reply app/[locale]/(protected)/profile/components/connected-accounts-card.tsx:96]: The newer inline-flow comment in this already-resolved thread is addressed in `d9b2bd7a`. Shared `Alert` now wraps the entire description in one paragraph, so a `t.rich()` array is one grid child and all text/link fragments flow inline; the action remains a separate grid cell. Focused regression tests assert that multiple rich links share the same paragraph, inherit alert styling, and do not become separate rows. The full suite and desktop/mobile browser verification pass.

benjiwagner:
@eljulsi Your existing-user notification concern is now implemented on the current head `476cbbc1`.

The existing daily lifecycle cron compares each recipient/company's audited document versions with the four current constants. It sends:

- one combined Terms/DPA/Privacy/Subprocessor email to each active system administrator for the documents relevant to that recipient;
- a Privacy email to other active users;
- a Subprocessor objection notice to administrators.

A `legal.notice_sent` audit event is persisted only after successful Resend delivery, failures remain retryable, and one authorised administrator's current Terms/DPA acceptance restores the company-wide UI. The 14-day clock starts from successful administrator notice evidence. Disposable real-database multi-user journeys, provider failure/retry tests, the full 1,262-test suite, and desktop/mobile browser verification pass.

The PR is still draft while fresh CI/Vercel completes. After it is green, your fresh Code Owner review is the remaining approval checkpoint.

app/[locale]/(protected)/legal-update/page.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+import { LegalUpdateView } from "./components/legal-update-view";
+
+export default async function LegalUpdatePage() {
+  await requireAccess({
+    skipLegalAcceptanceCheck: true,
+    skipSubscriptionCheck: true,
+  });
+  const user = await getUserService().getActiveUserOrThrow();
+  const status = await getGetLegalStatusInteractor().invoke(user);
```
Comment (benjiwagner) [inline app/[locale]/(protected)/legal-update/page.tsx]: Why do we need to pass the user to the interactor? Because we have other interactors that just implement a user accessor to get the current tenant user. Isn't this way simpler? Once you correct this, also double-check the other interactors on this missing behavior.

features/email/email.service.ts
```diff
@@ -11,10 +11,14 @@ type SendArgs = {
   from?: string;
 };
 
+type SendOptions = {
+  throwOnProviderError?: boolean;
```
benjiwagner, line 15: If we never use this option or property, then simply remove it completely. We don't want to add something in this pull request just for the sake of adding it.

features/legal/legal-audit.schema.ts
```diff
@@ -0,0 +1,61 @@
...
+  | (LegalAuditRecordBase & {
+      event: DomainEvent.LEGAL_NOTICE_SENT;
+      payload: LegalNoticeAuditPayload | null;
+    })
+  | (LegalAuditRecordBase & {
+      event: DomainEvent.LEGAL_DOCUMENTS_ACCEPTED;
+      payload: LegalAcceptanceAuditPayload | null;
+    });
+
+export function parseLegalNoticeAuditPayload(value: unknown): LegalNoticeAuditPayload | null {
```
benjiwagner: I don't get why we have this parse methods. We usually don't have them defined in the schema.ts files or use helper wrapper functions for those.

app/[locale]/(protected)/legal-update/page.tsx
Comment (benjiwagner) [reply app/[locale]/(protected)/legal-update/page.tsx]: Fixed in `5002fe0c`. `GetLegalStatusInteractor` now follows the established `@TenantInteractor()` plus `UserAccessor` pattern and exposes a no-argument `invoke()`. The legal page, route guard, and root layout all use that contract, and the page no longer reloads the active user. `@AllowInDemoMode` preserves the root-layout demo path while the interactor still returns the exempt state outside managed cloud. I audited every interactor and production/test call site touched by this PR: acceptance already uses `AuthenticatedInteractor`; registration and lifecycle remain intentionally system-scoped. Tests now lock the accessor-provided company scope, zero-argument route-guard call, demo/self-hosted exemptions, and administrator/member context switching. The five real-database journeys, 60 focused tests, and full 1,263-test suite pass.

features/email/email.service.ts
benjiwagner: I checked the premise across every production caller before changing this. The option is used by exactly one of the 14 `EmailService.send()` call sites: the legal-notice sender passes `{ throwOnProviderError: true }`. That call must reject before `legal.notice_sent` is published. Removing strict handling would allow a Resend provider error to resolve, record a false successful delivery, suppress the next cron retry, and incorrectly start the legal deadline. The other 13 callers intentionally retain the existing best-effort behavior, so making all mail globally strict would broaden this PR into unrelated lifecycle/auth flows. I therefore kept the small opt-in rather than remove a required safety control. Focused EmailService and legal-lifecycle coverage passes 24/24, including provider failure, no false audit, fail-fast behavior, and retry.

features/legal/legal-audit.schema.ts
benjiwagner: Fixed in `5002fe0c` by exporting the two Zod schemas and validating directly in `PrismaAuditLogRepo` at the persistence boundary. Both schema-file parse wrappers are removed. Malformed historical audit JSON still maps to `null`, so it cannot count as legal evidence or break the cron/layout. I scanned all 35 schema files and every schema/repository file touched by this PR; no other production parse-wrapper instance remains. Focused schema tests now exercise the schemas directly, while downstream tests retain malformed-payload and fail-closed coverage.

features/email/email.service.ts
benjiwagner: Why does this lifecycle method do this? It looks like a new pattern.

PR #36, opened 2026-08-03
feat(i18n): decouple application locales from public content locales, add three languages

benjiwagner:
## Summary

Two things, in order. First, `ROUTING_LOCALES = ["en","de"]` drove five unrelated concerns: the next-intl message catalog, the Fumadocs MDX content tree, sitemap and hreflang, the middleware's URL-prefix recognition, and, duplicated by hand, the Prisma display-language enum. Adding an application UI language therefore demanded a full public/SEO site in that language. Second, **French, Italian and Spanish are added as application-only languages**, which is what proves the split works: three new product languages, zero new public URLs.

`i18n/locale-registry.ts` is now the only place a locale is declared, and every consumer derives from the domain it belongs to:

| Domain | Meaning | Drives |
| --- | --- | --- |
| `APP_LOCALES` | offered as a persisted display language | profile language select, `User.displayLanguage` validation, background and cron emails |
| `CONTENT_LOCALES` | has published MDX under `content/` | Fumadocs resolution, `sitemap.xml`, hreflang, footer switcher, hero images, raw-docs manifest, MCP docs tools |
| `ROUTING_LOCALES` | the union | URL prefixes the proxy recognises, message loading, prefix stripping |

`tests/conventions/locale-consumer-audit.test.ts` fails on any locale list, union type or alias declared outside the registry, so this cannot quietly re-couple. `i18n/README.md` records the JSON-versus-MDX contract and both locale-addition procedures, including the files that must **not** change.

**Reviewer shortcut:** the split is mostly a one-to-five-line repoint per file, and the only thing to check per file is the **domain column** above. The three genuine rewrites are `app/sitemap.ts`, `core/fumadocs/metadata.ts`, and the content half of `tests/conventions/i18n-parity.test.ts`. The three language commits are separate and are almost entirely `i18n/locales/{fr,it,es}.json`.

## Context

Planning surfaced three live defects sitting directly on this path. Shipping the split on top of them would have made each worse, so they are fixed here.

1. **Permanently cached locale redirects.** `proxy.ts` rewrote next-intl's 307 locale redirects to **308**, while the target depends on the `NEXT_LOCALE` cookie and `Accept-Language` and carried no `Vary`. Unknown prefixes were 308'd to `/en/<prefix>/...`, so `/fr/*` was already being poisoned in browser caches for a locale that did not exist yet. next-intl emits 307 deliberately; the rewrite was fighting the library.
2. **Two contradictory `x-default` values.** next-intl's `alternateLinks` defaults on and emitted a second hreflang set through the HTTP `Link:` header whose `x-default` was **unprefixed** while the HTML said `/en/...`. `alternateLinks` is now off; HTML metadata is the single emitter.
3. **No `params.locale` validation anywhere.** There was no `app/[locale]/layout.tsx`, so an unsupported prefix could render the English homepage at **200**. The layout now applies next-intl's documented `hasLocale` guard.

Locale negotiation stays **enabled**, scoped by building a second `createMiddleware` instance from the content locales, which is next-intl's documented dynamic-configuration composition. An application-only locale is therefore structurally incapable of becoming an `Accept-Language` target, with no hand-rolled negotiation to maintain. Fumadocs runs with `fallbackLanguage: null`, a first-class documented option; without it a missing translation silently served default-language prose under a localized URL that still carried a localized canonical and hreflang.

`ROUTE_SOURCE_MAP["/"]` pointed at an empty slug and never resolved. Fixing it is what makes the sitemap's new existence check safe for the homepage; without it, `/en` and `/de` would have silently dropped out of `sitemap.xml`.

**Each language carries a Prisma migration**, because the display preference is a Postgres enum. `ALTER TYPE ... ADD VALUE` is purely additive and references the new value nowhere, which is what makes it safe inside Prisma's transaction on PostgreSQL 12+. PostgreSQL has no `DROP VALUE`, so retiring a language is a registry edit rather than a down-migration: `resolveUserLocale` resolves any stored value the registry no longer offers back to the default locale.

## Validation

`yarn typecheck`, `yarn lint`, `yarn test` (1,308 passed across 170 files), `yarn build` all green.

**No indexed URL changed.** The whole crawler-visible surface was captured from two locally served instances, one on this branch and one on the exact current `origin/main` commit (`3f093fa`) after rebasing, and diffed: 46 records covering 34 page or redirect responses, `robots.txt`, `sitemap.xml`, `llms.txt`, `llms-full.txt`, and eight locale-negotiation cases, recording HTTP status, `Location`, `Vary`, the `Link` header, canonical, every hreflang, `x-default`, title, description, `<html lang>`, `og:image`, `og:locale`, `robots` meta, JSON-LD `@type` list and the structured-data offer range. The capture was rerun against the final product commit, `ca2d0945`.

After normalising only the two local server ports, `robots.txt` and `llms-full.txt` are **byte-identical**. `sitemap.xml` holds the same 412 URLs, none added, none removed, in the same order.

Every difference falls into seven intended categories, with **zero unexplained lines**:

| # | Change | Count |
| --- | --- | --- |
| 1 | negotiated redirects 308 → 307 | 10 |
| 2 | `Vary: accept-language, cookie` added | 9 |
| 3 | duplicate `Link:` hreflang header removed, font preloads retained | 30 |
| 4 | `x-default` added to every sitemap entry, matching the HTML | 412 |
| 5 | unsupported prefix now returns a real 404 with `noindex` instead of a permanent redirect into `/en/<prefix>/…` | 16 |
| 6 | an application-only locale on a content route redirects to the default locale | 2 |
| 7 | structured-data offer range corrected from 12-59 to 12-69 to match the pricing page | 2 |

Two further differences are the point of the change rather than side effects: an unprefixed application path now negotiates over every routing locale, so `/contact`, `/company/subscription` and `/profile/connected-accounts` resolve to `/it` and `/es` where they previously collapsed to `/en`; and `llms.txt` derives its translated-locale sentence from the registry instead of naming German.

One change worth calling out because it was not in the original list. The homepage `lastmod` moves from the sitemap's own generation timestamp to the real content date, because `ROUTE_SOURCE_MAP["/"]` now resolves a page where it previously resolved nothing. Two entries are affected, `/en` and `/de`. A `lastmod` equal to generation time changes on every fetch and is the value crawlers discount, so this is a correction rather than a regression, but it is a sitemap output change and belongs in this table.

**All three verified against a locally served instance, authenticated as the synthetic seed user:**

- `/{fr,it,es}/auth/signin` return 200 with the matching `<html lang>` and translated copy ("Connectez-vous à votre compte", "Accedi al tuo account", "Inicia sesión en tu cuenta").
- The **authenticated product UI** was driven in a browser in Italian and Spanish: dashboard, contacts and profile settings all render translated, and the sidebar, breadcrumbs, filter chips and table headers carry no English. The only English left on screen is seeded user data (a custom column literally named "Phones", a "Sales Pipeline" select), which is correct in every locale.
- A sweep of 15 authenticated pages per locale, comparing rendered text against the English catalog, reports **zero** English catalog strings for `it`, `es` and `fr`.
- Changing the display language to Italiano in the profile select and saving persists `displayLanguage = it` through the widened enum, so the registry, the server action, Prisma and the database agree end to end.
- `/{fr,it,es}`, `/{fr,it,es}/pricing` and `/es/blog` redirect **307** to their English equivalents rather than rendering an untranslated page; `/{it,es}/dashboard` redirect to the locale's own signin. `/nl/pricing` is a **404** with no `Location`. No response anywhere in the matrix is a 308.
- `sitemap.xml` still has exactly **412** URLs, **0** `fr`/`it`/`es` entries and **0** `hreflang` for any of them. The hreflang set on `/en/pricing` is still `en`, `de`, `x-default`.

**Migration proof, disposable local database, synthetic seed only.** Schema deployed and seeded *without* the new migration, rows snapshotted, migration applied, rows snapshotted again. The enum went from `system,de,en` to `system,de,en,es,fr,it` across the two migrations and every `User` row was byte-identical afterwards, `updatedAt` included, so no row was rewritten. `prisma migrate diff` reports nothing for `Locale`, because the second migration uses `BEFORE`/`AFTER` so the stored order matches `schema.prisma`.

**Translation quality control.** Each bundle was assembled by a generator that refused to write unless the key set matched `en.json` exactly, every ICU argument name matched per key, every rich-text tag matched per key, and no em dash appeared. All five catalogs are complete at 1,527 keys. The values intentionally identical to English (140 fr, 97 it, 88 es) are JS type names and zod validation tokens that appear inside error messages, comparison operator symbols, brand names, and the synthetic seed contacts' proper nouns and example addresses.

`yarn i18n:audit` cross-references every locale against English and gained two improvements from translating at this volume. Its glossary now holds **stems** rather than surface forms, because the same English noun is legitimately a plural or a verb depending on the sentence; matching surface forms produced 32 false positives. And a new check reports namespaces where **two distinct English strings collapse onto one translation**. That found a real defect a text sweep cannot see, because every string it inspects is correctly translated: `between` and `in` both rendered as "tra" in Italian and "entre" in Spanish, making two operators indistinguishable in the same filter dropdown. German and French already distinguished them; Italian and Spanish now read "è uno di" and "es uno de".

New gates: `i18n-parity` now derives its collection list from the filesystem and checks every content locale, bringing **283 previously unguarded files per locale** under guard. Plus `locale-registry`, `locale-consumer-audit`, `fumadocs-i18n`, `localized-assets`, `core/seo/__tests__/sitemap`, and `__tests__/proxy-locales` covering the application-only, content-only, unsupported and default-locale matrix, asserting no response is ever a 308 and that every redirect chain terminates. The parity gate proved itself during this PR: rebasing onto current `main` brought 171 new keys per language and retired 145, and the audit caught every mismatch before publication.

## Impact and rollback

The split is code-only. The three languages add two additive migrations; reverting the code leaves unused enum values, which is harmless and is the documented way to retire a language.

Two changes are visible to search engines and are intended: the sitemap's alternate set now includes `x-default` (matching what the HTML already emitted), and the HTTP `Link` hreflang header is gone in favour of the HTML one. Neither changes a URL, a canonical, or an HTML hreflang set. The three new languages add no public URL at all.

One thing a revert cannot undo: 308s already cached in browsers from before this change.

**Found but out of scope:** `notFound()` called from a page rather than a layout returns **200** with the not-found body, because `app/[locale]/loading.tsx` opens a Suspense boundary that commits the status before the page renders. Every genuinely missing page is a soft 404 today (`/en/blog/no-such-post` returns 200). Pre-existing and unrelated to locales; the application-only-locale case is handled in the proxy so it does not depend on it. Worth its own issue, and relevant to CUS-70.

Refs CUS-49


prisma/migrations/20260803120000_locale_add_fr/migration.sql
```diff
@@ -0,0 +1,11 @@
+-- Adds French as an application display language.
```
benjiwagner: I want you to merge the two migrations and the pull request change set together for simplicity

i18n/README.md
```diff
@@ -0,0 +1,104 @@
+# Locales
```
benjiwagner: Really speaking, we don't like MD readme files because we prefer self-explanatory code instead. This is the same for comments, so therefore please double-check the pull request change set for this.

i18n/locales/en.json
```diff
@@ -1426,7 +1433,7 @@
     "fileTypePdf": "PDF document",
     "fileTypePowerpoint": "PowerPoint presentation",
     "fileTypeWord": "Word document",
-    "groupThread": "Group ({count} members)",
+    "groupThread": "Group ({count, plural, one {# member} other {# members}})",
```
benjiwagner, line 1219: This looks weird. Can you explain by replying here how this will resolve?

benjiwagner: Also, does this look the same across every other locale as well?

ee/subscription/create-checkout-session.interactor.ts
```diff
@@ -46,7 +48,7 @@ export class CreateCheckoutSessionInteractor {
 
     const requestOrigin = (await headers()).get("origin") ?? env.BASE_URL;
     const baseUrl = resolveRequestOrigin(requestOrigin, env.AUTH_ALLOWED_HOSTS, env.BASE_URL);
-    const redirectUrl = `${baseUrl}/company/subscription`;
+    const redirectUrl = `${baseUrl}/${appLocaleOrDefault(await getLocale())}/company/subscription`;
```
benjiwagner: I think we don't need the local prefix for the redirect URL here because it is fine if this is gonna be resolved on runtime. So please deeply check all the other cases across change set and in our code base for simplifications like this, where we hard code a URL endpoint.

prisma/migrations/20260803120000_locale_add_fr/migration.sql
benjiwagner: Merged. There is now one `20260803120000_locale_add_app_languages` adding all three values, and the separate `it`/`es` migration is gone.

Verified both directions rather than assuming, because deleting an applied migration is the risky part:

- **Fresh database** (what `main` will do, since production has applied neither): brought up a throwaway PostgreSQL 16 container and ran `prisma migrate deploy` from empty. Result `{system,de,en,es,fr,it}`, identical to what the two migrations produced, and in the same order as `schema.prisma`.
- **A database that already recorded both** (this branch's preview database, and my local one): `migrate deploy` exits 0 and applies the merged migration as a clean no-op. Prisma does not fail on an applied migration whose directory is gone, and the three `ADD VALUE IF NOT EXISTS` statements find every value already present. Enum unchanged, all `User` rows intact, and `prisma migrate diff` still reports nothing for `Locale`.

The `AFTER` clauses are what keep the stored order equal to the declaration order in `schema.prisma`, so `migrate diff` stays empty rather than reporting a reordering.

i18n/README.md
benjiwagner: Deleted, and I swept the whole change set for comments rather than just this file.

I measured the repository's actual convention instead of guessing at it, and it is not uniform:

| Area | Existing comment lines | My change set | Action |
| --- | --- | --- | --- |
| `scripts/*.ts` | 0 in both existing files | 42 in `audit-i18n.ts`, 11 in `lib/icu.ts` | stripped |
| `tests/conventions/*.test.ts` | 0 across the ones I sampled | 6 | stripped |
| `__tests__/proxy-locales.test.ts` | n/a, new file | 8 | stripped |
| `proxy.ts` | 18 already there | +11 mine | mine stripped, back to 18 |
| `prisma/migrations/*.sql` | 3 to 17 per migration | 12 | trimmed to 4, kept |

Migration comments are the one place I kept anything, because every existing migration carries them and the constraint they record is genuinely invisible in the SQL: `ALTER TYPE ... ADD VALUE` is only safe inside Prisma's transaction while nothing references the new value. If you would rather those go too, say the word and I will strip them and put it in the commit message instead.

Where a comment was carrying real information I moved it into a name rather than deleting the information. In `proxy.ts` the negotiation branch is now `negotiateOverLocalesThatCanServeIt`, which says what the ternary was explaining.

`i18n/README.md` held the two locale-addition procedures. Those are operating knowledge rather than code, so they moved to `engineering/conventions.md` in the knowledge base, which is the repository that owns durable procedure. That is knowledge-base#34, already open. Nothing in the product referenced the README.

`yarn typecheck`, `yarn lint`, `yarn test` (958) and `yarn i18n:audit` all pass after the removal.

i18n/locales/en.json
benjiwagner: That is an ICU plural, and it resolves to a single string chosen by the count. Answering both your questions together since the second is about the other locales.

The message is `"Group ({count, plural, one {# member} other {# members}})"`. `{count, plural, ...}` selects a branch by the plural category CLDR defines for the locale, and `#` prints the count formatted for that locale. The call site is `ee/messaging/thread-display.ts:18`, which passes `{ count: thread.participants.length }` as the fallback name for an unnamed group thread.

Rendered with `intl-messageformat`, which is the same formatter next-intl uses:

| count | en | de | fr | it | es |
| --- | --- | --- | --- | --- | --- |
| 1 | Group (1 member) | Gruppe (1 Mitglied) | Groupe (1 membre) | Gruppo (1 membro) | Grupo (1 miembro) |
| 2 | Group (2 members) | Gruppe (2 Mitglieder) | Groupe (2 membres) | Gruppo (2 membri) | Grupo (2 miembros) |

**Yes, all five locales use the same construction**, and the shape is enforced rather than hoped for: `i18n-parity.test.ts` fails if the ICU argument names differ per key, and I separately compiled all 385 interpolated messages against their own locale so a malformed plural cannot reach runtime.

Note French renders `0` as "Groupe (0 membre)", singular. That is correct French, not a bug: French puts zero in the `one` category while English and German put it in `other`. Writing this by hand instead of as a plural is exactly how that gets wrong.

Before this change the string was `"Group ({count} members)"`, so a two-person chat rendered "Group (1 members)". Two sibling strings had the same defect in a different disguise, `"{count} attachment(s):"` and `"Sent {count} invitation(s)."`, and both are now plurals too. All 8 plural keys in the catalog use a plural in all 5 locales, with no locale opting out.

ee/subscription/create-checkout-session.interactor.ts
benjiwagner: You were right, and it turned out to be more than a redundant prefix: my own proxy was what made it necessary.

**The bug.** `proxy.ts` negotiated *every* unprefixed path over the content locales alone. So `/company/subscription` could only ever resolve to English or German, no matter what the reader's `NEXT_LOCALE` cookie or `Accept-Language` said. Prefixing eight call sites was me working around that instead of fixing it.

Negotiation now picks the locale set that can actually serve the path. Marketing, blog and docs still negotiate over the content locales, because sending a visitor to a page nobody wrote helps no one. Everything else negotiates over all routing locales, because it renders from the message catalog and exists in all of them.

Measured on a running instance with `Accept-Language: it-IT`:

| URL | Before | After |
| --- | --- | --- |
| `/contact` | `/en/contact` | `/it/contact` |
| `/profile/connected-accounts` | `/en/...` | `/it/...` |
| `/company/subscription` | `/en/...` | `/it/...` |
| `/pricing`, `/blog`, `/` | `/en/...` | `/en/...` unchanged, correctly |

`NEXT_LOCALE=it` beats `Accept-Language: de`, which is the in-session case this comment is about. All eight prefixes and the `getLocale` plumbing are reverted. The guarantee moved from a checkout-specific test into the proxy suite; reverting the negotiation split fails it.

**The sweep you asked for.** I enumerated every absolute URL built from `BASE_URL` or `baseUrl` across `app/`, `core/`, `features/` and `ee/`, and classified each:

- **Simplified (locale dropped):** the eight above, plus two I had missed. `features/mcp-tools/deep-research.mcp-tools.ts:87,130` pinned record links to `DEFAULT_LOCALE`, so an Italian reader clicking a citation from an AI client landed on an English page. Contact, organization, deal, service and task routes are application routes, so those now resolve per reader too, verified live for it, de and es.
- **Kept explicit, deliberately:** `core/seo/schemas.ts` (JSON-LD), `core/seo/sitemap.ts` (sitemap entries), and the docs branch of the same MCP file. A canonical, an hreflang alternate and a sitemap `<loc>` are identity statements to a crawler and must not depend on request-time negotiation; documentation exists only in the content locales, so `DEFAULT_LOCALE` is correct there rather than lazy.
- **Already correct:** `features/docs/mcp-install-snippet.ts` points at `/api/v1/mcp`, an API route with no locale at all.

A second, independent agent sweep over the same question is still running; if it surfaces anything I missed I will add it here rather than let it drop.

`yarn typecheck`, `yarn lint`, `yarn test` (958) and `yarn i18n:audit` pass.

benjiwagner: Follow-up, as promised: the independent sweep finished. It **confirmed the two MCP sites I had already fixed**, found nothing else that should drop its locale, and turned up two things worth reporting honestly.

**One real defect, and it is not mine to fix here.** `/docs` and `/docs/:slug` both resolve to `content/docs/{locale}/intro-page.mdx`, so the sitemap publishes the documentation landing page twice per content locale, each self-canonicalising:

| URL | Canonical | Title |
| --- | --- | --- |
| `/en/docs` | `/en/docs` | Customermates documentation |
| `/en/docs/intro-page` | `/en/docs/intro-page` | Customermates documentation |

Nothing tells a crawler which is authoritative. Key-count checks could never catch it, because the sitemap has no exact duplicates: 408 entries, 408 unique. Only the content behind two of them is identical.

It is **pre-existing** and unchanged by this branch, verified by diffing `route-source-map.ts` against `origin/main`, and `/docs` is the only route with this shape. I deliberately have not fixed it here: the fix removes a live indexed URL, and the safety argument of this pull request is that no indexed URL changed. Doing both in one change would make that claim untrue and unreviewable. Filed as CUS-155 with the redirect-or-canonical requirement, since dropping it from the sitemap alone would orphan the URL rather than consolidate it.

**One claim I checked and rejected.** The sweep also argued that `profile-settings-form.tsx` should switch from `window.location.href` to next-intl's `router.replace`, partly because the raw assignment does not sync the `NEXT_LOCALE` cookie. That last part is wrong, and it matters now that negotiation reads that cookie, so I tested it rather than take it:

```
GET /it/contact  ->  set-cookie: NEXT_LOCALE=it; Path=/; SameSite=lax
GET /de/contact  ->  set-cookie: NEXT_LOCALE=de; Path=/; SameSite=lax
```

The assignment triggers a full document load, the proxy runs, and the middleware sets the cookie. The switch is therefore a style preference rather than a fix, and I have left it alone.

The remaining suggestions were to replace hand-built `/{locale}/...` hrefs with `IntlLink` in three components. All correct in principle and all in files this change set does not otherwise touch, so I have left them out rather than widen the diff.

app/[locale]/(static)/page.tsx
```diff
@@ -15,15 +13,9 @@ import { CTASection } from "@/components/marketing/cta-section";
 import { FAQSection } from "@/components/marketing/faq-section";
 import { FeatureSection } from "@/components/marketing/feature-section";
 import { JsonLd } from "@/components/seo/json-ld";
-import { generateMetadataFromMeta } from "@/core/fumadocs/metadata";
-import { homepageSource } from "@/core/fumadocs/source";
+import { homepageSource, pricingSource } from "@/core/fumadocs/source";
 import { organizationSchema, softwareApplicationSchema } from "@/core/seo/schemas";
 
-export async function generateMetadata({ params }: { params: Promise<{ locale: string }> }): Promise<Metadata> {
```
Comment (benjiwagner) [inline app/[locale]/(static)/page.tsx:22]: Can you explain to me why we moved to generate metadata over here?

core/fumadocs/source.ts
```diff
@@ -25,11 +25,7 @@ import {
   pricing,
 } from "@/.source/server";
 
-export const i18n = defineI18n({
-  defaultLanguage: ROUTING_DEFAULT_LOCALE,
-  languages: ROUTING_LOCALES as unknown as string[],
-  parser: "dir",
-});
+export { i18n };
```
benjiwagner: It looks a bit weird that we import and export i18n. Why don't we simply update the usages of i18n and use the proper import directly? Also go through the code again and look for other code occurrences where we did something not that elegant just for backwards compatibility, which should instead be reworked.

scripts/lib/icu.ts
```diff
@@ -0,0 +1,58 @@
+export function icuArgumentNames(message: string): string {
```
benjiwagner, line 1: For what exactly do we need the ICU stuff?

benjiwagner:
When I am logged in and manually remove the local from the URL, the one that is used is not the selected local in the settings. So, for example, if I manually select Italian as a language and then I remove it from the URL, it is then gonna be resolved to English, I think because my browser settings are on English by default.

app/[locale]/(static)/page.tsx
Comment (benjiwagner) [reply app/[locale]/(static)/page.tsx:22]: It went the other way: `generateMetadata` was **removed** from the homepage, and it had to be, because a fix elsewhere in this change set would otherwise have silently rewritten the homepage title.

On `main` the function existed but did nothing. `ROUTE_SOURCE_MAP["/"]` mapped the homepage to `path: []`, while the file is `content/homepage/{locale}/homepage.mdx`, so its slug is `["homepage"]`. `generateMetadataFromMeta` resolves the page and bails on line 25 with `if (!page) return {};`. Every homepage request therefore got an empty object, and the real title, description, `metadataBase` and favicon all came from the root layout.

That empty path is also a landmine for this change set. Once the sitemap only emits pages that actually exist, `getPage([], locale)` returning `undefined` would have dropped `/en` and `/de` out of `sitemap.xml` entirely. So `path` had to become `["homepage"]`. The moment it did, the previously inert `generateMetadata` started returning real metadata from the MDX frontmatter and overriding the layout.

I measured it rather than reasoning about it. Restoring `generateMetadata` on top of the corrected path, against a running instance:

| | `/en` `<title>` | `/de` `<title>` |
| --- | --- | --- |
| with `generateMetadata` | The Agentic Open-Source CRM | Das Agentic Open-Source CRM |
| without it (this branch, and `main`) | Customermates: The Agentic Open-Source CRM \| AI Agents, Unified Inbox, Pipedrive Alternative | Customermates: Das Agentic Open-Source CRM \| KI-Agenten, zentraler Posteingang, Pipedrive-Alternative |

The frontmatter title is the short in-page heading. The layout title is the keyword-targeted one the SEO loop owns, and it is what is indexed today. Keeping the function would have truncated the homepage title on the single most valuable page on the site, as an invisible side effect of a sitemap fix.

Removing it makes the homepage the one route with no `generateMetadata`, which is a fair thing to dislike. The tidier end state is to move `rootMetadata` out of `homepage.mdx` into a `RootMetadata` JSON namespace so the layout and the page stop competing for the same title. That is recorded as a follow-up rather than done here, because `defaultTitle` is SEO copy and moving it belongs in a change that can be measured on its own.

Canonical and hreflang are unaffected: `/en` still self-canonicalises and still emits `en`, `de`, `x-default`, verified on the same run.

core/fumadocs/source.ts
benjiwagner: Agreed, and removed. It was a compatibility shim so no import path had to move while the registry landed, and it turned out nothing was using it.

`i18n` used to be **defined** in `source.ts`. I extracted it to `core/fumadocs/i18n.ts` so `fumadocs-i18n.test.ts` could assert `fallbackLanguage === null` without importing `.source`, and left `export { i18n };` behind so the old path still worked. Checking every importer: the only consumer outside the module already imports `@/core/fumadocs/i18n` directly. The re-export had no users at all.

**The sweep you asked for found two more of exactly the same thing, both mine, both dead:**

| Shim | Why it existed | Actual usage |
| --- | --- | --- |
| `core/fumadocs/source.ts` `export { i18n }` | keep the old import path after extracting the module | none |
| `i18n/routing.ts` `export { ROUTING_LOCALES }` | keep `@/i18n/routing` working as a locale source | none; every consumer imports the registry |
| `i18n/routing.ts` `ROUTING_DEFAULT_LOCALE = DEFAULT_LOCALE` | same, for the default | only inside `routing.ts`, plus a stale test mock |

All three are gone. `routing.ts` now uses `DEFAULT_LOCALE` directly and re-exports nothing, so `i18n/locale-registry.ts` is the only module that names a locale set, which is what `locale-consumer-audit.test.ts` was written to enforce anyway.

Removing them exposed a real gap in `__tests__/proxy.test.ts`. Its `@/i18n/routing` mock still listed the two dead exports and was missing `isContentPage`, which `proxy.ts` has called since negotiation was split by route kind. That file never drives an unprefixed path, so it was latent rather than failing, but a mock that does not match the module it stands in for is a trap for whoever adds the next test. Fixed in the same commit.

I looked for other backwards-compatibility scaffolding in the change set and did not find more: no deprecated markers, no aliases, no parallel old and new code paths. The one thing that reads like a shim and is not is `stripLocalePrefix`, which normalises a Fumadocs `page.url` into a locale-free route key that `buildLocalePath` then re-prefixes. That is a real transformation between two representations, not a compatibility layer.

`yarn typecheck`, `yarn lint`, `yarn test` (958) and `yarn i18n:audit` pass, and I re-checked the public surface on a running instance after the removal: `/en` and `/de` unchanged, canonical and `en`/`de`/`x-default` hreflang intact.

scripts/lib/icu.ts
benjiwagner: It answers one question: **which interpolations does this message actually take?** Two gates need that answer and must agree, so the parser lives in one place instead of two.

`i18n-parity.test.ts` uses it to fail a translation that renames or drops a placeholder, and `yarn i18n:audit` uses it for the same check plus rich-text tags. Both import from `scripts/lib/icu.ts`; nothing else does.

A regex cannot answer it. ICU nests message bodies inside plural and select branches, and those braces do not contain arguments:

```
"{count} filter{count, plural, =1 {} other {s}}"
```

A naive `/\{(\w+)\}/g` reports `["count", "s"]`. The `{s}` is the letter s, the plural suffix, not an argument. So a correct German translation with an empty plural branch gets flagged as breaking interpolation, and the gate becomes noise people learn to ignore. The parser walks the message properly and returns `{count}`, which is right:

| message | arguments | tags |
| --- | --- | --- |
| `{count} filter{count, plural, =1 {} other {s}}` | `count` | |
| `{count, plural, one {# elemento selezionato} other {# elementi selezionati}}` | `count` | |
| `Read our <link>Privacy Policy</link>` | | `link` |

This is not hypothetical. I hit it while translating: my first pass used a regex, it read plural sub-messages as arguments, and it reported the German catalog as broken when it was correct. Extracting the real parser is what stopped that.

`richTextTagNames` lives beside it for the same reason: `next-intl` requires the call site to supply every tag as a render function, so a translation that invents `<enlace>` where the source says `<link>` throws at runtime rather than degrading. The parity gate now fails that at build time, and I mutation-tested it by renaming a tag in `es.json`.

If you would rather not have a `scripts/lib/` at all, the natural home is `core/i18n/`, since `tests/` and `scripts/` both consume it and neither owns it. Happy to move it if you prefer that shape.

app/[locale]/(static)/page.tsx
Comment (benjiwagner) [reply app/[locale]/(static)/page.tsx:22]: Pls doublecheck this. Why shouldnt we have metadata here.

app/[locale]/(protected)/profile/components/profile-settings-form.tsx
```diff
@@ -49,7 +50,7 @@ export const ProfileSettingsForm = observer(({ userDetails, emailVerified }: Pro
   const pathname = usePathname();
   const currentLocale = useLocale();
   const { setTheme, systemTheme } = useTheme();
-  const formId = useId();
+  const formId = "profile-settings-form";
```
Comment (benjiwagner) [inline app/[locale]/(protected)/profile/components/profile-settings-form.tsx]: Why are we hardcoing this here now?

app/[locale]/(static)/page.tsx
Comment (benjiwagner) [reply app/[locale]/(static)/page.tsx:22]: Double-checked and fixed in 0cc43cd5. Homepage metadata now lives on the homepage again, but it is built from the long rootMetadata SEO copy rather than the short page heading. The root layout now contains only route-neutral defaults. This preserves the localized long titles, descriptions, canonicals, Open Graph data, and en/de/x-default alternates on /en and /de, while preventing app-only auth routes such as /fr/auth/signin from inheriting English homepage metadata. I added focused metadata regression tests and verified the final production build and rendered head output locally.

app/[locale]/(protected)/profile/components/profile-settings-form.tsx
Comment (benjiwagner) [reply app/[locale]/(protected)/profile/components/profile-settings-form.tsx]: Removed in 0cc43cd5. The hardcoded profile form ID was unnecessary because FormActions is rendered inside AppForm, so the submit button is already associated with its ancestor form. I removed the form ID, the formId prop, and the brittle source-text convention assertion. anchorScope remains in place, so the stable profile-settings-save and profile-settings-reset automation anchors are unchanged.

benjiwagner:
Pushed the final locale-routing hardening in `4a8a5692`.

The review feedback is resolved as follows:

- **Homepage metadata:** metadata remains page-owned, but the homepage reads its long SEO-specific `rootMetadata` instead of the shorter H1 title. The root layout is now route-neutral, so app-only auth pages no longer inherit English homepage metadata. A full comparison of all 412 sitemap routes against the exact merge base found identical title, description, canonical, Open Graph, HTML language, and hreflang output; the only content difference is the separately approved pricing AggregateOffer correction.
- **Profile form ID:** removed the hardcoded form ID and matching `formId` prop. The actions are descendants of the form, so native form association is sufficient; stable automation anchors remain.
- **Manual locale preference:** application and public-content cookies are now separate. Explicit display-language choices survive locale-less app URLs, System follows browser negotiation, public-content visits cannot overwrite the app preference, and app-only → content links force a full navigation so the locale provider and `<html lang>` remount correctly.
- **Retired/stale preferences:** stored values outside the active app/formatting sets normalize safely to System instead of making profile saves fail.
- **Localized legal mail:** display language, formatting language, `<html lang>`, and locale-less legal links now use their proper domains.

Local verification on the pushed SHA: 1,361 tests, lint, typecheck, production build, i18n audit, disposable PostgreSQL migration/seed flow, authenticated preference changes across all five app locales, content/app cookie isolation, unsupported-locale 404s, and the 412-route SEO crawl all passed.

benjiwagner:
Added the complete remaining localization and locale-transition follow-up in `96b387f0`. This commit is now directly on PR #36; the temporary draft PR #47 was closed without discarding any work.

What this adds:

- Routes the remaining confirmed raw UI and accessibility copy through the five application catalogs.
- Localizes countries, shared form controls, data views, editor/inbox/API-key controls, recipient email layout copy, and localized JSON-LD breadcrumbs.
- Makes number/date/calendar behavior follow the formatting locale, including arbitrary System browser locales and regional week rules.
- Replaces backend English display prose with semantic values translated at presentation boundaries.
- Adds a recurrence guard for visible JSX/config copy and strengthens locale/translation contract tests.
- Fixes the reported navigation inconsistencies: authenticated bare `/` honors the stored application preference, and public-content navigation into protected routes resolves the profile preference instead of retaining the public locale.

Fresh verification for the new head:

- `yarn test`: 196 files / 1,486 tests
- `yarn lint`: zero warnings, including an independent commit-hook rerun
- `yarn typecheck`
- `yarn conventions:check`: 26 files / 200 tests
- `yarn i18n:audit`: 1,313 keys in each of en/de/fr/it/es, no blocking findings
- `yarn build`
- Production-mode local E2E: Italian → German → Italian; bare `/` → Italian dashboard; German public pricing → Italian protected dashboard; System → browser English; Italian calendar localization; no hydration or console errors.

The disposable local PostgreSQL container used only synthetic seed data and was removed after verification. GitHub checks are rerunning on PR #36's updated head.

Refs CUS-49, CUS-152.

components/forms/form-number-input.tsx
```diff
@@ -2,7 +2,7 @@
```
benjiwagner, line 2: Did you test at least one occurrence of a form number input fully end-to-end in the UI on localhost?

components/ui/calendar.tsx
```diff
@@ -3,23 +3,33 @@
 import * as React from "react";
```
benjiwagner, line 3: So did you test the calendar in the UI fully end-to-end locally?

core/stores/intl-number.ts
```diff
@@ -0,0 +1,79 @@
+function numberSymbols(locale: string | undefined) {
```
benjiwagner: This complete file looks extremely complicated, to be honest. Is there not an easier way to do this? It's hard to read and hard to really understand what this is for.

core/utils/toast-zod-error-tree.ts
```diff
@@ -3,55 +3,27 @@ import { toast } from "sonner";
```
benjiwagner, line 3: Did you create different expected Zod errors in the user interface locally end to end to make sure that the error mapping together with Zod is still working as expected?

features/auth/auth.service.ts
```diff
@@ -139,11 +141,15 @@ export class AuthService {
 
   async sendVerificationEmail(args: { to: string; url: string }): Promise<void> {
     const t = await getTranslations();
+    const locale = appLocaleOrDefault(await getLocale());
```
benjiwagner: We are having this a couple of times where we get the local and then pass this again, for example, the app local or default. So we get the local and curve in the backend, for example, nested with a different function or user interface as well. Isn't it a bit simpler instead to create these helper functions directly or hooks in the front end so we don't have those duplicates?

benjiwagner: Also, I'm not quite sure whether we actually need this, because the get local shutout the actual local user interface is currently having the session. Although I'm not quite sure, be critical.

features/auth/auth.service.ts
```diff
@@ -156,11 +162,15 @@ export class AuthService {
 
   async sendResetPasswordEmail(args: { to: string; url: string }): Promise<void> {
     const t = await getTranslations();
+    const locale = appLocaleOrDefault(await getLocale());
+    const layoutCopy = await getEmailLayoutCopy(locale);
```
benjiwagner, line 167: What is the email layout copy for?

features/auth/continue-with-socials.interactor.ts
```diff
@@ -26,8 +26,8 @@ export class ContinueWithSocialsInteractor {
     private readonly findUserRepo: FindUserRepo,
   ) {}
 
-  @Enforce(Schema)
-  async invoke(data: ContinueWithSocialsData): Promise<{ ok: true; data: null } | Redirect> {
+  @Validate(Schema)
+  async invoke(data: ContinueWithSocialsData): Promise<Awaited<Validated<null>> | Redirect> {
```
benjiwagner, line 30: Typing wise, this looks a bit weird. Are you sure this is the correct way to do it? Compare this with the other usages of the validate function.

benjiwagner:
## Final hosted happy-path verification

Verified final deployed head `d8787e6990bac8d1456d604c6d4e3c191d61a1ed` with the synthetic preview account.

### Hosted UI coverage

- Public homepage → Features → Docs → Pricing.
- Signed-in dashboard, Contacts list/detail, Inbox list/thread/composer, Profile settings, API & Connectors, connected channels, company Members/user modal, Add Contact, Feedback, and Audit logs.
- Opened the translated API-key/calendar and contact forms; canceled without creating or sending data.
- Changed display language Italian → German → Italian.
- Final preference matrix: Italian → System produced `/en/profile/settings` with `lang=en`; System → Italian restored `/it/profile/settings` with `lang=it`.
- Signed-in bare `/` resolved to `/it/dashboard`.
- Italian protected UI → English docs → “Go back” returned to exactly `/it/dashboard`.

### Blocker found and fixed

The old deployment produced `/it/it/dashboard` on the docs-to-app return path.

- `ea03cfdd` normalizes/replaces an existing locale prefix before reconciliation and adds the exact regression cases with query/hash preservation.
- `d8787e69` constrains the client navigation to the current origin, satisfying CodeQL without weakening the behavior.

The final deployed reproduction passes: `/en/docs` → “Go back” → `/it/dashboard`, Italian shell, no 404, no double prefix.

### Final gates

- Focused locale/proxy/AppLink suites: 33/33 passed after each fix.
- Typecheck and targeted/full lint passed.
- GitHub CI passed (4m45s), CodeQL passed, repository policy passed, Vercel READY.
- Final browser log window: 0 errors and 0 hydration failures. The only console noise was the existing Recharts initial `width(-1)/height(-1)` warning on dashboard mount.

Worktree is clean and matches the pushed branch.

components/forms/form-number-input.tsx
benjiwagner: Yes. I tested two real `FormNumberInput` consumers on localhost. In Italian service settings, I changed Data Migration from `25.000` to `1.234,5`, blurred, saved, reloaded, and confirmed the same stored value; malformed `1,2,3` was rejected without truncating or persisting, then I restored `25.000`. In a real deal, I changed a service quantity from `1` to `1,5`, verified the line and aggregate totals changed, saved/reloaded, and restored the original quantity. The focused parser/component suites and the full 1,513-test run are green.

components/ui/calendar.tsx
benjiwagner: Yes. I exercised the calendar through real localhost consumers: API-key expiry (single-date plus disabled-date limits) and a deal custom date range. I navigated months, verified past/today and the 364-day boundary, selected/reopened/cancelled the API-key date, and changed a cross-month range, saved, and reloaded it. I also switched the actual formatting preference through English, German, French, Italian, Spanish, and System and checked captions, weekday order, and accessible day labels. The final production build and full suite are green, and the test data was restored afterward.

core/stores/intl-number.ts
benjiwagner: Good pushback. There is no inverse parser in `Intl.NumberFormat`, and the strict logic is needed to avoid silently changing values for locale digits, bidi signs, whitespace groups, and 3/2 grouping such as `en-IN`. I kept the behavior but made it easier to follow by extracting the locale-syntax and grouping helpers and caching syntax per exact locale. Tests cover all five preferences, System isolation, `en-IN`, `ar-EG`, valid signs/decimals, malformed grouping, and suffix rejection; the real Italian service/deal save-and-reload paths also passed.

core/utils/toast-zod-error-tree.ts
benjiwagner: Yes, and this UI pass exposed a real gap that is fixed in `c5acfb81`. On localhost I exercised localized signup validation, a sign-in root error, duplicate invalid invite recipients, invalid webhook URL plus missing events, and a blank contact field. The toasts contained localized messages only—no `firstName`, `emails[0]`, or synthesized English labels. I added tests using real `z.treeifyError`, localized email/URL/event error codes, deduplication, nested arrays, and root errors. The empty webhook form now emits exactly one URL issue and one events issue.

features/auth/auth.service.ts
benjiwagner: You were right that the boundary was duplicated. I added one server helper, `getRequestAppLocale()`, and replaced all four direct `getLocale()` plus `appLocaleOrDefault()` occurrences. Normalization is still required: `getLocale()` represents the request/routing locale, whose domain can include content-only locales, while these auth emails and validation catalogs require `AppLocale`. A React hook cannot be used in Better Auth callbacks or server interactors. Tests cover a valid app locale and a content-only locale falling back safely.

features/auth/auth.service.ts
benjiwagner: `EmailLayoutCopy` carries the two shared localized footer values—the product tagline and address country—into the pure React email layout. Email templates, background jobs, and previews do not all have a next-intl request context, so resolving copy at the composition boundary keeps the layout synchronous and explicit. Recipient emails now get their application locale; operator notifications intentionally receive the explicit English default. Tests verify Italian/French recipient layouts and the operator default.

features/auth/continue-with-socials.interactor.ts
benjiwagner: I compared this with the other `@Validate` interactors and kept the type because it is the established shape. `Validated<T>` is itself a Promise, while this method may return either the awaited validation result or a redirect, so `Promise<Awaited<Validated<null>> | Redirect>` avoids a nested-Promise branch and represents both outcomes. I added coverage for invalid callback/provider input, valid Google and Microsoft paths, provider redirect, new/existing users, and the unverified-user redirect; full typecheck passes.

PR #37, opened 2026-08-03

benjiwagner:
## Summary

Ships the complete customer-facing hosted AI Assistant for CUS-125 as one feature.

- **Chat.** Customermates-branded compact and expanded chat, streaming Markdown, stop, retry, one editable queued follow-up, jump-to-latest, responsive layout, keyboard access and screen-reader semantics.
- **Safe activities.** Expandable semantic activity summaries that never expose UUIDs, raw arguments or results, secrets, hidden reasoning, page-context markup, provider identifiers or costs. UI-command outcomes stay truthful, so a failed navigation or tour is reported as a failure.
- **History.** Multiple conversations with server-side title and sanitized-message search, 25-conversation and 50-message cursors, upward transcript loading, new/switch/archive/undo/restore, sequence-based unread support replies, and permanent deletion from the archive.
- **Page awareness.** Exactly three deterministic suggestions per route from one centralized catalog keyed on route, data state, permissions, locale and custom terminology; guided tours; and the panel opening itself when the current page is genuinely empty, since it already carries that page's suggestions and a composer.
- **Workspace setup.** Conversational discovery, review, apply and cleanup covering custom columns, minimal linked records and dashboard widgets, with provenance so cleanup can be reversed safely.
- **Credits.** Tiered monthly hosted-AI credits with atomic reservation before provider access, boundary-safe settlement, and a full-width monthly meter.

This replaces the previous fixed `$2/week` cap entirely. Conversation history is delivered here rather than deferred; CUS-171 remains a duplicate of CUS-125.

## Surface boundary

The Assistant owns its own surface and nothing else. It adds the launcher and the panel, and decides when to open itself. It does not change how any existing page looks: `data-view-empty.tsx` and `data-view-empty-state.tsx` are **byte-identical to `main`**.

The one remaining difference under `components/data-view/` is `data-view-container.tsx`, which carries the shared back-to-top control from CUS-180. That arrived here because PR #46 was stacked on this branch and merged into it, so it rides along rather than being part of the Assistant.

An earlier revision rendered an inline block on empty pages. That duplicated what the panel already offers and made a shared component import from a feature, so it was removed. Empty-state and per-page loading presentation are deliberately a separate effort.

One change was reverted along with it and is worth picking up there: `data-view-empty.tsx` had derived `canCreate` from `userStore.can(resource, Action.create)` rather than `store.canManage`. That is a real permission refinement, but it changes existing pages. Reverting restores `main`'s behaviour rather than introducing a regression.

## Panel design and the approval model

The launcher carries a chat icon rather than the product mark, so it reads as a conversation rather than as branding. The panel itself is a card surface with a rounder radius, roomier spacing and one shared card treatment across the support, error, approval and workspace-setup blocks. Assistant replies are plain text with no avatar and a copy action that appears on hover once the reply is finished; only the user's own turns keep a pill. A running turn shows three animated dots and a live step list, and settles into "Thought for Ns". The send control becomes a stop control while a turn runs rather than sitting beside one. The credit meter is a quiet line of reset date and percentage, and only draws its bar once usage passes three quarters. Every animation is behind `motion-reduce`.

Approvals now match the risk. Read-only tools are no longer gated at all: gating keys off the `readOnlyHint` annotation, so the nineteen read tools run without interrupting the user. Ordinary creates and updates offer **Approve once**, **Approve always** and **Cancel**. Everything else, including deletes, outbound messages, workspace setup and cleanup, and support escalation, offers only **Approve once** and **Cancel** and says plainly that it will ask every time. "Approve always" is therefore only reachable for the ten create and update tools across the five entity types.

## Context

**Commercial contract.** Starter 200, Pro 500, Business 1,200 credits per active user per month. Enterprise uses a finite internally configured allowance and fails closed when it is missing. Trials get 500 credits per active trial user with no workspace cap. Credits reset on the billing-anniversary day including annual plans, trial-to-paid starts a fresh paid allowance, new paid seats get whole-credit proration, upgrades raise the ceiling immediately and downgrades clamp it immediately. There is no rollover, top-up, redistribution or automatic overage in v1. One credit is one started cent of internally metered processing: a turn is charged `max(1, ceil(total provider cost / $0.01))`. Pre-provider failures cost zero; provider-started failures settle known usage or retain the bounded reservation. Dollars, tokens, model identifiers and provider pricing stay server-only. Only Customermates-hosted processing is metered, customer-funded external MCP and API-key clients are not, and the hosted Assistant is cloud-only while self-hosted keeps MCP access.

**Persistence.** One additive forward migration, `20260812090000_agent_assistant_turn_ledger`, adds ordered `AgentMessage.sequence` and sanitized `searchText`, conversation archive/selection/read-sequence state, `AgentTurnRequest` for replay and provider-start durability, the explicit reserved/settled/retained/released credit ledger with plan and allowance snapshots, `AgentWorkspaceSetup` and its resources, the subscription credit anchor and Enterprise override, the user seat-activation timestamp, and a nullable `SupportTicket.agentConversationId` with `onDelete: SetNull`. It drops the legacy `@@unique([companyId, userId])` that allowed only one conversation per user. Database CHECK constraints enforce the money invariants on the deploy path; note that `prisma db push --force-reset` on the local reset path drops them, so the application enforces the same rules independently.

**Legacy backfill.** Existing rows are migrated deterministically: message sequence is renumbered by `(createdAt, id)` rather than heap order, `selectedAt` is backfilled so a never-selected conversation cannot outrank a recent one under `ORDER BY ... DESC`, and unread state is derived from the legacy `userLastReadAt`. `searchText` is derived from visible text and fails closed: any legacy row still carrying a private marker is blanked rather than made searchable. Legacy usage rows become settled ledger records on an epoch sentinel period, so the billing history is retained but can never be charged against a live allowance, and a zero-cost legacy row stays at zero instead of inventing a charge.

**Relationship to CUS-68 / PR #42.** This PR is deliberately decoupled from the unmerged activity-timeline widget work. Every `Widget` column reference the Assistant needs is isolated in `features/agent-chat/agent-widget-dependencies.ts`, so re-coupling to the `WidgetKind` union after PR #42 merges is a small change confined to that one file. The migration is dated to sort after PR #42's latest in either merge order, and the two migration sets are column-disjoint. `tests/helpers/database-test.ts` is copied byte-identically from PR #42 so the database-test opt-in merges cleanly whichever lands first.

**Overlay contract.** The branch adopts the CUS-60 contract that arrived on main: the tour tooltip moves from `AppModal` to the shared collision-aware `Popover`, safe-area and viewport sizing move to the shared tokens, and `components/modal/app-modal.tsx` and `components/ui/dialog.tsx` return to main's versions with zero drift.

## Validation

Local gates at `e7d91936`:

- `yarn typecheck`: passed. This removes two `as unknown as` Prisma shims that were previously hiding 90 type errors across 74 call sites, so those call sites are now genuinely checked.
- `yarn lint`: passed with zero warnings.
- `yarn test` with no database and no `RUN_DATABASE_TESTS`: 192 files passed, 3 skipped; 1,694 tests passed, 15 skipped. This proves the suite stays green on a machine without PostgreSQL.
- `yarn test` with `RUN_DATABASE_TESTS=true` against local PostgreSQL 16: **195 files and 1,709 tests passed**.
- `yarn build`: passed; `/api/agent/messages` is present in the production route manifest.
- Generated OpenAPI and raw docs: regenerated with no tracked diff.
- `prisma migrate deploy` on a fresh database followed by `prisma db seed` twice: passed, matching the CI idempotency check.
- `prisma migrate diff --from-config-datasource --to-schema`: no drift between the applied migrations and `schema.prisma`.

Migration evidence, in `prisma/__tests__/agent-assistant-migration.database.test.ts` against disposable databases:

- Legacy conversation, message, usage, support-ticket and widget fixtures upgrade cleanly.
- Sequence backfill is **deterministic**: the same fixture built twice with tied `createdAt` rows inserted in opposite physical order produces identical sequences.
- `searchText` strips a UUID and an `sk-` shaped token while keeping surrounding words searchable, and blanks a message carrying page-context markup.
- Unread support state derives correctly from the legacy read timestamp.
- The legacy unique index is gone and a second conversation for the same user inserts.
- Deleting a conversation nulls `SupportTicket.agentConversationId` and retains both the ticket and the usage rows.
- A legacy-upgraded database converges on exactly the same columns, indexes and check constraints as one built from scratch.
- The ledger check constraints reject a charge above its reservation, a reserved row that is already settled, a released row carrying a charge, and a negative amount.

Credit evidence:

- `features/agent-chat/__tests__/agent-credit-ledger.database.test.ts` proves atomic reservation against a real database: with three credits remaining, five concurrent reservations admit exactly three and reject two, and the ledger totals exactly the allowance. Mocked repositories cannot express this.
- Pure-function coverage for rounding at every boundary (0, 1, 999,999, 1,000,000, 1,000,001, …), month-end anchors that clamp into February and are then restored in March, leap-year and leap-day anchors, the exact anniversary instant, and monthly reset under annual billing.

Gates re-run at `ced627e3`: `yarn typecheck` passed, `yarn lint` passed with zero warnings, `RUN_DATABASE_TESTS=true` ran **196 files and 1,716 tests, all passing**, and `yarn build` passed.

Browser evidence, driven against a local `APP_MODE=cloud` server with the real provider:

- English journey: streaming turn, read tools running ungated, a `create_contacts` approval offering Approve once / Approve always / Cancel, a `delete_records` approval offering only Approve once / Cancel with its reason line, approve-once creating the record while `preAuthorizedAgentTools` stayed empty, and the credit meter staying quiet at low usage.
- German journey: create and delete end to end in one conversation, with "1s nachgedacht", the live step rail, "Einmal freigeben / Immer freigeben / Abbrechen" on the create, "Einmal freigeben / Abbrechen" plus "Diese Aktion fragt jedes Mal nach." on the delete, and the record confirmed created and then deleted in the database.
- Light and dark themes: the panel is a card surface, and the user pill stays legible in light mode where `--muted` and `--background` are nearly identical.

Three defects were found by driving the assistant in German and are fixed here:

- The tool selector recognised "erstellen" and "hinzufügen" but not "anlegen", "erfassen", "eintragen" or the separable verbs, so a natural German create request never received `create_contacts` and the turn reported that creation was unavailable. Covered by new tests over five real phrasings plus a negative case.
- Delete copy was always plural, so removing one record read "1 Kontakte dauerhaft löschen". Deletion now shares the counting helper the create and update paths already used.
- The assistant configuration request had no deadline, and because the in-flight promise is shared, one request that never settled could leave the panel unmounted for the rest of the session. It now fails after fifteen seconds and lets the existing backoff retry.

Accessibility pass over the reworked panel, read from the live accessibility tree:

- The panel is `role="dialog"` labelled "Assistant", and is deliberately not `aria-modal`: it is a dockable side panel, so it must not trap focus away from the page behind it.
- Every control carries an accessible name, and tab order runs header controls, transcript, per-message copy actions, composer.
- Turn status is announced through an `aria-live="polite"` region; the step list is `aria-live="off"` so a long turn does not narrate every step.
- One defect found and fixed: opening the panel unmounted the launcher, so focus fell back to `<body>` and the next tab restarted at the top of the page. Opening from the launcher now moves focus into the composer, and the panel that opens itself on an empty page deliberately does not, so arriving on a page never takes the caret away from what the user was doing.

Permission-limited pass, walking Elena Hoffmann (may write contacts, organizations, tasks and inbox messages, but not deals or services) through the Deals page:

- The three offered suggestions are read-only on a page whose records she cannot create; no "add a deal" is offered.
- With her seat not credit-activated, the panel replaces the composer entirely with "Allowance unavailable" and a plans link, so she cannot spend.
- Approving a deal create still wrote nothing: the row count stayed at zero, the step turned red, and the reply said no changes were made. The approval card is a consent gate, not the security boundary, and the boundary held.

## Known gap, deliberately not fixed here

A permission denial currently surfaces as a generic failure: the user is told the CRM operation failed and is invited to retry, rather than that they lack the role. The sanitized tool error surface is right in principle but flattens "not allowed" into "something went wrong". Separating the two is a contained change to the error taxonomy and reaches beyond this rework.

## Localization after #36

PR #36 landed the locale registry and added Spanish, French and Italian while this branch was open. Integrating it was more than a merge: the new locale audit bans the pattern the assistant's copy layer was built on, and rightly so, because that layer would have shown an English assistant inside an otherwise translated interface.

`features/agent-chat/agent-activity.ts` held its own English and German tables plus about ninety `language === "de"` ternaries. All of it now resolves through the message catalogs. `agentActivityCopy` takes a translator instead of a locale string, and the module keeps only the ids it needs to route a key. German verb agreement, previously a ternary on a singular flag, is a real ICU plural, so one record no longer reads "1 Kontakte".

The workspace setup labels moved the same way, through a translator threaded from `TranslationSync` on the client and `createTranslator` on the server. Recovering them surfaced a latent bug: sixteen entries in the old German map had unquoted keys, including the deal stage and its options, so they were never being translated at all.

Spanish, French and Italian are properly translated, not English placeholders, which is the bar the rest of the catalogs already meet at about five percent English. Every translated message was verified against its English source with an ICU parser for identical arguments, identical structure and valid syntax, and the sanctioned review lists were used only where a string is genuinely identical, such as pure interpolation or a product name.

All five catalogs are at key parity, 1,697 keys each.

## Impact and rollback

Additive persistence plus one dropped constraint. No existing customer data is rewritten: the only updates are backfills of newly added columns. Existing widgets, CRM records, support tickets and usage rows are preserved, and the migration was verified against a database that already held data.

Immediate rollback is the existing feature flag: set `AGENT_CHAT_ENABLED=false`. The Assistant is additionally gated behind `CLOUD_HOSTED` at the protected layout, all server actions and the SSE route. Keep the additive persistence in place; do not attempt a destructive schema downgrade. Code rollback is to revert this PR, after which the agent tables remain inert.

Code Owner approval, the merge, the production migration and post-merge production verification remain human-only.


benjiwagner:
Closing this stale PR because #62 explicitly supersedes it after the branch history rewrite and carries the same feature work plus subsequent changes. The active implementation continues in #62.

benjiwagner:
Superseded by #62.

That branch carries this work plus everything since, on a clean history that merges `main` rather than rewriting it, so it needs no force-push. Since this PR was opened the assistant also gained AI-composed guided tours and lost roughly 2,100 lines: "Approve always" and its standing per-user grant (including the Prisma column), the hard-coded tour tables, and the code an audit confirmed nothing reaches.

Closing in favour of #62. `feat/agent-chat` is retained for now and can be deleted once #62 merges.

PR #38, opened 2026-08-03
feat(contacts): open sidebar entity drawers from the left and add channels while creating a contact

benjiwagner:
Linear: [CUS-61](https://linear.app/customermates/issue/CUS-61). Not using a closing keyword: the issue closes only after production deployment is verified.

## Summary

The sidebar Add picker and the shared entity drawer now open from the left edge, matching the sidebar they belong to. One change covers all five entity types because they share a single `EntityDrawer`.

Contact channels can now be added while creating a contact. `contact-detail-view` rendered `ContactChannels` only when `fetchedEntity` existed, and that is `null` during create, so the whole section was structurally unreachable until after the first save. The section now renders in the draft and staged identifiers commit atomically with the contact through the existing create contract. Only the inbox deep-link, which genuinely needs a persisted id, is withheld before the first save.

Four defects found in exactly this path are fixed alongside:

- The new-contact reset omitted `taskIds`, so tasks from a previously viewed contact silently attached to the next contact created.
- Client dedupe keyed on exact `provider + value` while the database unique key is `(companyId, channelClass, value)`, so one address added under two email providers passed the client check and then lost a channel on write.
- `validateIdentifierConflicts` never rejected duplicates within a single create payload, because `selfContactId` is `undefined` there and the guard compared `undefined` to `undefined`.
- `popTop` pushed a history entry instead of replacing, so open, close, Back reopened a closed drawer. Closing a dirty drawer also bypassed the unsaved-changes guard entirely.

## Context

The backend already supported everything needed: `BaseCreateContactSchema.identifiers` accepts the array, `validateIdentifiers` normalizes it, `ContactWritePrecheckInteractor.create` already passes `selfContactId: undefined`, and `PrismaContactRepo.createContactOrThrow` writes contact plus identifiers inside one `@Transaction` with a per-company advisory lock. No data model, schema, or transaction work was needed, and no client-side create-then-link sequence was added.

Design decisions worth a reviewer's attention:

- **Channel permission is now mode-aware.** Draft mode requires `contacts:create`, editing keeps requiring `contacts:update`. `CheckChannelConflictInteractor` was widened from `action: update` to `{permissions: [create, update], condition: "OR"}` using the existing decorator shape. This is necessary because `TenantInteractor` throws `ForbiddenError` rather than returning `ok:false`, so a create-only user previously got an unhandled server-action error. It leaks nothing new: the interactor returns a bare boolean, the owners lookup is `companyId`-filtered, and the same signal is already obtainable by submitting the create.
- **The close guard is scoped to the drawer's own store**, held in local state in `entity-drawer.tsx`, rather than routing `writeStack` through the guarded router. `.eslintrc.json` documents the raw router in `use-entity-drawer-stack.ts` as intentional; routing through the intl router would also guard drawer *opening* (both go through `writeStack`) and would double the locale prefix, since `writeStack` builds its URL from `next/navigation`'s `usePathname()` which already includes `/en`.
- **`popTop` uses `replace`, not `router.back()`.** `back()` is correct for open-then-close, but on a deep link straight to `?open=contact:new` it either does nothing (drawer stuck open) or leaves the app. `replace` is correct in all three traces at the cost of one visually inert Back press.
- **A new `duplicateChannel` error code** rather than reusing `channelAlreadyLinked`, whose message says "already linked to **another contact**" and is actively misleading when the duplicate is in the draft you are looking at.

Deliberately out of scope: the `slide-in-from-*` classes in `sheet.tsx` are dead code (`tailwindcss-animate` is not installed), so nothing animates today and "opens from the left" means resting position plus border edge. Enabling them would switch on animation for every Sheet and Dialog app-wide. `components/ui/sheet.tsx` is untouched, including its `side = "right"` default, which the mobile sidebar and public navbar rely on.

**One existing test was intentionally inverted.** `core/validation/__tests__/validators.test.ts` had `it("passes when one batch row repeats its own identifier")` asserting `addIssue` was never called, pinning the duplicate bug as expected behavior. It now expects `duplicateChannel`. Flagging this explicitly rather than leaving it to be discovered in review.

## Validation

All gates green on this branch:

| Gate | Result |
| --- | --- |
| `yarn typecheck` | pass |
| `yarn lint` | pass, 0 warnings |
| `yarn vitest run tests/conventions` | 14 files, 63 tests |
| `yarn test` | 115 files, **924 tests** |
| `yarn build` | pass |

(The issue's criteria name `yarn lint:check`, which does not exist; `yarn lint` was run.)

Browser acceptance was driven against a local instance in an isolated worktree with a disposable PostgreSQL 16 container, synthetic seed only, `APP_MODE="cloud"`. The change was stashed to `origin/main` mid-session and the same steps re-run, so the baseline is measured rather than assumed:

| Behavior | Baseline | After |
| --- | --- | --- |
| Entity drawer | `left: 640, right: 1280`, slide-from-right | `left: 0, right: 640`, slide-from-left |
| Channels in create drawer | absent | present |
| Dirty drawer close | closes silently | prompts; Discard resets |
| Open, close, Back | reopens | stays closed |

Highlights:

- A mutation observer armed before the picker click recorded **0 frames** containing `slide-in-from-right`, and exactly one sheet panel mounted at any sample, so no double overlay and no cross-screen flash.
- One save persisted the contact plus both identifiers (`channelClass` `email` and `phone`); a full reload showed both. Zero rows existed before the save.
- Forced a mid-transaction failure via a temporary `CHECK` constraint: 0 contacts, 0 identifiers, 0 orphans, and the form stayed recoverable.
- At 375x812 the mobile off-canvas sidebar (itself a left Sheet) hands off to the now-left Add picker with exactly one panel and one overlay at 0/150/350/800ms, no horizontal overflow, and a real tap on Save created a contact.
- Public marketing navbar still opens from the right.

Full evidence, including the residual items below, is attached to CUS-61.

Known, verified as pre-existing on the `origin/main` baseline and not introduced here: a dev-only Radix `useId` hydration mismatch in the sidebar, and focus returning to `document.body` after the drawer closes (the drawer has no `SheetTrigger` for Radix to restore to).

## Impact and rollback

No migration, no schema change, no data repair. The create contract and its transaction boundary are unchanged.

Two behavior changes a reviewer should weigh:

1. Closing a drawer with unsaved edits now prompts. This is the contract every other modal in the app already follows and matters more now that channels can be staged before save, but it is new friction on a path that previously closed silently.
2. Identifiers duplicated within one create or update payload are now rejected instead of silently collapsing. Previously the second `upsert` took the update branch without writing `provider`, so a channel vanished without feedback. API and MCP callers that were unknowingly sending duplicates will now get a validation error.

Rollback: revert this commit. The drawer-side change, the draft-capable channels UI, the guard, and their tests are one coherent unit. If only the create-mode Channels section proves problematic, re-adding the `fetchedEntity &&` guard at `contact-detail-view.tsx` restores the previous behavior while keeping everything else.

### DIFF
--- a/features/contacts/upsert/validate-identifiers.ts
+++ b/features/contacts/upsert/validate-identifiers.ts
@@ -63,8 +63,19 @@ export function validateIdentifierConflicts(
   const claimedBy = new Map<string, string | undefined>(owners);
 
   contacts.forEach(({ selfContactId, identifiers }, index) => {
+    const claimedInPayload = new Set<string>();
+
     (identifiers ?? []).forEach((identifier, i) => {
       const keys = channelStrings(identifier).map((value) => identifierKey(identifier.provider, value));
+
+      if (keys.some((key) => claimedInPayload.has(key))) {
+        ctx.addIssue({
+          code: "custom",
+          params: { error: CustomErrorCode.duplicateChannel },
+          path: [...basePathFor(index), i, "value"],
+        });
+        return;
+      }
 
       if (keys.some((key) => claimedBy.has(key) && claimedBy.get(key) !== selfContactId)) {
         ctx.addIssue({
@@ -75,7 +86,10 @@ export function validateIdentifierConflicts(
         return;
       }
 
-      for (const key of keys) claimedBy.set(key, selfContactId);
+      for (const key of keys) {
+        claimedBy.set(key, selfContactId);
+        claimedInPayload.add(key);
+      }
     });
   });
 }
--- a/app/[locale]/(protected)/contacts/components/__tests__/add-channel.store.test.ts
+++ b/app/[locale]/(protected)/contacts/components/__tests__/add-channel.store.test.ts
@@ -1,0 +1,96 @@
+import type { RootStore } from "@/core/stores/root.store";
+import type { IdentifierInput } from "@/features/contacts/contact.schema";
+
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { MessagingProvider } from "@/generated/prisma";
+
+const contactActions = vi.hoisted(() => ({
+  checkChannelConflictAction: vi.fn(),
+  getContactsAction: vi.fn(),
+  searchChannelCandidatesAction: vi.fn(),
+}));
+const inboxActions = vi.hoisted(() => ({ resolveProviderProfileAction: vi.fn() }));
+
+vi.mock("../../actions", () => contactActions);
+vi.mock("../../../inbox/actions", () => inboxActions);
+vi.mock("sonner", () => ({ toast: { error: vi.fn(), success: vi.fn() } }));
+
+import { AddChannelStore } from "../add-channel.store";
+
+function makeStore(channels: IdentifierInput[]): AddChannelStore {
+  const root = {
+    contactDetailStore: { channels, addChannel: vi.fn() },
+    connectedAccountsStore: { ensureLoaded: vi.fn(), usableSendersFor: () => [] },
+    userStore: { user: null, can: () => true, canManage: () => true, canAccess: () => true },
+  } as unknown as RootStore;
+  return new AddChannelStore(root);
+}
+
+const EMAIL = "shared@example.com";
+
+beforeEach(() => {
+  vi.clearAllMocks();
+});
+
+describe("AddChannelStore.addAsNewOptions", () => {
+  it("offers Mail for an address the contact does not have", () => {
+    const store = makeStore([]);
+    store.query = EMAIL;
+
+    expect(store.addAsNewOptions).toEqual([MessagingProvider.mail]);
+  });
+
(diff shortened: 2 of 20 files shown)


features/contacts/upsert/validate-identifiers.ts
```diff
@@ -63,9 +63,20 @@ export function validateIdentifierConflicts(
   const claimedBy = new Map<string, string | undefined>(owners);
 
   contacts.forEach(({ selfContactId, identifiers }, index) => {
+    const claimedInPayload = new Set<string>();
+
     (identifiers ?? []).forEach((identifier, i) => {
       const keys = channelStrings(identifier).map((value) => identifierKey(identifier.provider, value));
 
+      if (keys.some((key) => claimedInPayload.has(key))) {
```
benjiwagner, line 71: Are we completely certain that the validate identifiers was not handling this already before?

PR #42, opened 2026-08-04
feat(dashboard): add a filterable activity timeline widget

benjiwagner:
## Summary

Completes [CUS-68](https://linear.app/customermates/issue/CUS-68/add-a-filterable-activity-timeline-widget-to-the-dashboard) by adding a configurable activity timeline widget alongside existing dashboard charts.

Dashboard widgets become a typed `chart | activityTimeline` discriminated union, and `timelineFilters` is the sole activity-filtering contract across the editor, preview, runtime, persistence, REST/OpenAPI and MCP. Relationship filters cover Contacts, Organizations, Deals, Services and Tasks with `in`, `notIn`, `hasSome` and `hasNone`; the value-less operators never render a selector. Active fields are ANDed, values inside one `in` are ORed, and unlinked activity fails positive filters while passing negative ones.

**Three changes reach beyond the widget and are called out deliberately:**

- **`hasSome`/`hasNone` are redefined app-wide** from "has some of these values" to "has any related record at all". `core/base/filter-compat.ts` migrates saved filters (`hasSome`+value → `in`, `hasNone`+value → `notIn`); the translation is lossless because both compiled to byte-identical Prisma fragments on `main`, and `filter-syntax.mdx` already documented them as value-less.
- **`get_activities` changes shape** on MCP and `POST /v1/messaging/activities/search`: flat `entityType`/`entityId` becomes a nested `scope` object, and the MCP tool schema is now `.strict()`. The issue permits shape changes without versioning, but old callers will get a hard error rather than a warning.
- **A `segmented` tabs variant is added** to `components/ui/tabs.tsx` for the widget editor's Data/Appearance switch. It is purely additive: `default` and `line` are untouched, and `display-options.tsx` is not part of this diff.

### Minimality ledger

**Reverted** — 41 formatting-only hunks across 13 files, each proven safe by token-normalized equality with the original, so the diff shows only semantic change; the never-referenced `activityScopeEmpty` error code and its translations; a redundant `.min(1)` that rejected an empty `connectedAccountIds` list the repository already treats as absent.

**Reduced** — the editor preview no longer simulates the timeline (it mounts the real per-widget store, deleting 467 production lines and a 325-line test that re-implemented the filter semantics a second time); duplicated display-option fallbacks collapsed into typed defaults plus one merge helper; the viewer permission derivation that two repositories restated verbatim is now shared; the single-token activity summary builder is inlined into its only consumer.

**Kept with rationale** — the `Promise<boolean>` delete-confirmation contract (cannot be half-migrated across its call sites); accessibility additions; the illustrated chart-type picker; docs table realignment; the generation counters in the widget editor store.

**Closed as pattern-consistent** — `core/base` importing `features/messaging` was flagged as layering inversion, but `core → features` imports are established practice (135 in `core/di.ts`, 116 in `core/openapi/openapi-spec.ts`, and `base-data-view.store.ts` already had them before this branch).

## Context

The previous draft maintained two models for the same intent: a bespoke activity scope plus generic timeline filters. That produced deeply nested UI, duplicated record selection, a widget-only persistence field, and a contract MCP could not reuse. The replacement has one composable contract, and returned activity filters can be passed straight back into `get_activities`.

Permission checks stay authoritative in repositories and interactors. Indirect sources (messages, account activity, calendar) resolve relationship filters through permitted contacts; direct audit events match their mapped CRM entity via an exhaustive `Record<DomainEvent, EntityType | null>`, so adding an event without classifying it is a compile error. List and count share one `where` builder per source, expansion is cached, and the 500-contact cap fails closed.

## Validation

**Merged `origin/main` (`2aedc86b`)** which decoupled application locales from public content, added Spanish, French and Italian, turned chart data points into a `labelKind` discriminated union and tightened the i18n convention tests. Thirteen files conflicted; this branch's architecture was kept and main's semantics re-applied where the logic had moved. Locale catalogues were merged key by key against the merge base, this branch's 105 new strings were translated into the three new locales with ICU arguments verified, and reconciling the stricter registry removed two error codes that no longer had emitters, forty-seven translation keys with no consumer, and a `localeCompare` that made an activity cache key locale-dependent.

**Gates on `cc4b7b12`** — `yarn typecheck`, zero-warning `yarn lint`, **238 test files / 2,021 tests** under CI conditions (`CI=true RUN_DATABASE_TESTS=true`), `yarn build`, clean `git status` after build. All five locale catalogues are key-identical at 1,365 keys.

**Gates on `acf88c7d`** — `yarn typecheck`, zero-warning `yarn lint`, **203 test files / 1,767 tests** under exact CI conditions (`CI=true RUN_DATABASE_TESTS=true`), `yarn build`, clean `git status` after build, `git diff --check`.

**Database** — the four unreleased widget migrations were collapsed into one (`20260808160000_widget_activity_timeline`); net delta is the `WidgetKind` enum, `kind`, `timelineFilters`, three nullable chart columns, three CHECK constraints and one index. Absent configuration is now stored as real SQL `NULL` rather than a JSON `null`, which let the constraints tighten from `IS NULL OR = 'null'::jsonb` to plain `IS NULL`. Verified against a real database: fresh apply from scratch, legacy chart upgrade with omitted-kind compatibility, pre-kind rollback, and rejection of rows mixing chart and activity configuration.

**Filter semantics, against the real query engine** — all 20 relationship field × operator combinations; OR within a field proven exactly (`in [D1]`=2, `in [D2]`=2, `in [D1,D2]`=4); AND across fields proven (`dealIds hasSome`=13 ∩ `taskIds hasSome`=20 → 0); all eight activity-type combinations. Indirect resolution proven: `contactIds` returns 24 messages plus a calendar event even though `MessagingMessage` has no contact column, and `hasNone` returns audit-only while `hasSome` returns the messaging rows. Channel filtering reuses the CUS-80 contract, including its intersection rule (`provider linkedin` + `channel google` → 0, no broadening) and audit exclusion under channel/provider filters.

**Local browser** — dashboard with seven charts plus Messages and Record changes; full widget lifecycle (create → edit → reset → template → delete) with zero server errors; editor hydration of all seven `displayOptions` including non-default `barColors`, and a save that changed one field while preserving the other six; stale-ID masking rendering `Contact in Unavailable value` with no raw UUID in visible text or any aria-label and a fail-closed empty result; the refresh registry updating a mounted timeline ~1s after a record save with no reload; load-older pagination; all five entity-detail timelines; German locale including localized preview data. Record context renders in the activity detail header through the shared `AppChipStack` (moved off the timeline row in `084cac49`, because inline chips crowded the row and duplicated the row title on inbound one-to-one threads): chips deep-link to the record (verified navigating to a contact), the width-driven `+N` dropdown opens, and chips re-expand after the container widens again. Timeline rows stay title, subtitle and time.

**MCP** — omitted create `kind` → chart; update `kind` rejected; cross-kind fields rejected; omitted `timelineFilters` preserved; `timelineFilters: []` clears; value-less operators accepted without a value and rejected with one; returned filters accepted verbatim by `get_activities`.

**Deployment** — `dpl_EHLoQ3BMEoSqvFrnV3goZsX6DKai` READY on `acf88c7d`; build reports 30 migrations with none pending and a clean synthetic seed; no build errors; runtime logs show only 200/307/308 with zero error or warning entries; no Sentry issue attributable to this change.

### Not verified locally

Authenticated UI acceptance on this preview has not been run — it needs a sign-in I cannot perform. Connected-account loss and recovery, the row-detail modal opening, chart display-type switching in the browser, focus restoration, and drag/resize persistence are covered by targeted render- and state-level suites (`widget-chart` across all six display types, `widget-interaction` for click-vs-drag, `widget-layout` for geometry and clamping, `activity-widget-state` for the account-loss and permission matrix) but not by browser interaction.

Separately, a React hydration mismatch on Radix-generated ids appears in the sidebar (`NavHeader`, `SidebarMenuItem`, tooltip and dropdown triggers). This branch touches no sidebar, nav, tooltip, collapsible or dropdown file; `main`'s legal commit modified `app-sidebar.tsx`, `navigation/legal-update-alert.tsx` and `navigation-switch.tsx`. Flagged rather than fixed here.

## Impact and rollback

**Migration** — one forward-only migration adds the `WidgetKind` enum, `Widget.kind` (NOT NULL DEFAULT `chart`, load-bearing so a rolled-back reader that omits `kind` can still create widgets), `Widget.timelineFilters`, relaxes three chart columns to nullable, and adds three CHECK constraints plus `Widget_userId_kind_idx`. Every existing row is a chart, so the backfill is the default itself. No CRM or activity-source data is created, moved or deleted.

**Rollback** — before deploying a pre-`WidgetKind` build, delete only `Widget` rows whose `kind` is `activityTimeline` while the current schema can still identify them, then deploy the older reader. Chart widgets, layouts, CRM data and activity-source data are untouched. If activity-widget configuration must be preserved, ship a forward fix instead. Applied migrations are never rewritten; production migration and release remain human checkpoints.

**Environment** — no new variables. `pg` becomes an explicit `devDependency` (it was resolved transitively before) because the two new `.database.test.ts` suites import it directly, and CI gains `RUN_DATABASE_TESTS=true` so those suites actually run.

## Rebased onto the filter-operator change

`origin/main` now carries #57 (CUS-209), the relationship-existence operators this branch depends on. Merged in at `cc1f2aa9`; the four conflicts were all in the files extracted to #57 and were resolved in favour of main, which additionally carries the two fixes found while testing that PR end to end. This branch held only the pre-fix copies, so nothing from CUS-68 was lost. OpenAPI regenerated on the merged tree. Diff is now 166 files.

### The same defect existed here, in the activity schema

Testing #57 end to end surfaced a defect class worth re-checking on this branch, and it was present: `activityRelationFilterSchema` (`ee/messaging/activities/activities.schema.ts:127`) declares its `hasSome` / `hasNone` arm `.strict()`, and Zod treats a key that is present with the value `undefined` as unrecognized.

The entity-detail timeline feeds the shared filter popover straight into the activities store (`features/messaging/activities/activities-panel.tsx:51`), and that popover emits `{ field, operator, value: undefined }`. So filtering any entity's history by Contacts / Organizations / Deals / Services / Tasks with **has some** or **has none** failed validation and the panel rendered **"Activity could not be loaded"**. Reproduced in the browser on a contact detail page before the fix.

The dashboard widget path never hit it, because `activityFiltersForSave` rebuilds the filter without a `value` key — which is also exactly why the existing tests passed: they construct that same shape by hand.

Fixed at the schema boundary in `29589ea7` so the store, REST and MCP entry points are all covered, keeping strictness so `hasSome` carrying a real value is still rejected. `activity-filters.schema.test.ts` now asserts the shape the modal actually emits across all five relationship fields; **10 of those assertions fail without the fix**. Verified in the browser afterwards: `has some` and `has none` return complementary, correct result sets on the entity-detail timeline.

## Architecture-conformance audit

At Benji's request, every backend file (83 across `core/ features/ ee/ prisma/ app/api/`) was re-reviewed against `origin/main` with one question: does this follow how the codebase already does things, or does it invent something? Four reviewers produced **41 deviation claims; each was then adversarially checked for precedent. 34 died** — 24 had direct house precedent, 9 had wrong facts, 1 is mandated by the issue's acceptance criteria.

**Exonerated with named precedent** (kept as-is): `ActivityContactRepo` + the `CompanyWide` method convention (both pre-exist on main), the activities repo's internal growth (mirrors `prisma-messaging.repository`'s hydrate/resolve private-method style), the `setScope`/setter handshake (main's own activities repo already did this), the per-instance store context, `*.view-model.ts`, generation counters, sibling-cloned test harnesses, and the MCP `.strict()` tightening (main already strict-wraps every tool schema).

**Per-symbol audit of the activities repository** (`758abbd6`): after the file-level audit, every one of the **84 symbols** in `prisma-activities.repository.ts` received an individual verdict against `origin/main` — **27 pre-existing** (carried from main's own version of the file), **19 matching a named house pattern** (analog cited per symbol, mostly `prisma-messaging.repository`), **33 mandated** by a quoted CUS-68 acceptance criterion, and **5 flags**. Adversarial verification killed two flags (`NamedModel` has direct precedent in `BaseRepository`'s own `ListableModel`; the `findContactLabels` inline name idiom is main's own in the analogous role) and confirmed the `listThreads` back-fill as mandated (stale-selected-id chip labels). The three real flags are fixed: the class now declares `GetWidgetActivityFilterableFieldsRepo` in its `implements` clause (it was satisfied only structurally at the DI site; house convention names every contract), the redundant `resolvedScope` promise memo is collapsed back to main's `resolveScope` shape, and a provably unreachable fail-closed re-check is deleted.

**Fixed in `f9a4fc15`** (the 6 surviving deviations):
- `GetActivitiesApiInteractor` subclass deleted — the only decorated-interactor subclass in the codebase, double-running the tenant decorator per request. Now mode-`"api"` construction like all twelve sibling factories. Unknown keys / `p13nId` on `POST /v1/messaging/activities/search` are now ignored like every sibling endpoint (api mode never honored `p13nId`); the strict schema remains the OpenAPI/MCP documentation schema. A new test pins single decorator execution and fails on the subclass shape.
- The two overlapping precheck opt-outs collapsed into one `allowedUnavailableEntityIds: "all" | ReadonlyMap` option.
- The activity-timeline refresh registry moved from module scope onto the root store (`ActivityTimelineRegistry`, mirroring `NavigationGuardController`), removing a runtime `core → features` import from `core/base/base-data-view.store.ts` and a dead export. `core/base/base-get.schema.ts:10` still value-imports `@/features/custom-column`, unchanged from main, so this does not leave `core/base` free of runtime feature imports.
- Dead `findContactEmailsCompanyWide` deleted.
- Page ceiling enforced once (translated error only; test asserts exactly one issue).
- Pure line-wrap churn vs main reverted in three files.

**Kept by decision:** the seeded demo dashboard stays at the signed-off 7-widget layout; the demo-coverage loss of the removed "Deal Value By Type" chart is noted in CUS-216.

## Scope reduction

This PR was reviewed against CUS-68 file by file and everything not required by the issue was removed, extracted, or filed. Diff went from 187 files / +20037 -3082 to 175 files / +19206 -2860, and drops further once the dependency below merges.

### Depends on #57

The shared filter framework change (`hasSome` / `hasNone` becoming value-less relationship-existence operators, `filter-compat.ts`, the URL codec, and the REST/MCP contract shape on nine non-activity endpoints) is not CUS-68 work and now lives in #57 / CUS-209. **#57 merges first, then this branch takes it in.**

It cannot simply be reverted here: this widget's own filter editor renders through `isStandaloneOperator` and `hasValidFilterConfiguration`, so without it every value-less activity filter is treated as invalid and silently dropped from the widget's chips and counts. #57 also fixes a real regression the original change carried, where `hasSome` on a custom column lost its `columnId` scoping and matched any record holding a value for any custom column.

### Reverted here

| What | Why | Issue |
| --- | --- | --- |
| Saved filter presets validated on read, whole preset dropped on any failure | Silent loss of user data across six entity types | CUS-210 |
| `dealFilters` cleared on read and written back as `[]` | Silent loss of stored chart configuration; the UI already hid them | CUS-211 |
| `bulkDelete` and the delete-confirmation `Promise<boolean>` contract | Out of scope; reverted to main's contract across six files | CUS-212 |
| Seven inert props on `AppChipStack`, `FormSwitch` and `FormAutocomplete` | No caller ever passed any of them | |
| The repo-wide sweep in `overlay-contract`, plus the CI-cannot-skip guard and the `pg-connection-string` re-parse (`ui-tabs` remains as a two-file pin on `tabs.tsx` and the widget editor, not a repo-wide sweep) | Repo policy, not this feature. The loopback check stays: it is what makes the new destructive database suites safe | CUS-215 |
| Half of `dashboard-widget-ui.test.ts` | Pinned discarded intermediate implementations, exact occurrence counts and tailwind geometry rather than behaviour | |
| Editor formatting churn in six `.mdx` docs, the contact repository and two prisma fixture tests | No prettier config for mdx and no format step in CI; `en/mcp.mdx` alone went from 154 changed lines to 28 | |

### Kept deliberately

- **`getWidgetById` ownership scoping** (CUS-213). A genuine cross-user read fix. Shipping the branch without it would be worse than carrying it, but it is filed so it gets reviewed on its own.
- **Shared design-system changes** (CUS-214) — the `form-autocomplete` rewrite and its thirteen consumers, `filter-input-select`, the `tabs` `segmented` variant, the `tabs` `line` variant, `AppModal`'s new sizes, `selectable-card`, the `ENTITY_ICON` move and the `unavailableValue` copy change. All user-visible on pages outside this feature, so reverting them would change the look of unrelated screens.
- **Demo chronology re-dating** (CUS-216) — the signed-off demo dashboard needs a year of history to page through. The dates are absolute and already stale, which is the actual problem.

### Corrections found while doing this

- The `Enter`-to-create path in `FormAutocomplete` was reported as a regression. It is not: cmdk's create `CommandItem` handles it, and it now has a test proving it fires exactly once.
- `Common.table.columns.provider` was reported as an orphan key. It is not: it labels the `FilterFieldKey.provider` filter field in this widget's own editor.

## Rebased onto the page-state architecture (#58) — head `d0ec045e`

This replaces all earlier evidence; every result below is from the exact head `d0ec045e`, rebased onto `520cd53c`.

**Rebase.** The branch was squashed to one commit before rebasing (tree hash verified identical to the pre-rebase state) and then replayed onto current `main`. This was deliberate: replaying 47 commits, 6 of them merges, across #58's restructuring is the scenario that previously lost code in auto-merged regions on this repo. One commit means one deliberate conflict resolution instead of 41 blind replays. A `cus68-prerebase-backup` tag preserves the original.

**Conflicts (8), resolved preserving both sides.**
- `widgets-grid.tsx` → `dashboard-page-view.tsx`: main's selector + exhaustive switch + `PageState background` kept; the CUS-68 grafts re-applied (activity filterable fields, interactive-target opt-out via `isInteractiveTarget`/`isWidgetOpeningClick`/`openWidgetEditor`, `draggableCancel`).
- `base-data-view.store.ts`: main's request-state bookkeeping and the timeline fan-out are additive — both kept at all four sites.
- `activities.store.ts`: the store's hand-rolled generation counter collided with the base's private one and duplicated its job. Removed; the store now commits through one gated `setItems` override, so stale results are suppressed by the base. `refreshFor` keeps main's explicit rejection handling.
- `data-view-empty.tsx`, `vitest.config.ts`, and both add/add test files merged so that main's suites and this branch's suites both survive.

**Failure contract.** An initial failure commits an empty error state; a subsequent failure rejects, so content is retained and fire-and-forget callers toast. That satisfies both "retained content on refresh failure" and "explicit rejection handling".

**Architecture adoption.** `EntityTimelinePanel` and `ActivityWidgetCard` now derive state through pure selectors and render one exhaustive switch with a `never` check. `connected-accounts.store` drops its parallel `refreshError` (removed from the base by #58) and reads `dataRequest`.

**Defect found and fixed during verification.** The company-wide feed surfaces audit events for webhooks, roles, and connected accounts, whose field names have no `Common.table.columns.*` entries — so rows rendered raw keys such as `Common.table.columns.enabled` and logged `MISSING_MESSAGE`. Entity timelines never hit this because they are scoped to CRM records. Added the 11 missing keys across all five locales and extended the curated allowlist. The dashboard now renders "Enabled · Description"; zero raw keys remain in the rendered output.

**Legacy scan.** `DataViewContainer`, `PageSkeleton`/`PageSkeletonSpec`, `RouteLoading`, the route registry, `PageState.skeleton`, the data-view barrel, and the `*Card` orchestrators return zero production hits; the only matches are main's own guard test, i18n key namespaces, and the unrelated `EntityDetailRouteLoading`.

**Gates at `d0ec045e`.** typecheck 0 errors · lint 0 warnings · `yarn test` 2293 passed (the 5 `registration-real-db` cases pass under `RUN_DATABASE_TESTS=true` with `DATABASE_URL`, verified loopback) · database-enabled suites 348 passed across 30 files · `yarn vercel-build` success · 30 convention files / 230 tests pass, including main's `page-state-contract` guard.

**Browser at `d0ec045e`.** Dashboard loading (route `loading.tsx` → `DashboardPageSkeleton`), content (7 widgets incl. the activity timeline), entity-detail timeline, light and dark themes, mobile 375×812, and console output free of new errors. Reduced motion is covered by the global `[data-page-skeleton-loading] [data-skeleton-motion]` rule plus `motion-reduce:animate-none` on the grid.

**Rebase drop-check.** Because a squash-and-rebase can silently lose code in auto-merged regions, seven independent read-only audits compared the pre-rebase tag `cus68-prerebase-backup` against `d0ec045e` across the whole feature (backend activities, widget domain, dashboard UI, timeline UI, shared filters, API/MCP/docs/i18n, tests). **1448 symbols compared, zero losses.** Several areas proved it at tree level rather than symbol level: `git diff cus68-prerebase-backup HEAD -- app/api/ features/mcp-tools/ core/openapi/ content/docs/ i18n/locales/ public/v1/openapi.json` is 55 insertions and zero deletions, and the backend-activities and shared-filters feature diffs are byte-identical pre- and post-rebase. The only two files that left the feature diff are accounted for: `widgets-grid.tsx` was renamed by #58 to `dashboard-page-view.tsx` (all five CUS-68 grafts verified present there), and `connected-accounts.store.ts` is now identical to main because its only remaining delta was the `refreshError` field #58 deleted.

## Review round: six questions answered, four changes made (`f7c589a0`)

Six critical questions were investigated against actual code and the knowledge base, each answer adversarially re-verified. Three concerns were **not** substantiated and three were.

**Not substantiated.** (1) *"CUS-68 creates a specific API interactor."* It does not: `getGetActivitiesApiInteractor` already exists on `origin/main` as one of **13** such factories, none of which is a class, and there are zero `class *ApiInteractor` definitions in the repo. CUS-68's change is two lines switching this route to the factory the MCP tool already used. (2) *"The record-options interactor is unnecessary."* Main's own `technical-id-loading-contract.test.ts:70-71` bans both cheaper alternatives, nothing existing resolves a UUID to a label, and it is symmetric with `get-activity-thread-options.interactor.ts` which predates the branch. (3) *"The change set is misaligned with the knowledge base."* No machine-enforced convention is broken; the KB defines no size rule, and only 34% of additions are hand-written product code.

**Substantiated and fixed.**

1. **Database CHECK constraints removed.** Zero of the 30 pre-existing migrations use a CHECK, a trigger, a function or a rule; every prior `ADD CONSTRAINT` is a Prisma foreign key. The three constraints were also redundant: `UpsertWidgetInteractor` validates every write through `@Write({ input: Schema })` against the discriminated union whose chart arm requires `entityType`/`groupByType`/`aggregationType`, and only two production writers exist, both behind it. The migration is now 23 lines of Prisma-shaped DDL. The mixed-configuration guarantee is covered by six existing interactor, schema and MCP tests, so the database-level test that duplicated it was removed rather than rewritten.

2. **Parallel request state collapsed.** `ActivitiesStore.filtering` and `.loadError` duplicated `refreshing` and `refresh-error`, which made the shared `resolveResourcePageState` unusable on this store and forced two surfaces to hand-roll their own state unions. `applyFilters` now drives the visible-mode `refreshQuery()` once the store is ready, a failed refresh rejects so the base records `refresh-error` and retains the rows already shown, and only a narrowly-scoped `olderPageError` remains for the load-older path.

3. **Two user-visible defects fixed.** The widget-editor preview rendered "No activity matches the current filters" when the server actually failed; it now derives state through `resolveResourcePageState` and renders one exhaustive switch with a real error branch. Both `void store.applyFilters(...)` call sites lacked a rejection handler and would raise an unhandled rejection once a refresh failure rejects.

4. **PageState composition adopted.** A colocated, server-compatible `ActivityTimelineSkeleton` (built from `SkeletonShape`) now backs the widget card's loading state and both the card's and the entity panel's true-empty states through `PageState background={...}` — the same skeleton animated for loading and `animated={false}` when empty. It is registered in `page-state-contract.test.ts`'s pure-skeleton guard, covered by a render test proving both variants are structurally identical, and its wiring is pinned.

5. **Shared overlay behavior reverted.** `app-modal.tsx` and `delete-confirmation-modal.tsx` no longer change dismissal for every modal in the app; only the additive `description` prop and the `3xl`/`5xl` sizes the widget editor needs remain.

Gates at `f7c589a0`: typecheck 0 errors, lint 0 warnings, 2298 tests passing, 347 database-enabled tests across 30 files, `yarn vercel-build` success.

## Minimality pass on the shared widget UI (`2fa592b8`)

A file-by-file duplication review of the dashboard surfaces. Every item removes code without changing what a user sees; the two behavioural deltas are called out below.

**Restored the shared component main already used.** `origin/main`'s widget editor footer was `<FormActions showInitially anchorScope="widget-modal" store={widgetModalStore} />`. This branch had replaced it with a hand-rolled `WidgetEditActions` component plus a duplicated `requestClose` and a footer Cancel button no other modal in the repo has — the five `FormActions` modals all rely on the overlay's own X, Escape and outside-click. `FormActions` is back, `WidgetEditActions` and `requestClose` are gone, and the Cancel button with them. This also restores the `anchorScope="widget-modal"` contract that `anchor-ids.test.ts` requires, which the hand-rolled version had silently broken by emitting the ids as literals.

**One filter chip instead of two.** `chart-widget-card.tsx` and `activity-widget-filters.tsx` rendered the same interactive `FilterChipValue` wrapper. They now share `widget-filter-chip.tsx`. This completes the interactive opt-out the rebase left half-done: the chart chip stopped grid drags only via `onPointerDown` propagation, so it was invisible to `draggableCancel`; both chips now also carry `data-widget-interactive`.

**One progress bar.** The widget creation steps and the onboarding wizard each had their own. `components/shared/wizard-progress.tsx` serves both, taking its label and value text as props so no new translation key was needed for the widget editor. The onboarding wizard gains the `progressbar` role and accessible name it lacked (`OnboardingWizard.progressLabel`, 5 locales); its geometry is unchanged.

**One aggregation label.** The 4-branch aggregation-label rule was written twice, in the editor's metric select and in the preview's screen-reader summary. `useAggregationTypeLabel()` is now the single source and is exhaustive over `AggregationType`, which removed the `t(translationKey)` non-literal site and its two `i18n-key-resolution.test.ts` registrations.

**Dead code and hand-written lists.** The three activity type cards derive from `ACTIVITY_TYPE_VALUES` instead of being enumerated by hand; the widget editor's grid template literal is one shared constant; the `summaries` array's separator branch was unreachable after an acceptance criterion reduced it to one value; `renderFilterSettings` only called `renderChartFilters`; `filters.map((filter, index) => …)` ran twice in `activity-filter-fields.tsx`; `isWidgetOpeningClick`'s `tolerance` option and `SelectableCard`'s `className` prop were never passed; and both widget cards wrapped a lone `<h2>` in a flex container, which is now back to main's markup.

**Formatting reverts finished.** The last of the 15 formatting-only hunks flagged earlier is gone (`auditCategory`'s return type in `activities-row.tsx` re-wrapped across four lines); `eslint` accepts the original single-line form, confirming the explosion was gratuitous. Zero formatting-only hunks remain in the diff.

Two convention pins moved with the design rather than being deleted: the widget-modal pin now asserts the shared `FormActions` and the absence of a local `requestClose` or Cancel label, and the activity-metadata pin asserts `<WidgetFilterChip>` in both cards.

Net effect on the diff: **−178 insertions, +21 deletions** against `origin/main`.

**Verified.** typecheck 0 errors; `yarn lint` 0 warnings; full suite 2298 passing with only the 5 known `registration-real-db` failures that need a live database; convention suite 414 tests green. Browser on `localhost:4000`: activity chip and chart chip both deep-link into the editor at the right tab and expanded filter; edit footer shows Save alone when clean and Save + Reset when dirty, and Reset reverts the form and the live preview; create flow shows step 1 of 2 with the bar at 50% and no footer, step 2 with the bar full and Back + Create, and Back returns to step 1; all three metric labels resolve (`Count Services`, `Deal value related to Service`, `Number of Services in Deals`); mobile drawer renders the same footer with a full-width Save; dark theme renders headings, chips and feed correctly; no console errors.

**Three findings from the same review were refuted rather than acted on**, because the precedent went the other way. The widget modal's body ternary chain is the house idiom for modals (`api-key-modal.tsx` nests three), not a `page-state` violation — the selector + exhaustive switch rule is for page views. `WidgetModalSkeleton` using raw `<Skeleton>` rather than `SkeletonShape` matches `app-topbar.tsx` and `selection-loading.tsx`; `SkeletonShape` is used exclusively by route-level `*PageSkeleton` files. And the widget card's local empty state is not duplicating `TimelineEmptyState`: one fills and centres inside a card body, the other is a bordered standalone card, so merging them would change one surface visually.

## Registry contract collapsed to one method (`dbc12575`)

Follow-up to a review question about `ActivityTimelineRegistry`. The pattern itself is not new: `NavigationGuardController` on `origin/main` is the same RootStore-held class with a private `Set` of live stores and the same `register`/`unregister` pair, subscribed from the same `useEffect` shape (`components/modal/use-navigation-guard.ts:14-17`); `RootStore.modalStores` + `closeAllModals` is the same iterate-and-fan-out; and `BaseDataViewStore.onChangesCallbacks` is a third subscribe/fan-out registry. Nor is the behaviour new: `origin/main` already refreshed a timeline after a record mutation, through a lazy singleton (`root.store.ts:85,204`) and a direct `activitiesStore.refreshFor(id)` call (`base-custom-column-entity-modal.store.ts:236`). This branch deletes both, because the dashboard mounts one store per activity widget and a singleton with one `timelineEntityId` cannot serve N differently scoped timelines. The direct call was replaced in place: `upsertItem`, one line above the deleted call, now carries the fan-out.

What the audit did find was a redundancy. `RegisteredActivityTimeline` declared `coversEntity` **and** `refreshFor`, so every fan-out ran the scope check twice, once in the registry to select a matching id and again inside `refreshFor`. The selected id was then discarded, because `refresh()` re-queries the store's own scope and the entity id carries no information into the fetch. One `refreshIfCovers(entityType, entityIds)` replaces both: the interface loses a method, the registry loop loses its `find()`, and the misleading single-id parameter is gone. `coversEntity` remains a public method of `ActivitiesStore`, where it is unit-tested directly. Behaviour is unchanged and still pinned by `activity-timeline-refresh.test.ts`, which asserts one refresh per matching store for a 100-id batch across four mounted timelines.

**Deliberately not changed.** The dashboard widget mounts unscoped (`activity-widget-card.tsx:38` passes `{}`), and `coversEntity` returns `true` unconditionally without a scope, so a widget filtered to specific records would still refetch on unrelated mutations. This is currently unreachable: from the dashboard every route into a CRM mutation navigates away first, since the activity detail modal's record chips are plain `<a href>` links and nothing pushes onto the entity drawer stack over the grid, so the widgets unmount before any broadcast arrives. Rather than add speculative scope-derivation code to this branch, it is filed as [CUS-224](https://linear.app/customermates/issue/CUS-224/scope-dashboard-activity-widgets-so-they-stop-matching-every-crm) with the exact trigger conditions. The `.catch(() => toastError(...))` in the refresh is likewise left alone: it is inherited verbatim from `origin/main` and changing it would alter behaviour beyond this widget.

**Verified at `dbc12575`:** typecheck 0 errors; `yarn lint` 0 warnings; full suite 2297 passing. Two failures are environmental, not from this change, and both reproduce with the change stashed: the 5 known `registration-real-db` cases need a live database, and `docs-mcp-tools` needs a generated OpenAPI manifest this verification clone cannot build.

## `use-filter-select-items.tsx` review (`bbbc8ad3`)

Reviewed on request. **The architecture holds; two behavioural defects did not.** The three things that prompted the question were each checked against `origin/main` and each is forced or precedented:

- **The store to context swap was forced.** The hook read `useRootStore().activitiesStore`, a singleton this branch deletes because timelines are now component-owned and multi-instance. Reverting to a store would be wrong. Nullable-context branching is already the idiom in this exact directory (`components/forms/form-context.tsx:18`, consumed at `filter-input-select.tsx:33,48,50,111`), and prop-drilling the alternative would push an activity-specific prop onto the generic `FilterPopover` and `DataViewActiveFiltersBar`.
- **The `@/ee/` imports extend an edge that already existed.** `git grep "@/ee" origin/main -- 'components/**'` returns two hits, and one of them is a value import in this very file (`getProviderIcon`, `:18` on main). There is no rule to violate: `.eslintrc.json` has no `@/ee` entry in `no-restricted-imports`, no convention gate covers import direction, and the licensing boundary explicitly permits AGPL code under `components/` to link against `ee/`. Both new value imports are isomorphic (a numeric constant and a Map lookup) and nothing under `ee/messaging/activities/` is `server-only`.
- **Type-importing from an `*.interactor.ts`** is the best-precedented thing in the file: line 1 already does it on main (`GetResult` from `@/core/base/base-get.interactor`), as do four other `components/` files.

Also cleared after tracing: `validActivityFilters` is right to validate element-wise rather than array-wise, because the arrays reaching it routinely carry placeholder rows and a whole-array parse would silently discard a configured `connectedAccountId` scope; the client-side `z.uuid()` guard and the silent drop are correct, since malformed, stale and deleted ids all converge on `Common.inputs.unavailableSelection` and a raw id is never rendered; and `maxSelectedValues` keying on context rather than field is inert, because the cap is a correct upper bound for all nine reachable fields and the two where it binds are exactly the ones `activityEntityTypeForFilterField` does not cover.

**Fixed: resolved chips flickered and refetched on unrelated filter edits.** `resolveItems` wrapped `getSelectedItems` in a fresh async arrow and depends on `getItems`, which rebuilds whenever `timelineScopeKey` changes. Only the `timelineThreadId` entry reads that key, but every field paid for it. Toggling an activity-type card in the editor changed the key, produced a new resolver, missed the three `selectionResult.resolver` guards, flipped `isLoading` and refired the effect: already-resolved organization chips fell back to skeletons while a second `getActivityRecordOptionsAction` ran. `contactIds` escaped only by accident, because it maps to a module-level constant whose identity never moves. `getSelectedItems` now returns `FilterSelectItem[]` and `resolveItems` returns it unwrapped, so the reference survives the churn. `GetSelectedItemsFunction` goes with it, 12 lines removed for 6.

**Fixed: selected contact chips lost their avatar.** `listRecordOptions` returned `{ ...ref, label }` and dropped the `avatarUrl` it had already loaded, while the sibling record-context path forwards it. `ActivityRecordRefDto` declares the field, so `@ValidateOutput` and `tsc` stayed silent; the visible symptom was a chip with no photo sitting next to a dropdown row for the same contact that had one. One line in the repository. The mocked repository test now gives one contact a real avatar, so it fails against the previous behaviour.

**Deliberately left.** `activityQueryRef` (`:108-109`) is provably redundant today, because `timelineScopeKey` already serializes exactly the payload the closure sends. It stays because that safety depends on the memo key continuing to encode the full payload, and the ref removes the coupling for 3 lines. Note `react-hooks/exhaustive-deps` is `"off"` repo-wide (`.eslintrc.json:222`), so lint would not catch a future divergence.

**Filed:** [CUS-225](https://linear.app/customermates/issue/CUS-225/cover-the-activity-record-label-resolution-path-with-behavioural-tests) for the coverage gap that let both defects ship. Both consumer test files mock the hook outright, and the only gate naming the new code asserts that three strings appear in the file.

**Verified at `bbbc8ad3`:** typecheck 0 errors; `yarn lint` 0 warnings; full suite 2311 passing with only the 6 environmental failures that need the live database and generated OpenAPI manifest CI provides.

## Record-options action review (`6c075280`)

Reviewed on request: is a dedicated action for resolving activity filter labels an established pattern?

**It is not an endpoint.** `GetActivityRecordOptionsInteractor` has no REST route, no OpenAPI fragment and no MCP tool. It is reached through one server action whose only caller is the `getSelectedItems` memo in `use-filter-select-items.tsx`, so the public surface is unchanged. Server-action-only interactors are unremarkable here: the repo has 213 `*.interactor.ts` files against 58 REST routes, and the sibling `getActivityThreadOptionsAction` has no REST or MCP surface either.

**The shape is house-standard; the job is new.** Thirteen interactors on `origin/main` carry the same conjunction of in-file abstract repo, `AuthenticatedInteractor`, `@AllowInDemoMode`, `@TenantInteractor`, `@Validate` and `@ValidateOutput`. The closest structural twin is not the folder sibling but `features/custom-column/get-custom-columns-by-entity-type.interactor.ts`, which matches line for line at the same 39-line length. The job, however, has no precedent: nothing on `origin/main` takes a caller-supplied id set and returns display labels. `get-activity-thread-options` predates the branch but **lists** available options rather than **resolving** known ids, so it is not the analogy an earlier revision of this PR body claimed.

**Why it is close to forced.** An id-`in` filter against the existing list actions is inexpressible: `git grep 'field: "id"'` returns zero hits repo-wide, `FilterFieldKey` has no `id` member, and `core/base/base-query-builder.ts:551` silently drops filters whose field is not in the repository's `filterableFields`, so the request would have returned an unfiltered first page with a 200 and no error. The generic fallback cannot serve either: `use-filter-select-items.tsx` calls `getItems({})`, and `base-query-builder.ts:141` resolves an absent pagination argument to `take: 100`, a value that cannot be widened because `PaginationRequestSchema.pageSize` is the literal union `5|10|25|100`.

Correcting an earlier claim in this body: this does **not** repair a pre-existing bug on the activity timeline. `activity-filterable-fields.ts` does not exist on `origin/main`, and the timeline exposed only `timelineKind`, `timelineThreadId`, `provider` and `connectedAccountId` there, so the five relationship fields ship together with their label resolution. The 100-row truncation is nevertheless a live defect on other views: `contactIds` is a filterable field on deals (`prisma-deal.repository.ts:125`) and organizations (`prisma-organization.repository.ts:112`), and both still resolve their chips through the capped fallback after this branch, because `getSelectedItems` is gated on `hasActivityQuery`. That is worth its own issue and is not an argument against this action.

**Permissions verified rather than assumed.** `listRecordOptions` delegates to the same private `resolveRecordLabels` the timeline uses at three other call sites; every leaf query applies `accessWhere`, which `core/base/base-repository.ts:46-66` resolves to `{ companyId }` on read-all, an owner predicate on read-own, and a fail-closed `{ id: { in: [] } }` with no permission. Unresolvable refs are dropped rather than surfaced, so there is no existence oracle. The database test proved omission only for a nonexistent id, while the cross-tenant contact the file already seeds was exercised on the timeline path alone; it is now asserted on this path too, and CI ran it against a real database.

**Left deliberately.** The input schema uses `ActivityScopeRecordGroupsSchema.min(1)` without the group-merging `.overwrite` that `ActivityScopeSchema` wraps around the identical shape, so five groups of one entity type accept five times the per-type cap. It is hard-bounded at 250 ids on an indexed `id IN` lookup, the sole caller dedupes and self-caps at `ACTIVITY_FILTER_VALUE_MAX`, and closing it means duplicating the merge logic.

**Worth a separate decision.** `tests/conventions/technical-id-loading-contract.test.ts:68-71`, unchanged by this branch, still requires the `getItems({})` fallback to exist and bans two literal strings, while this branch's own `dashboard-widget-ui.test.ts` ratifies the server-side resolver. Both pass at once because the ban is a substring check on an identifier that has never existed as implementation code. Nothing is broken, but there are now two rules for one job and the older one pins the mechanism that truncates at 100.

## Precheck options bag removed (`c68e25a4`)

Challenged in review: does `QueryParamsPrecheckOptions` belong on this PR, and shouldn't the query check be unified without options? The challenge was right on both counts, and the option is gone. `core/base/query-params-precheck.interactor.ts` is byte-identical to `origin/main` again.

**The `get_activities` caller was a strictness regression, not a new allowance.** It passed `allowedUnavailableEntityIds: "all"`, which disabled entity-id validation outright. `origin/main` already validated entity ids on that path: `timelineThreadId` and `connectedAccountId` are both `entityId`-kind and both were filterable there. This branch additionally repoints `app/api/v1/messaging/activities/search/route.ts` from the interactive interactor to the api one, so the relaxation reached REST as well as MCP. Ids that main rejected became silent empty 200s on both surfaces. Nothing pinned the behaviour, and no feature needed it: the precheck runs only in api mode, while the dashboard widget and all five entity-detail timelines use the interactive interactor and never reach it. The only justification was a sentence this PR wrote into its own tool description and both `mcp.mdx` locales, which are corrected here.

Retaining unresolvable ids is also not reliably fail-closed. A `notIn` rule carrying an unknown id widens the result rather than narrowing it, so rejecting the request is the safer answer as well as the consistent one.

**The `upsert-widget` caller has a real need, but it does not belong on a shared interactor.** A widget whose saved filter references a since-deleted contact must stay editable; otherwise editing a different filter row fails on a value the user never touched. Correcting an earlier description in this body: this is not the rename case, because a name-only save omits `timelineFilters` and returns before the precheck runs. The interactor now hands the precheck a throwaway filter array with retained ids stripped, so new ids are still validated and stale ones are not. Behaviour is identical, including the emitted issue path, and the five `upsert-widget` tests that pin retain, remove, and reject-a-new-id-beside-a-retained-one still assert it, one of them on the exact validator argument.

Net **+18 / −110**, including deleting the option's dedicated test, whose two cases are now either covered by the widget tests or gone with the option.

**Filed rather than fixed:** [CUS-230](https://linear.app/customermates/issue/CUS-230/filter-id-validation-errors-point-at-the-wrong-value-index). `check-ids.ts` derives each issue path from the index of the array it receives, so a compacted list reports the error at the wrong chip. The removed option had this defect and the caller-side pre-filter inherits it; fixing it properly is incompatible with any per-value relaxation and needs the validator to carry original indices.


benjiwagner:
Exact-head verification is complete for `47284b08` (rebased onto `7278afa5`).

- Full suite: 186 files passed / 3 skipped; 1,552 tests passed / 26 skipped.
- Typecheck, zero-warning lint, production build, and diff checks pass.
- The disposable local dashboard was reset at the widget/layout level only—CRM data was untouched—and now contains seven colorful charts plus **Messages** and **Record changes** timelines.
- Local wide/narrow browser acceptance passed, including exact activity header/body alignment, responsive stacking, value-less operators, and all requested editor interactions.
- [Exact-head Vercel preview](https://feat-cus-68-activity-timeline-widget.customermates.com/en/dashboard) is READY; authenticated smoke passed with zero browser errors and no 5xx runtime entries.
- CI, repository policy, CodeQL, and Vercel all pass.
- Final read-only semantic and migration audits found no remaining findings.

The PR remains draft for human product review and merge/release remain human-only.

.github/workflows/test.yml
```diff
@@ -15,6 +15,8 @@ jobs:
     env:
       APP_MODE: self-hosted
       DATABASE_URL: "postgresql://postgres:postgres@localhost:5432/customermates_test"
+      ACTIVITY_RELATIONSHIP_TEST_DATABASE_URL: "postgresql://postgres:postgres@localhost:5432/customermates_test"
```
benjiwagner: Why do we need these two?

app/[locale]/(protected)/company/components/company-invite/company-invite-modal.tsx
```diff
@@ -47,7 +47,7 @@ export const CompanyInviteModal = observer(() => {
 
         <AppCardBody>
           <Tabs defaultValue="link">
-            <TabsList>
+            <TabsList variant="default">
```
Comment (benjiwagner) [inline app/[locale]/(protected)/company/components/company-invite/company-invite-modal.tsx]: Redefining the default variant is a bit weird? Do we need this?

app/[locale]/(protected)/company/components/role/role-modal.store.ts
```diff
@@ -88,19 +88,20 @@ export class RoleModalStore extends BaseModalStore<UpsertRoleData> {
   };
 
   delete = async () => {
-    if (!this.form.id) return;
+    if (!this.form.id) return false;
```
Comment (benjiwagner) [inline app/[locale]/(protected)/company/components/role/role-modal.store.ts]: Why do we need to return a boolean now instead of void?

app/[locale]/(static)/pricing/components/pricing-section.tsx
```diff
@@ -49,7 +49,7 @@ export function PricingSection({
             value={isAnnual ? "yearly" : "monthly"}
             onValueChange={(key) => setIsAnnual(key === "yearly")}
           >
-            <TabsList>
+            <TabsList variant="default">
```
Comment (benjiwagner) [inline app/[locale]/(static)/pricing/components/pricing-section.tsx]: Do we need to redefine the variant default? If not needed. Can we add a linter rule so we dont do this in the code?

prisma/schema.prisma
```diff
@@ -696,29 +698,39 @@ model ContactOrganization {
 }
 
 model Widget {
-  id                    String            @id @default(uuid())
-  userId                String
-  companyId             String
+  id                    String             @id @default(uuid())
+  userId                String?
+  companyId             String?
+  rollbackUserId        String?
```
benjiwagner: Why do we need the rollback stuff?

core/base/base-custom-column-entity-modal.store.ts
```diff
@@ -198,7 +198,6 @@ export abstract class BaseCustomColumnEntityModalStore<
         if (this.fetchedEntity) this.fetchedEntity = res.data;
         this.onInitOrRefresh(this.initFormWithCustomFieldValues(res.data));
         if (isCreate) this.lastCreatedId = res.data.id;
-        this.rootStore.activitiesStore.refreshFor(res.data.id);
```
benjiwagner, line 236: Why did we remove this? Do we not need this?

app/[locale]/(protected)/company/components/company-invite/company-invite-modal.tsx
Comment (benjiwagner) [reply app/[locale]/(protected)/company/components/company-invite/company-invite-modal.tsx]: Agreed. `TabsList` already defaults to the standard variant, so I removed the redundant prop in `fd1eab65`.

app/[locale]/(static)/pricing/components/pricing-section.tsx
Comment (benjiwagner) [reply app/[locale]/(static)/pricing/components/pricing-section.tsx]: Agreed here too. I removed the redundant prop. Rather than add a brittle component-name-specific ESLint selector, I extended the existing Tabs convention test to reject explicit `variant="default"` across consumers while keeping intentional `segmented` and `line` variants explicit.

.github/workflows/test.yml
benjiwagner: Good question. These were not separate databases; they duplicated `DATABASE_URL` only as opt-in markers for suites that otherwise skip. I replaced both with one explicit `RUN_DATABASE_TESTS=true` switch. All three suites now use the existing `DATABASE_URL` through a shared guard that validates the effective `pg` host is loopback, including rejecting `?host=` overrides. Local runs still skip unless explicitly enabled; CI now fails on a missing or unsafe URL instead of silently passing. The guard and database suites pass on `fd1eab65`.

benjiwagner:
Final review-fix verification for `fd1eab6577c4dc36b7b5a8ebd741967156ae9478`:

- Replaced the two duplicated CI database aliases with `RUN_DATABASE_TESTS=true` plus one shared, loopback-only effective-host guard over the existing `DATABASE_URL`.
- Removed redundant default Tabs props and added repository convention coverage.
- Removed the temporary rollback-quarantine schema through a forward migration that restores complete ownership, rejects partial state, and returns Widget ownership to the established required model. The documented pre-kind rollback now deletes only `activityTimeline` widget configuration rows; charts and CRM/activity data are preserved.
- Kept the old singleton refresh removal and fixed its replacement so matching message/account/calendar-only timelines also invalidate after entity or relationship mutations.

Verification:
- 196 test files / 1,690 tests passed with database suites enabled.
- 46 focused database/guard tests passed against the isolated local PostgreSQL container.
- Typecheck, zero-warning lint, production build, Prisma validate/generate, and diff checks passed.
- The local Max Bergmann database was reset/reseeded through every migration; signed-in dashboard smoke passed with all 9 seeded widgets and no console errors.
- GitHub checks are 7/7 passing.
- Exact-head Vercel deployment `dpl_13p1HN3TB9YWMUq2a7obUUogZ83z` is READY at the [isolated preview](https://feat-cus-68-activity-timeline-widget.customermates.com/en/dashboard); the final runtime audit found no warnings, application errors/fatals, or 5xx responses.

All five open review threads have replies. I left them unresolved for reviewer confirmation.

app/[locale]/(protected)/dashboard/components/chart-widget-card.tsx
```diff
@@ -0,0 +1,159 @@
...
+type Props = {
+  widget: ChartWidgetDto;
+};
+
+export const ChartWidgetCard = observer(({ widget }: Props) => {
+  const t = useTranslations();
+  const filterFieldLabel = useFilterFieldLabel();
+  const { intlStore, widgetModalStore, widgetsStore } = useRootStore();
+  const customColumns = widgetsStore.customColumns;
+  const dealCustomColumns = useMemo(
```
Comment (benjiwagner) [inline app/[locale]/(protected)/dashboard/components/chart-widget-card.tsx]: So in terms of the components for the widget stuff, make sure that we have everything stateful that makes sense in the store. For example, all these memorizations can also easily be part of the store, if I'm not mistaken. Don't just guess, deeply go through the front-end related files again for proper state management with MobX.

app/[locale]/(protected)/profile/components/connected-accounts.store.ts
```diff
@@ -54,9 +57,21 @@ export class ConnectedAccountsStore extends BaseDataViewStore<ConnectedAccountDt
     return this.items.filter(accountNeedsAction).length;
   }
 
+  override setItems(args: GetResult<ConnectedAccountDto>): void {
+    super.setItems(args);
```
Comment (benjiwagner) [inline app/[locale]/(protected)/profile/components/connected-accounts.store.ts]: It looks like we are inventing a new pattern here, which we usually don't like.

app/api/v1/messaging/activities/search/route.ts
```diff
@@ -3,13 +3,16 @@ import type { NextRequest } from "next/server";
...
-import { getGetActivitiesInteractor } from "@/core/di";
+import { getGetActivitiesApiInteractor } from "@/core/di";
 import { handleError } from "@/core/api/interactor-handler";
+import { ActivitiesApiParamsSchema } from "@/ee/messaging/activities/activities.schema";
 
 export async function POST(request: NextRequest) {
   try {
-    const data = await request.json();
-    const result = await getGetActivitiesInteractor().invoke(data);
+    const parsed = ActivitiesApiParamsSchema.safeParse(await request.json());
```
benjiwagner: This does not look like the typical behavior we usually use. So we are having the schemas usually on the interactors and not within the routes. So, therefore, please double check what the common way to do this is and update it.

components/data-view/filter-modal/inputs/use-filter-select-items.tsx
```diff
@@ -145,79 +166,119 @@ export function useFilterSelectItems(
           }),
         })),
       [FilterFieldKey.timelineThreadId]: () => {
-        const { timelineEntityType, timelineEntityId } = activitiesStore;
-
+        const filters = validActivityFilters(activityQuery?.filters);
```
benjiwagner: This also looks like something that is inventing new pattern. Please double-check how we do it everywhere else and simplify accordingly.

components/data-view/filter-modal/edit-filters-modal.store.ts
```diff
@@ -47,6 +47,8 @@ export class EditFiltersModalStore extends BaseModalStore<UpsertFilterPresetData
     this.expandedField = field;
   };
 
+  owns = (store: unknown): boolean => this.tableStore === store;
```
benjiwagner: Don't quite get what we need this for.

core/di.ts
```diff
@@ -928,6 +935,8 @@ export const getGetWidgetFilterableFieldsInteractor = () =>
     getDealRepo(),
     getServiceRepo(),
     getTaskRepo(),
+    new PrismaActivitiesRepo(),
```
benjiwagner: This looks like a pattern we usually don't follow.

.github/workflows/test.yml
benjiwagner: Down to one line now: `RUN_DATABASE_TESTS: "true"`. It is load-bearing rather than convenience — the shared database-test guard throws when `CI` is set and this flag is not `true`, so the database suites cannot skip unnoticed in CI. The second addition you were looking at is gone, reverted in the minimality pass.

app/[locale]/(protected)/company/components/company-invite/company-invite-modal.tsx
Comment (benjiwagner) [reply app/[locale]/(protected)/company/components/company-invite/company-invite-modal.tsx]: Agreed, and it is gone — `company-invite-modal.tsx` is no longer touched by this branch at all (`git diff origin/main...HEAD` on that path is empty). It was reverted in the minimality pass along with the other formatting-only churn.

app/[locale]/(static)/pricing/components/pricing-section.tsx
Comment (benjiwagner) [reply app/[locale]/(static)/pricing/components/pricing-section.tsx]: The redefinition is gone — `pricing-section.tsx` is no longer part of this diff. On the linter rule: I did not add one here. There are 13 pre-existing `variant="default"` sites elsewhere in the repo, so a rule (or a convention test, which is the idiom this repo already uses for this kind of guard) would fail until those are cleaned up too. That is a worthwhile separate change rather than something to bundle into this PR — happy to open it if you want.

prisma/schema.prisma
benjiwagner: Removed — there is no rollback text left in the schema. `git diff origin/main...HEAD -- prisma/schema.prisma` now shows only the Widget model changes: the `kind` discriminator, `timelineFilters`, and the three chart columns relaxed to nullable. The rollback procedure lives in the PR description instead, where it belongs.

core/base/base-custom-column-entity-modal.store.ts
benjiwagner: It is not lost, just de-duplicated. The line immediately above is `await this.entityStore.upsertItem(res.data)`, and `BaseDataViewStore.upsertItem` (core/base/base-data-view.store.ts:607-611) now calls `refreshActivityTimelinesForMany(this.entityType, [target.id])` itself. So the direct `activitiesStore.refreshFor(...)` call here would have fired the refresh twice.

The registry version is also strictly better than what it replaced: the old call refreshed one store, whereas the registry refreshes every mounted timeline that actually covers the entity (`coversEntity`), coalesced through a microtask so several saves in a tick produce one refresh. That is what makes the dashboard widget and an open contact timeline both update from a single save.

app/[locale]/(protected)/dashboard/components/chart-widget-card.tsx
Comment (benjiwagner) [reply app/[locale]/(protected)/dashboard/components/chart-widget-card.tsx]: Done in `4e6981cf`, and you were right that this file was the outlier — but the fix went the other way from moving them into the store, so worth explaining.

I surveyed how the codebase actually derives state before touching anything. The convention is consistent: derivations over store observables live either as registered MobX computeds (`BaseDataViewStore` registers 16 of them; `WidgetModalStore` already has `activeFiltersCount`, `customColumns`, `showDealFiltersTab`) or as plain inline consts in the component (`display-options.tsx:104`, `mass-update-popover.tsx:28`). `useMemo` shows up only when a non-observable React input enters the expression, or when a third-party library needs a stable reference (`data-table.tsx:164` for useReactTable). Under `observer`, re-render is already gated on observable change, so memoizing a store read mostly buys nothing.

By that rule all five memos here were the deviation, including inside this file: `inlineFilters` is by far the heaviest derivation and was never memoized, and this branch's own `activity-widget-card.tsx` filters a store array inline. So they are now inline consts.

Two of them could not have become store computeds anyway: `activeEntityFilters`/`activeDealFilters` derive from the per-widget `widget` prop while `widgetsStore` is a singleton, and a MobX `computed` takes no arguments. The repo's own escape hatch for that shape is a plain store method (`activities.store.ts` `coversEntity`), not a component memo.

The `subheader` memo was also a real bug, not just style: it called `intlStore.formatCurrency`/`formatNumber` inside the memo callback, so those observable reads happened outside the observer's tracking pass and the card stopped re-rendering when the company currency or the user's formatting locale changed. Computing inline restores that subscription. The preview built the byte-identical string, so the join is now one shared helper with unit tests, and a convention test pins that neither widget card reintroduces `useMemo`.

I left `WidgetsStore` alone deliberately — adding a `customColumnsByEntityType` index there would have duplicated the one `WidgetModalStore` already has. Happy to go further if you would rather see these as store members regardless.

app/api/v1/messaging/activities/search/route.ts
benjiwagner: Checked and changed in `c9cb0df7`. The common way is: the route does `const data = await request.json()` and hands it straight to the interactor, the interactor validates with `@Validate(...)`, and the `...ApiSchema` variants exist only for the OpenAPI spec and the MCP tool definitions. `app/api/v1/contacts/search/route.ts` is the exact template, and `GetQueryParamsApiSchema` appears nowhere at runtime.

One thing I did not do, because it would have quietly widened the public contract: simply deleting the parse. `ActivitiesApiParamsSchema` is `.strict()` and drops `p13nId`, so it is deliberately stricter than the interactive schema the shared interactor validates with — an API caller could otherwise pass unknown fields and `p13nId` and be accepted. So the schema moved onto a `GetActivitiesApiInteractor` rather than being removed. The route now matches its siblings, validation happens on the interactor, and the two rejection cases moved from the route test to a contract test next to the schema.

core/di.ts
benjiwagner: Fixed in `c9cb0df7`. Every other repository in the container is reached through a factory (`getContactRepo`, `getTaskRepo`, `getWidgetRepo`, all `() => new PrismaXRepo()`), and the activities repository was the only one being constructed inline — at five separate call sites, so each interactor was getting its own instance with no single place to change the wiring. There is now a `getActivitiesRepo` and all five sites use it.

components/data-view/filter-modal/edit-filters-modal.store.ts
benjiwagner: It scopes a shared singleton modal to the table that opened it. `editFiltersModalStore` lives on the root store, so `isOpen` alone is global: before this branch that was invisible because only one `FilterPopover` was ever mounted at a time, but the activity timeline adds a second one (the panel and the widget both render filters), and without the guard opening filters on one would pop the overlay open on both. The only consumer is `filter-popover.tsx:144`.

I did try to remove it and inline the comparison as `modalStore.tableStore === store`, which would have avoided adding an API at all. That does not compile: `tableStore` is `BaseDataViewStore<HasId> | undefined` while the popover is generic over `E`, so TypeScript rejects the comparison as having no overlap (TS2367). Taking `store: unknown` is what erases that variance. Happy to rename it to something more self-explanatory like `isOwnedBy` if the name is the sticking point.

components/data-view/filter-modal/inputs/use-filter-select-items.tsx
benjiwagner: Checked the neighbours in this same map and this branch follows them rather than departing: the `connectedAccountId` branch (`:188`) and the record-options branch (`:222`) added here use exactly the same `if (!result.ok) throw new Error(...)` shape, and the pre-existing `getTasksAction` branch above it is the same then-map-to-items form. The thrown error is what `form-autocomplete` turns into its `loadError` state, so it is the established channel for an option loader failing rather than a new one.

The one real difference is that this branch's loaders take a body (`scope`, `filters`, `connectedAccountIds`) instead of just `params`, which is why they are block-bodied while the older ones are one-line arrows. If what you are pointing at is that shape, tell me and I will pull the request-building into small named helpers so the map entries read like the rest. I would rather confirm than guess at which part reads as invented.

app/[locale]/(protected)/profile/components/connected-accounts.store.ts
Comment (benjiwagner) [reply app/[locale]/(protected)/profile/components/connected-accounts.store.ts]: Fair flag, and worth being precise about what is new here. `ensureLoaded` and the `loadPromise` de-duplication already existed on main; what this branch adds at this spot is the `loadError` observable, the `setItems` override that clears it, and the catch that sets it.

It is at least applied consistently rather than one-off: `activities.store.ts` carries the same `loadError`, and both feed shared state resolvers (`entity-timeline-state.ts`, `activity-widget-state.ts`) that decide between the error, empty and content states. `use-filter-select-items` exposes the same name for the same idea.

You are right that the underlying concept is missing from the base, though: `BaseDataViewStore` has no error state at all, which is why each store re-declares one. I tried lifting `loadError` into `BaseDataViewStore` (set on a failed `refresh`, cleared when one starts) so it would be one shared concept instead of three. It works, but it changes semantics for every data view store and it broke two existing tests in ways that showed the reset point is genuinely subtle: `setItems` clearing the flag races with the activities store, which sets it and then returns a result that flows through `setItems`.

I backed that out rather than land a shared-base change of that blast radius late in review. Happy to do it as its own PR where it can be reviewed on its own merits — say the word and I will open it.

app/[locale]/(protected)/dashboard/components/widget-subheader.ts
```diff
@@ -0,0 +1,5 @@
+export function widgetSubheader(count: number, formattedTotal: string, groupsLabel: string): string | null {
```
Comment (benjiwagner) [inline app/[locale]/(protected)/dashboard/components/widget-subheader.ts:1]: It looks like we can easily inline this if this is not used in mutliple files

features/messaging/activities/activities-store.context.tsx
```diff
@@ -0,0 +1,74 @@
+"use client";
```
benjiwagner: This looks odd.

app/api/v1/messaging/activities/search/route.ts
```diff
@@ -3,13 +3,13 @@ import type { NextRequest } from "next/server";
 import { NextResponse } from "next/server";
 import { z } from "zod";
 
-import { getGetActivitiesInteractor } from "@/core/di";
+import { getGetActivitiesApiInteractor } from "@/core/di";
```
benjiwagner, line 6: This looks odd.

components/data-view/filter-modal/inputs/filter-input-select.tsx
```diff
@@ -3,147 +3,209 @@
 import type { Filter } from "@/core/base/base-get.schema";
```
benjiwagner, line 3: This looks odd.

components/data-view/filter-modal/inputs/use-filter-select-items.tsx
```diff
@@ -1,6 +1,7 @@
 import type { GetResult } from "@/core/base/base-get.interactor";
```
benjiwagner, line 1: This looks odd.

components/data-view/filter-modal/edit-filters-modal.store.ts
benjiwagner: This looks odd.

core/stores/activity-timeline-registry.ts
```diff
@@ -0,0 +1,46 @@
+import type { EntityType } from "@/generated/prisma";
```
benjiwagner: This looks odd.

core/stores/root.store.ts
```diff
@@ -201,10 +201,6 @@ export class RootStore {
     return (this._threadComposeStore ??= new ThreadComposeStore(this));
   }
 
-  get activitiesStore() {
```
benjiwagner, line 204: This looks odd.

ee/messaging/activities/get-activity-record-options.interactor.ts
```diff
@@ -0,0 +1,44 @@
...
+export type ActivityRecordOptionsData = Data<typeof Schema>;
+export type ActivityRecordOption = Data<typeof ActivityRecordRefSchema>;
+
+export abstract class ActivityRecordOptionsRepo {
+  abstract listRecordOptions(data: ActivityRecordOptionsData): Promise<ActivityRecordOption[]>;
+}
+
+@AllowInDemoMode
+@TenantInteractor()
+export class GetActivityRecordOptionsInteractor extends AuthenticatedInteractor<
```
benjiwagner, line 26: This looks odd.

app/[locale]/(static)/pricing/components/pricing-section.tsx
Comment (benjiwagner) [reply app/[locale]/(static)/pricing/components/pricing-section.tsx]: The guard now has a home: CUS-215 (Reintroduce the repo-wide convention guards on their own) already lists the TabsList `variant="default"` sweep verbatim, next to the other repo-wide rules that were stripped from this branch. Landing it there means the commit that introduces the rule also cleans up the 13 pre-existing sites it would flag. Nothing of this remains in #42, so resolving.

components/data-view/filter-modal/edit-filters-modal.store.ts
benjiwagner: Gone in 58884054. The real blocker was FilterPopover's generic: `Props<E extends HasId>` where `E` is phantom, it only appears in the `store` prop and the body never uses it. After dropping the generic and typing the prop `BaseDataViewStore<any>`, which is exactly how this modal store already erases the same variance in `openFor` (edit-filters-modal.store.ts:119 on main), the popover compares identity directly: `open={modalStore.isOpen && modalStore.tableStore === store}`. No new API on the store.

features/messaging/activities/activities-store.context.tsx
benjiwagner: Agreed, and it is gone in 58884054. Two things were genuinely odd: `ActivitiesStoreContext`/`useActivitiesStore` were dead code (every descendant already receives the store as a prop, the house pattern across data-view and entity-detail), and the provider fed the whole store instance into a context typed `{scope, filters}` via structural typing, which has no precedent. The file is replaced by `activity-query-context.tsx` (plain value context, same module shape as `topbar-actions-context.tsx`; the panels now pass `scope`/`filters` explicitly) and `use-owned-activities-store.ts` (the construction hook, same file idiom as `use-navigation-guard.ts`). Per-mount store construction itself is the established mechanism of `RootStoreProvider` (`useMemo(() => new RootStore(...))` behind a context) and `AppForm` (store instance carried by context), and it is required here because the dashboard mounts several timelines at once.

app/api/v1/messaging/activities/search/route.ts
benjiwagner: This one is correct as is, and it moves the route toward the convention rather than away. All 13 `get*ApiInteractor` factories on main are the same interactor class constructed with mode "api" (never a subclass), every sibling public search route uses its Api factory, and this handler is character-for-character the contacts/organizations template. Main's activities route was the only API route wired to the interactive interactor, while the MCP tool already used `getGetActivitiesApiInteractor` (messaging.mcp-tools.ts:212 on main), so REST and MCP served the same entity through different modes. Mode "api" skips p13n read/write and runs the query-params precheck, which turns silently dropped invalid filters into a 400.

components/data-view/filter-modal/inputs/filter-input-select.tsx
benjiwagner: Rebuilt in 58884054 on top of main's file. It is now main's component plus exactly four CUS-68 grafts: the 50-value cap (`nextFilterSelection`, disabled options past the limit, the limit footer), `onValueChange` (the channel filter clears the stale thread filter), and the resolved-or-masked remove label. Everything else that had accumulated is deleted: the `buttonVariants` overlay trigger, the focusable remove buttons plus the rAF focus helper, the CheckIcon/loop/autoFocus polish, the aria props no caller passed, and the two branch-only i18n keys. Unresolved chips now mask through main's existing `Common.inputs.unavailableSelection` path (which never leaks raw ids), `filter-select-label.ts` is deleted, and `Common.filters.unavailableValue` is back to main's copy. As part of the same cleanup `form-autocomplete.tsx` is reverted to main wholesale: its consumer set is byte-identical to main, no CUS-68 file uses it, and the merged loading-contract conventions pass against main's version.

components/data-view/filter-modal/inputs/use-filter-select-items.tsx
benjiwagner: Trimmed in 58884054. Dropped: the `connectedAccountIds` argument and its `selectedChannelIds` helper. The server derives the same narrowing from `filters` (interpretFilters -> `query.connectedAccountIdsIn`, which takes precedence in `listThreads`), so the client was intersecting a set with itself. Tightened: the record-option resolver guards non-uuid and over-cap values instead of erroring into the retry banner, restores avatar parity for resolved contact chips, and the loader memo is keyed on the scope string as on main rather than the context object. What stays is deliberate: the separate `getSelectedItems` resolver exists because resolving selected ids through `getItems` with an id filter is explicitly forbidden by tests/conventions/technical-id-loading-contract.test.ts, and main's page-1-only resolution cannot round-trip a saved widget's chips; it reuses the repo's `resolveRecordLabels`, the same ACL-scoped resolver that labels timeline rows.

core/stores/activity-timeline-registry.ts
benjiwagner: Simplified in 58884054. The two parts with no precedent are gone: the microtask flush queue (`queueMicrotask` has zero uses on main, and request coalescing already lives in the store's shared in-flight map keyed by normalized query), and `register()` returning a disposer. It is now a plain Set with paired `register`/`unregister`, mirroring the only two registries main has: `NavigationGuardController` (navigation-guard.controller.ts:21-27) and `RootStore.modalStores` (root.store.ts:72, closeAllModals fan-out), both eagerly instantiated on RootStore exactly like this one. Also renamed to `activity-timeline.registry.ts` to match the folder's dot-role naming (`navigation-guard.controller.ts`, `root.store.ts`).

core/stores/root.store.ts
benjiwagner: This deletion is correct as is. The getter could not survive the feature: the store's constructor now takes options (scope, page size, initial hydrate), and a singleton cannot represent several concurrently mounted timelines (the entity panel plus any number of dashboard widgets). All four consumers main had are migrated and `git grep activitiesStore` on the branch returns zero hits: use-filter-select-items reads the new query context, activities-panel constructs its own store via `useOwnedActivitiesStore`, the custom-column modal's `refreshFor` call is covered transitively by `upsertItem` fanning out through `rootStore.activityTimelines`, and the contact-detail test stub became an `upsertItem` assertion. RootStore's own split stays intact: registries eager (`navigationGuard`, now `activityTimelines`), stores lazy.

ee/messaging/activities/get-activity-record-options.interactor.ts
benjiwagner: Good instinct, two real issues were hiding near this line, both fixed in 58884054: the local `RecordSelectionSchema` duplicated the exported `ActivityScopeRecordsSchema` (with a drift-prone twin constant, ACTIVITY_FILTER_VALUE_MAX vs ACTIVITY_SCOPE_MAX_IDS_PER_TYPE, both coincidentally 50), and `ActivityRecordOption` re-named the existing `ActivityRecordRefDto`. The input is now `records: ActivityScopeRecordGroupsSchema.min(1)` and the DTO type is imported, matching how the sibling get-activity-thread-options.interactor.ts imports rather than redefines. The declaration itself is the house shape: `AuthenticatedInteractor` (116 direct subclasses on main), the decorator stack in the same order as all 47 `@AllowInDemoMode` users, the colocated abstract repo (90 files on main, and required by di-boundaries.test.ts since interactors cannot import prisma repositories), and the multi-line wrap is just Prettier at 120 columns, identical to the sibling's.

PR #49, opened 2026-08-10
fix(seo): enforce landing-page shipping contracts

eljulsi:
## Summary

Reworks [CUS-196](https://linear.app/customermates/issue/CUS-196/close-the-locale-parity-orphan-page-and-hero-coverage-holes) on current `main` (`520cd53c`) and turns the landing-page publishing conventions into executable pre-merge contracts.

- Adds bidirectional hero coverage and 1920×1080 PNG-header validation for 168 landing slugs / 672 localized theme assets.
- Parses MDX with the real MDX AST and rejects root-relative links whose published route does not exist; repairs the 11 existing dead-link instances without changing surrounding CUS-164 copy.
- Replaces four footer filter Sets with one ordered preference manifest plus deterministic selection from live content.
- Paginates `/blog`, `/compare`, `/features/all`, and `/for` at 24 cards while preserving locale fallback and equivalent EN/DE page membership.
- Makes pagination URLs strict and indexable: malformed/repeated/out-of-range pages are wire-level 404s, `?page=1` is a wire-level 308 to the clean URL, and valid later pages have self-canonicals, reciprocal hreflang, and sitemap entries.
- Uses a square-root bucket pager so the 2,256-page backlog remains within four clicks without rendering 94 page-number links on every list page.
- Removes the root-host `Crawl-delay: 10` while preserving the preview/app-subdomain `Disallow: /` branch.

The old PR’s i18n-parity hunk is intentionally gone. Current `main` already has the stronger `CONTENT_LOCALES` / `DEFAULT_LOCALE` collection-wide parity implementation and explicitly disables Fumadocs language fallback. This branch keeps that authority rather than restoring stale two-locale logic.

## Context

The current catalog has 168 landing pages per content locale: 71 blog posts, 31 comparison pages, 24 feature pages, and 42 industry pages. The v5 backlog raises that toward 2,256 pages. Before this change, several failures were either silent or visible only after deployment:

1. A missing hero filename produced a broken image request because the `<slug>.png` relationship lived only in templates.
2. MDX prose was not typechecked as a route graph; 11 hrefs already targeted four absent URLs.
3. Footer entries were four independent literal Sets; removal silently shortened a column.
4. Every hub rendered its entire collection, so HTML weight grew linearly.
5. The previous pagination draft accepted values such as `2junk`, turned invalid/out-of-range URLs into duplicate page-one responses, canonicalized every page to page one, and rendered every numbered page link.
6. Next App Router streaming caused page-level `notFound()` and `permanentRedirect()` to remain HTTP 200 on the wire. The final design validates in `proxy.ts` before rendering and retains page-level checks as defense in depth. This follows Next’s documented streamed-response behavior.

### Final pagination contract

- Missing `page` means page 1.
- Exactly one canonical positive decimal is accepted. Empty, repeated, zero, signed, zero-padded, decimal, suffixed, unsafe, and out-of-range values return HTTP 404 with `X-Robots-Tag: noindex`.
- Exactly `page=1` returns HTTP 308 to the clean localized hub URL while preserving unrelated query parameters.
- Page 2+ returns HTTP 200, self-canonicalizes to that query URL, and emits EN/DE/x-default alternates for the same page.
- Page 2+ URLs are present in the sitemap. The sitemap grows from 412 to 420 URLs: eight new localized hub-page entries.
- One default-locale ordering chooses each 24-slug slice, then localized records are mapped into that slice. EN page N and DE page N therefore represent the same entities even when display sorting differs by language.
- The pager renders real server-visible anchors, localized Previous/Next labels, `aria-current`, visible focus rings, and 44px targets.

### Why the square-root pager

A three-click guarantee from home requires page 1 to link directly to every list page. At 2,256 details / 24 cards, that means 94 numbered anchors on the hub. The final model sets bucket size to `ceil(sqrt(pageCount))` and renders every bucket start plus all pages in the current bucket, with explicit Previous/Next links.

`home → hub page 1 → bucket start → target list page → detail`

This makes list-page diameter at most two and detail depth at most four. At the 94-page backlog, the largest pager measured 20 hrefs / 6,647 bytes instead of 94 numbered links. The current three-page pager is at most four hrefs / 1,585 bytes.

## Validation

All work ran in the isolated CUS-196 worktree against a disposable PostgreSQL 16 container bound only to `127.0.0.1:55432`, seeded with deterministic synthetic fixtures.

### Final repository gates

```text
yarn test (RUN_DATABASE_TESTS=true)  240 files / 1,880 tests passed
yarn lint                            passed, zero warnings
yarn typecheck                       passed
yarn i18n:audit                      passed (advisory findings only)
yarn build                           passed; optimized Next.js production build
git diff --check                     passed
```

### Rendered local preview

The final port-40196 local preview was exercised with both a browser and an HTTP crawler. Port 4000 was already owned by an unrelated CUS-68 worktree and was deliberately not stopped.

- Crawled all 16 EN/DE hub responses and all 336 localized detail routes.
- Every hub response rendered at most 24 cards; the largest observed response was about 463 KB.
- Maximum observed anchor count was 109, including global navigation and footer links.
- Every detail route was reachable from its localized home within four clicks.
- EN/DE slug membership matched for every corresponding hub page.
- Canonical and EN/DE/x-default alternate tags matched each valid page URL.
- `?page=1` returned 308; malformed, repeated, and out-of-range queries returned 404; valid pages returned 200.
- `/fr/blog?page=2` redirected to `/en/blog?page=2`, preserving the query under the app-locale/content-locale split.
- `/sitemap.xml` contained 420 URLs, including all eight localized page-2/page-3 hub URLs.
- `/robots.txt` emitted `Allow: /`, no crawl delay, and retained the separate subdomain exclusion behavior.
- Browser review covered `/en/blog`, navigation to `/en/blog?page=2`, and `/de/compare?page=2`; content rendered, pagination controls were visible and 44px high, no Next error overlay appeared, and no runtime error was logged. One pre-existing logo aspect-ratio warning remains unrelated to this diff.

### Mutation proofs

Each induced failure was restored before the final run:

| Mutation | Gate response |
| --- | --- |
| Remove `content/for-pages/de/healthcare.mdx` | Locale parity reports the exact missing file |
| Remove `public/images/dark/de/healthcare.png` | Hero coverage reports the exact missing asset |
| Add an orphan PNG | Hero coverage reports the asset with no landing page |
| Change a repaired href back to `/features/self-hosting` | AST link gate reports the exact file, line, and target |
| Synthesize a malformed/wrong-size PNG header | Header validator rejects it |
| Synthesize feature slug `all` | Reachability gate rejects collision with static `/features/all` |

### Content repairs

All repairs are href-only; no page was added or removed and all surrounding PR #51 / CUS-164 copy is preserved.

| Old target | New target | Instances |
| --- | --- | ---: |
| `/features/self-hosting` | `/features/self-hosted` | 2 |
| `/blog/crm-introduction` | `/blog/what-is-crm` | 5 |
| `/blog/crm-implementation-guide` | `/blog/crm-implementation` | 2 |
| `/compare/salesforce-pricing` | `/blog/salesforce-pricing` | 2 |

The Salesforce destination is intentionally the live pricing article because the unchanged sentence promises a full per-edition pricing breakdown. The generic Salesforce-alternative page is not semantically equivalent.

### New convention coverage

- Hero gate: 168 unique slugs × `{light,dark}` × `{en,de}` = 672 assets, both directions, PNG signature/IHDR, 1920×1080, and flat-namespace collision protection.
- Link gate: Markdown links, reference links, direct single/double-quoted JSX hrefs, and static string-literal JSX expressions, including multiline JSX. Query, fragment, trailing slash, and known locale prefixes are normalized.
- Reachability gate: renders the real pagination component, extracts its actual hrefs, builds BFS over card slices, verifies the current content graph and a synthetic 2,256-detail / 94-list-page case, and checks all four production routes remain wired to the shared helper.
- Proxy gate: keeps its lightweight page-count registry locked to the content-derived counts so content growth cannot make pre-render validation stale.

### Impact

- No schema migration, environment-variable change, database write, external publication, or page-file creation/deletion.
- Four hubs intentionally read `searchParams`, making their pagination request-aware. They were already dynamic under the localized application shell; this is an explicit URL/metadata contract, not accidental client state.
- Current 24-link footer output is unchanged. Preferred entries are editorial preferences, not a ranking engine; if one disappears, selection deterministically tops up from live content and the convention test guarantees six valid, locale-aligned links.
- Existing five app locales and two content locales remain intact. `/fr`, `/it`, and `/es` content requests continue through `contentLocaleOrDefault` rather than trying to load nonexistent content trees.
- The public three-item navbar remains deliberately curated.
- Removing crawl delay can increase Bing’s request rate. That is accepted because the public surface is CDN-fronted static marketing content. Google ignores the directive; at the new 420-URL sitemap, the old 10-second delay would impose 4,200 seconds / 70 minutes per honoring crawler pass, and roughly 12.8 hours at backlog scale.
- CUS-202-owned `footer-content.tsx` and `tests/conventions/locale-consumer-audit.test.ts` are untouched.

### Rollback

The change can be reverted as one commit (`932e1372`) with no data reconciliation. A partial rollback must keep coupled contracts together:

- Revert hub components, metadata, sitemap entries, proxy validation, and reachability tests together.
- Revert the AST link test only together with any link repair that would otherwise become invalid.
- Revert footer selection and its selection assertions together.
- Reverting `app/robots.ts` alone restores the former crawl delay.

### Follow-up

CUS-196’s Knowledge Base companion must be refreshed after this product PR merges so its source pin can cite the actual merged product commit. Current KB `main` moved during this rework and now needs the union of the shipping gates plus the 420-URL sitemap arithmetic. That post-merge documentation update is required before CUS-196 closes, but it is not a blocker to reviewing this product PR.

Independent read-only product review found no unresolved high- or medium-severity code findings after the runtime-status, href-only content, and static-route-collision fixes.

### DIFF
--- a/__tests__/proxy-locales.test.ts
+++ b/__tests__/proxy-locales.test.ts
@@ -191,6 +191,23 @@ describe("locale routing configuration", () => {
 });
 
 describe("proxy locale routing", () => {
+  it("normalizes page one and rejects invalid hub pagination before streaming", async () => {
+    const pageOne = await call("/en/blog?page=1&utm_source=proof");
+    expect(pageOne.status).toBe(308);
+    expect(pageOne.location).toBe("http://localhost:4000/en/blog?utm_source=proof");
+
+    for (const query of ["page=0", "page=01", "page=2junk", "page=2&page=3", "page=4"]) {
+      const invalid = await call(`/en/blog?${query}`);
+      expect(invalid.status, query).toBe(404);
+      expect(invalid.location, query).toBeNull();
+      expect(invalid.response.headers.get("x-robots-tag"), query).toBe("noindex");
+    }
+
+    const lastPage = await call("/en/blog?page=3");
+    expect(lastPage.status).toBe(200);
+    expect(lastPage.location).toBeNull();
+  });
+
   it("never emits a permanently cached redirect", async () => {
     for (const path of PATHS) {
       const { status } = await call(path);
--- a/app/[locale]/(static)/blog/page.tsx
+++ b/app/[locale]/(static)/blog/page.tsx
@@ -1,54 +1,94 @@
 import type { Metadata } from "next";
 
-import { getLocale } from "next-intl/server";
-import { notFound } from "next/navigation";
+import { getLocale, getTranslations } from "next-intl/server";
+import { notFound, permanentRedirect } from "next/navigation";
 
 import { BlogPostCard } from "./blog-post-card";
 
 import { Footer } from "@/app/components/footer";
+import { HubPagination } from "@/components/marketing/hub-pagination";
 import { PostGridShell } from "@/components/marketing/post-grid-shell";
 import { generateMetadataFromMeta } from "@/core/fumadocs/metadata";
 import { blogPostsSource, blogSource } from "@/core/fumadocs/source";
-import { contentLocaleOrDefault } from "@/i18n/locale-registry";
+import {
+  HUB_PAGE_PARAM,
+  hubPageCount,
+  hubPageHref,
+  paginateLocalizedHubPages,
+  resolveHubPage,
+} from "@/core/seo/hub-pagination";
+import { DEFAULT_LOCALE, buildLocalePath, contentLocaleOrDefault, formattingTagFor } from "@/i18n/locale-registry";
 
-export async function generateMetadata({ params }: { params: Promise<{ locale: string }> }): Promise<Metadata> {
+type Props = {
+  params: Promise<{ locale: string }>;
+  searchParams: Promise<Record<string, string | string[] | undefined>>;
+};
+
+export async function generateMetadata({ params, searchParams }: Props): Promise<Metadata> {
   const { locale } = await params;
-  return generateMetadataFromMeta({ locale, route: "/blog" });
+  const resolution = resolveHubPage(
+    (await searchParams)[HUB_PAGE_PARAM],
+    hubPageCount(blogPostsSource.getPages(DEFAULT_LOCALE).length),
+  );
+
+  if (resolution.kind === "not-found") notFound();
+
+  return generateMetadataFromMeta({
+    canonicalPath: hubPageHref("/blog", resolution.page),
+    locale,
+    route: "/blog",
+  });
 }
 
-export default async function BlogPage() {
+export default async function BlogPage({ searchParams }: Props) {
   const locale = contentLocaleOrDefault(await getLocale());
   const page = blogSource.getPage(["blog"], locale);
-  const posts = blogPostsSource.getPages(locale);
 
   if (!page) notFound();
 
-  const sortedPosts = [...posts].sort((a, b) => {
-    const dateA = new Date(a.data.blogPost.date).getTime();
-    const dateB = new Date(b.data.blogPost.date).getTime();
-    return dateB - dateA;
-  });
+  const referencePosts = blogPostsSource.getPages(DEFAULT_LOCALE);
+  const resolution = resolveHubPage((await searchParams)[HUB_PAGE_PARAM], hubPageCount(referencePosts.length));
+
+  if (resolution.kind === "not-found") notFound();
+  if (resolution.kind === "redirect-page-one") permanentRedirect(buildLocalePath(locale, "/blog"));
+
+  const referenceCollator = new Intl.Collator(formattingTagFor(DEFAULT_LOCALE));
+  const paginated = paginateLocalizedHubPages(
+    referencePosts,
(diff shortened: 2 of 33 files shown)


benjiwagner:
Closing this PR unmerged during post-release cleanup. This work is not obsolete: the unique CUS-196 shipping-contract implementation remains preserved on branch `fix/page-shipping-gates` at `932e1372`. The overlapping content-link repairs have since landed through #63/#50 and current main must remain authoritative if this work is revived. CUS-196 will return to Backlog with the remaining implementation scope and KB/product mismatch recorded.

PR #62, opened 2026-08-12

benjiwagner:
## Summary

Ships the complete customer-facing hosted AI Assistant for CUS-125 as one feature. **Supersedes #37**, which was opened from a branch whose history had to be rewritten; this branch carries the same work plus everything since, on a clean history that merges `main` rather than replaying onto it.

- **Chat.** Customermates-branded compact and expanded chat, streaming Markdown, stop, retry, one editable queued follow-up, jump-to-latest, responsive layout, keyboard access and screen-reader semantics.
- **Safe activities.** Expandable semantic activity summaries that never expose UUIDs, raw arguments or results, secrets, hidden reasoning, page-context markup, provider identifiers or costs. UI-command outcomes stay truthful, so a failed navigation or tour is reported as a failure.
- **History.** Multiple conversations with server-side title and sanitized-message search, 25-conversation and 50-message cursors, upward transcript loading, new/switch/archive/undo/restore, sequence-based unread support replies, and permanent deletion from the archive.
- **Page awareness.** Exactly three deterministic suggestions per route from one centralized catalog keyed on route, data state, permissions, locale and custom terminology; guided tours; and the panel opening itself when the current page is genuinely empty.
- **Workspace setup.** Conversational discovery, review, apply and cleanup covering custom columns, minimal linked records and dashboard widgets, with provenance so cleanup can be reversed safely.
- **Credits.** Tiered monthly hosted-AI credits with atomic reservation before provider access, boundary-safe settlement, and a full-width monthly meter. This replaces the previous fixed `$2/week` cap entirely.

**Five languages, for real.** The Assistant is fully localized in de, en, es, fr and it. All five catalogs hold 1,854 keys at exact parity with zero ICU argument or structure mismatches.

## Context

**What changed since #37.** Five rounds of review-driven work, each its own commit:

1. **Localisation completed.** 609 user-visible strings still branched on an English/German prefix check, so Spanish, French and Italian users met English suggestion chips, tour notes and error copy inside an otherwise translated product. That copy moved into the catalogs (144 new keys × 3 locales) and the modules now take a translator instead of a locale string. `agent-tours.ts` shrank from 215 lines to 76; the rebuilt tours are byte-identical to the previous implementation for all 7 tours in both original languages.

   This also uncovered a functional gap: `selectAgentToolNames` matched only English and German keywords, so an es/fr/it user asking to create a contact never had `create_contacts` offered to the model and the Assistant simply could not perform the action. The hint tables now cover all five languages, with a test asserting the write tools are reachable in each.

2. **The audit blind spot is closed.** `tests/conventions/locale-consumer-audit.test.ts` passed throughout because the offending strings used shapes it did not match. It now rejects language-prefix branching, two-locale narrowing ternaries and sibling locale keys carrying copy. Run against the pre-fix sources the new detectors flag 10 branch sites and 26 copy tables; against this branch, zero.

3. **Tenancy hardening and schema integrity.** The turn now asserts that the session it executes under is the one it was admitted for, before any tool can fire. `AgentUsageEvent` queries carry `companyId`, the UI-command upsert key is scoped by tenant, and a forward migration adds the seven missing `Company` foreign keys, including on the credit ledger which previously had no relation at all.

4. **Decorator alignment.** Fourteen interactors adopt `@Write` and `@ValidateOutput` to match the 49 and 81 files that already use them, **without moving a single transaction boundary**: every file that had a transaction still has one, and the six non-transactional writes pass `tx: false` because `runInTransaction` takes a company-wide advisory lock that would otherwise serialise archiving and read-marking. Three interactors deliberately take no output schema, most importantly `send-agent-message`, whose result is spread into `runAgentLane` as a 13-field context that Zod would silently strip.

5. **Duplication and formatting.** `assertValidDate` was defined twice by this branch and now lives in `core/utils/date.ts`; the `request_support` input schema is shared rather than copied. Three panel dates formatted from the UI language instead of `user.formattingLocale`; all three now go through `IntlStore`, leaving no hand-built `Intl.DateTimeFormat` anywhere in the protected app.

**Surface boundary.** The Assistant owns its own surface and nothing else. It adds the launcher and the panel and decides when to open itself; it does not change how any existing page looks. The one difference under `components/data-view/` is `data-view-container.tsx`, which carries the shared back-to-top control from CUS-180 because PR #46 was stacked on this branch and merged into it.

**Commercial contract.** Starter 200, Pro 500, Business 1,200 credits per active user per month. Enterprise uses a finite internally configured allowance and fails closed when it is missing. Trials get 500 credits per active trial user. Credits reset on the billing-anniversary day including annual plans, new paid seats get whole-credit proration, upgrades raise the ceiling immediately and downgrades clamp it immediately. No rollover, top-up, redistribution or automatic overage in v1. One credit is one started cent of internally metered processing: a turn is charged `max(1, ceil(total provider cost / $0.01))`. Dollars, tokens, model identifiers and provider pricing stay server-only. The hosted Assistant is cloud-only; self-hosted keeps MCP access.

**Persistence.** Two additive forward migrations. `20260812090000_agent_assistant_turn_ledger` adds ordered `AgentMessage.sequence` and sanitized `searchText`, conversation archive/selection/read state, `AgentTurnRequest` for replay durability, the reserved/settled/retained/released credit ledger with plan snapshots, `AgentWorkspaceSetup` and its resources, the subscription credit anchor, the user seat-activation timestamp, and a nullable `SupportTicket.agentConversationId` with `onDelete: SetNull`. It drops the legacy `@@unique([companyId, userId])` that allowed only one conversation per user. `20260812130000_agent_tenant_foreign_keys` adds the tenant foreign keys and the scoped unique index. Legacy rows migrate deterministically: sequence is renumbered by `(createdAt, id)` rather than heap order, and `searchText` fails closed, blanking any legacy row still carrying a private marker rather than making it searchable.

**Merge, not rebase.** `main` moved 9 commits since the branch point. With 18 commits on this side, replaying them risks silently dropping code in auto-merged regions, so `main` was merged once and verified. Four conflicts were resolved: the inbox list and data view take `main`'s page-state rewrite verbatim with the back-to-top wrapper re-applied around it, `globals.css` keeps both animation sets, and the overlay allowlist drops `app/[locale]/loading.tsx`, which `main` rewrote onto the shared page-state component so it no longer holds a raw fixed surface.

## Validation

Local gates at `8828f3d9`, after the merge:

- `yarn typecheck`: passed.
- `yarn lint --max-warnings=0`: passed.
- `RUN_DATABASE_TESTS=true yarn test` against local PostgreSQL: **252 files and 2,197 tests, all passing.**
- `yarn build`: passed.

Merge integrity, verified rather than assumed:

- Catalogs went 1,850 to 1,854 keys with exact parity across all five locales.
- Every marker of this branch survived: 26 tour-step keys, the suggestion and runner catalogs in es/fr/it, 10 `@Write` interactors, 6 `tx: false` sites, `IntlStore.formatDayMonth`, `core/utils/date.ts`, `agent-translator.ts`, all 43 agent-chat feature files and both migrations.
- `main`'s own #59 dash-punctuation fix is present and intact.

Localisation evidence:

- 1,854 keys × 5 locales at parity, 0 ICU syntax errors, 0 argument mismatches, 0 em dashes.
- `i18n-key-resolution` required registering all five new dynamic key sites, so every one of the 144 new keys is provably reachable from a real call site and no key is orphaned.
- Driven in the browser at `/es` and `/fr`: panel chrome, empty-state intro and suggestion chips all render in-language.

Decorator evidence:

- Every converted file previously stacked `@Validate` above `@Transaction`, which is exactly how `@Write` composes them, so the conversions are verbatim.
- Transaction membership before and after is identical, checked file by file.

Migration evidence, in `prisma/__tests__/agent-assistant-migration.database.test.ts` against disposable databases: legacy fixtures upgrade cleanly, sequence backfill is deterministic under tied timestamps inserted in opposite physical order, `searchText` strips a UUID and an `sk-` shaped token while keeping surrounding words searchable, a legacy-upgraded database converges on exactly the same columns, indexes and check constraints as one built from scratch, and the ledger check constraints reject a charge above its reservation, a double settlement, a released row carrying a charge and a negative amount.

Credit evidence: `agent-credit-ledger.database.test.ts` proves atomic reservation against a real database, admitting exactly three of five concurrent reservations when three credits remain. Mocked repositories cannot express this.

**Known limits, stated plainly.** A live guided tour could not be exercised end to end locally: demo mode blocks `SendAgentMessage` as a non-`@AllowInDemoMode` write, and self-hosted reports `blockedReason: self_hosted`. Neither is caused by this branch. Tour notes are verified at module level in all five locales through the same translator that produced the verified French chrome. Workspace terminology overrides still cover only en and de and degrade to no substitution elsewhere, which is a pre-existing limit of that feature.

## Impact and rollback

Disable `AGENT_CHAT_ENABLED` to turn the Assistant off; the launcher and panel disappear and no agent route is reachable. Persistence is additive and both migrations are forward-only, so disabling the flag leaves the schema in place and loses nothing.

The credit model replaces the previous fixed `$2/week` cap, so that is the one behavioural change visible to existing users and it is intended. Reverting the merge restores `main`'s behaviour; the migrations would remain applied but unused, since every new table is agent-scoped and every added column is nullable or defaulted.

Outside the Assistant, the only shared-surface changes are the back-to-top control from CUS-180 in `data-view-container.tsx`, the `deactivateUser` to `deactivateUserOrThrow` renames forced by `repo-naming.test.ts`, and the `content/*.mdx` edits forced by `hosted-ai-pricing-contract.test.ts`.


benjiwagner:
Superseded by #82, which carries an identical tree on a clean history.

This branch's history contained a merge commit with a non-conventional subject that `repository-policy` rejected, and correcting it in place would have meant rewriting published history. #82 is the same work on a fresh branch: same tree, all commits conventional, policy green.

Verified before closing: the tip tree of this branch (`4c2b97bb`) is byte-identical to commit `38146144` on #82, so nothing here is unreachable from there.

PR #70, opened 2026-08-17
feat(deals): surface a stage-weighted pipeline value

benjiwagner:
## Summary

Adds a probability-weighted pipeline value. Each option of a designated deal `singleSelect` column carries a win probability, and `Deal.weightedValue` stores `totalValue * weight / 100`. The figure appears on the deals list, the deal detail, kanban column headers, and a new dashboard metric, and is configured in company settings.

Closes CUS-233.

## Context

Deals carry a value but no likelihood, so the only pipeline figure the product could produce was the raw sum of deal values, which counts a first-contact deal and a final-negotiation deal identically.

Three constraints shaped the design, all verified against `origin/main`:

- There is no stage field. Per `content/docs/en/concepts.mdx:18` a pipeline is a user-defined `singleSelect` custom column, which can be renamed, deleted, or duplicated.
- Nothing marked which column was the pipeline. Kanban grouping lives in `P13n.groupingColumnId`, which is per user, not per company.
- `Deal.totalValue` is already a stored, indexed rollup maintained by `recalculateTotals`.

So storage splits by the kind of fact. Which column drives the forecast is a company-wide singleton and lives on `Company.dealWeightingColumnId` as a foreign key with `ON DELETE SET NULL`, so deleting the column clears the pointer in the database rather than in application code. What each stage is worth is per-option data and lives on the option inside the column's `options` JSON, so it survives renames and reorders and is deleted with the option. Storing the percentages on the company keyed by option id was rejected because it orphans on option deletion.

`weightedValue` is stored rather than computed on read, because a paginated list cannot be sorted server-side by a value the database does not hold, and the widget aggregate would otherwise load every deal with its service rows to render one tile. It is nullable: null means weighting is not configured, which is a different fact from a measured zero, and writing `0` would show honest "not set up" workspaces a forecast of nothing.

Percentages are editable in two places backed by one source of truth: company settings, which also owns designation, and the custom column modal, where stages are already named and ordered. The modal editor is gated on `company:update` because editing a deal custom column only requires `deals:update`, so without the gate a colleague could change the company-wide forecast from a kanban column header.

The default deal column created at registration becomes seven weighted stages.

### Recalculation has one rule

`totalValue` and `weightedValue` derive from four inputs: service amount and quantity, the deal's stage value, the stage column's option weights, and the company pointer. Every write path that touches those recalculates, and after this branch every trigger lives in **the repository performing the write**. No interactor triggers recalculation.

The company settings path used to be the exception, calling `recalculateDealWeightedValues` from the interactor. The cause was a bypass: the company repository wrote stage weights straight into `CustomColumn.options` with its own Prisma call, so the repository that owns custom columns never saw the write and its existing trigger could not fire. Stage weights now go through `PrismaCustomColumnRepo.setOptionWeights`, and the compensating call was deleted rather than relocated. Repeating a recalculation is cheap because `recalculateTotals` skips the update when stored values already match.

### Group sums became an entity-declared aggregate

Kanban column totals were first bolted onto `BaseGetInteractor` as a special case: the base branched on `EntityType.deal`, probed the repo with a `typeof` check, and read Deal column names from shared schema, which all fourteen interactors inherited. The base already aggregated per group through `getCount`, so field sums are now the same kind of thing, parameterised. An entity declares which numeric columns its groups summarise, the sums travel as a plain field-keyed record so the client store derives the fields from the response, and `sumNumericFields` moved onto the port, where it belongs because `BaseRepository` already gives every repository that method.

That made filtered totals cheap to add, so `valueSums` now returns the totals of an entity's declared numeric columns across the whole filtered query, not just the page. `list_records` returns them as `sums` and says so, because an agent could previously answer a pipeline question confidently from one page of ten records. Numeric columns are summable; single-select and other custom fields are filter or group dimensions.

### Also in this branch

- **A single select can have no default option.** The Default control was a radio input, whose `onChange` never fires when an already checked radio is clicked, so the store's toggle could only set a default and never clear one. Every backend consumer already guarded with `if (defaultOption)`, so no default was representable everywhere except in the UI. It is now a toggle button carrying `aria-pressed`, and deleting the default no longer force promotes another option.
- **Deals with no stage are disclosed rather than dropped.** They were excluded from the settings pipeline figure entirely, understating it. Total pipeline now counts them and a third row reports the value carrying no weight. Unknown is not treated as 0%.
- **The demo seed configures the forecast.** Seeded Status options carry win probabilities, the company points at that column, and seeded deals store their weighted value. The demo dashboard is deliberately unchanged. Scoped to the synthetic fixtures; the removal of general demo seeding in `9ec82bdd` stands.
- **Audit bodies render structured values as chips.** Custom column options previously read `Color: warning · Index: 0 · Label: Open · Value: 17000000-… · Weight: 30`, per option and per side of the diff. Options, stage weights and record naming now render as the option chips themselves, and terminology presets resolve through the localised catalogue, so a renamed workspace reads `Deals → Opportunities` rather than `Deals → opportunity`. The audit body still never resolves the workspace's current terminology, which the boundary gate enforces for that file.
- Smaller fixes: relationship diagram connectors meeting their cards at any width, even data model spacing on small screens, `TruncatedText` honouring caller alignment, and the services row widths.

## Validation

Local gates on the merged tree: `tsc --noEmit`, `yarn lint` (`--max-warnings 0`), and the full suite at **2675 passed, 1 skipped, 317 files**, run with `RUN_DATABASE_TESTS=true` against a throwaway Postgres 16 so the real-database tests executed rather than skipping.

End to end against a local Postgres with the synthetic seed. Figures were checked against SQL rather than read off the screen:

1. Settings: Status selected from seed, weights 30/100/0/0, `Total pipeline EUR 1,965,900.00`, `Weighted pipeline EUR 763,150.00`.
2. Kanban headers show 30% / 100% / 0% with total to weighted; cards carry Weighted Value beneath Deal Value.
3. Recalculation, through the real save path: changing Open from 30% to 50% moved the weighted total to `EUR 908,250.00`, exactly `725,500 x 0.50 + 545,500`. Clearing the weighting column nulled all ten weighted values. Editing a service from 120 to 130 moved one deal from `342,000 / 102,600` to `345,000 / 103,500`. Deleting the weighting column cleared the pointer through the foreign key and nulled the weighted values while leaving totals intact.
4. No default: cleared it in the column modal, saved, and confirmed all four options persisted as `isDefault: false`; a reload still showed no default. Re-selecting restored it.
5. Unassigned disclosure, through the real journey: with no default, a newly created deal stored `totalValue` 120 and `weightedValue` null, and settings showed Total `EUR 1,966,020.00`, Weighted `EUR 763,150.00` unchanged, and `Not weighted, no Status EUR 120.00`.
6. Renaming Deals to Opportunities resolved through the forecasting UI and rendered `Deals → Opportunities` in the audit body. All 190 preset labels across five locales were resolved and asserted present, trimmed and capitalised.
7. Demo dashboard unchanged: the original seven widgets render with zero overlaps, verified by measured geometry.

Preview deployment was not used for acceptance; local verification was requested instead. Not covered: behaviour at scale, since every check ran against ten deals and a settings change recalculates the whole company synchronously, and concurrent writes.

## Impact and rollback

One additive migration: `Company.dealWeightingColumnId`, `Deal.weightedValue` plus its index, and the `dealWeightedValue` enum member. No column is dropped or renamed and no existing option data is rewritten.

Existing workspaces are unaffected until an admin designates a column: `weightedValue` stays null and no weighted figure renders anywhere. The expanded seven-stage default applies only to companies created after this merges. Seed changes affect the synthetic demo fixtures only. `public/v1/openapi.json` gains `weight` on custom column options and `valueSums` on list responses, both additive, which places the stage weight on the public REST surface.

Clearing `Company.dealWeightingColumnId` nulls every `weightedValue` by design, so a mis-click in settings drops the stored forecast until a column is designated again. The figures recompute on the next designation.

Reverting the code leaves the three schema additions in place and unread, so no data is lost and no forward migration is needed. Drop the columns in a separate later migration, and only once no stored widget references `dealWeightedValue`, since such a row would otherwise fail to parse.


benjiwagner:
## Acceptance evidence

All eight required checks pass: `ci`, `repository-policy`, `rendered-hubs`, both CodeQL analyses, and Vercel. PR state is MERGEABLE, so the Tests job genuinely ran.

Local gates: `yarn typecheck`, `yarn lint` (`--max-warnings 0`), `yarn test` (2562 passed, 40 skipped), `yarn build`, `yarn prisma db seed`.

### Journey, run against local Postgres 16 with the synthetic seed in `self-hosted` mode

| Step | Result |
| --- | --- |
| Settings, unconfigured | Empty state shown, no weighted figures anywhere |
| Select Status column | Its four options listed with percentage inputs |
| Enter Open 30, Won 100 | Preview showed `Weighted pipeline EUR 0.00` against `After saving EUR 763,150.00` |
| Save | Persisted total became EUR 763,150.00, matching the preview |
| Board headers | `Open 3, EUR 725,500, EUR 217,650`; `Won 4, EUR 545,500, EUR 545,500`; `Lost 2, EUR 538,500, EUR 0` |
| Deal card | Workplace Hardware Rollout, EUR 342,000 at Open, showed EUR 102,600 |
| Deal detail | `EUR 342,000.00 x 30% (Open)` and `EUR 102,600.00` |
| Dashboard widget | Weighted Deal value rendered EUR 763,150 |

Database check confirmed the stored column rather than a rendered value: `Workplace Hardware Rollout 342000 -> 102600`, `Enterprise Integration Program 212000 -> 212000`, `Data Center Refresh 418500 -> 0`.

### Two defects found and fixed during verification

- `prisma/seeds/webhooks.ts` and `prisma/seeds/audit-logs.ts` selected deal rows without `weightedValue` and then parsed them with `DealDtoSchema`, so `prisma db seed` aborted with a ZodError. CI runs the seed twice before the gates, so this would have failed the pipeline. Both selects now include the field.
- `setDealStageWeights` used `findUniqueOrThrow` without an `OrThrow` name, breaking the repo-naming convention gate. It now uses `findUnique` and returns early, which is also the correct behaviour when no column is designated.

### Preview status

The preview is live and READY at `feat-deal-weighted-pipeline-value.customermates.com`, and the build applied the migration successfully. The signed-in journey has **not** been run there: the preview serves `APP_MODE=cloud`, so it has no demo auto-auth and its sign-in page offers only email and password, Google, or Microsoft. Driving it requires authenticating, which is a human step. Local `self-hosted` mode was used instead to exercise every write path.

benjiwagner:
## Seed-select drift guard

Follow-up to the seed crash noted above. Both seed files hand-maintained an identical deal select and then parsed the rows with `DealDtoSchema`, so a new required DTO field broke `prisma db seed` while every unit test still passed. The fixture suites mock Prisma and hand-build their rows, which hid the drift until CI reached a real Postgres.

The select now lives in `prisma/seeds/deal-select.ts` and both seeds import it. `prisma/__tests__/deal-seed-select.test.ts` asserts it covers every field the DTO requires.

Verified the guard actually guards, rather than only that it passes: removing `weightedValue: true` from the shared select makes it fail with `expected [ 'weightedValue' ] to deeply equal []`. Restored afterwards.

Re-ran on a throwaway database (`seedprobe`, created and dropped, shared dev database untouched): `migrate deploy` then `prisma db seed` twice, both exit 0, 10 deals seeded with `weightedValue` null, which is correct since the synthetic seed designates no weighting column.

Gates after the refactor: `yarn typecheck`, `yarn lint`, `yarn test` (2574 passed, 40 skipped), `yarn build`.

## Note on the Vercel failure

The Vercel check went red on `cc4169f6` with `next/font/google` module-not-found for the Lora font, a transient font fetch during the build and unrelated to this change. Confirmed by a clean local `yarn build` on the same commit, then redeployed that same commit rather than pushing anything new. It is green.

PR #77, opened 2026-08-17
fix(role): refuse edits to the role the caller currently holds

benjiwagner:
Closes [CUS-239](https://linear.app/customermates/issue/CUS-239/block-a-user-from-editing-the-permission-set-of-the-role-they-hold).

## Summary

`UpsertRoleInteractor` is gated on `users.create` AND `users.update`, and its only self-protection was a system-role check. **Nothing compared the target role against the caller's own `roleId`**, so a holder of a non-system role carrying user management could rewrite that role's permission set and grant themselves company administration, delete rights, or read-all on every resource. The change lands on the caller immediately.

The neighbouring guard covered the adjacent case but not this one: `AdminUpdateUserDetailsInteractor` blocks changing *which* role you hold, so editing the role itself was the remaining path.

## Context

**This one was reachable.** Unlike CUS-219, no other guard stood behind it.

Scope of exposure: system roles are already protected, so the built-in Admin role was never editable this way. The exposure is workspaces that created a custom role carrying user management, which is the normal way to delegate member administration.

Two corrections to the issue as filed, both from reading the code:

- **There is no MCP or REST path to role writes.** The issue's scope asked to cover the MCP surface. `UpsertRoleInteractor` has exactly three references (`core/di.ts`, `app/[locale]/(protected)/company/actions.ts`, its own file); `find app/api -ipath "*role*"` is empty, and `manage_team` only reaches `AdminUpdateUserDetailsInteractor`. The single write path is the `upsertRoleAction` server action, so the one guard covers everything. No MCP change was made.
- **The existing system-role `throw` is left alone.** The issue asked for validation results instead of raw throws; I applied that to the new refusal only, since the acceptance criteria also require system-role behaviour to be unchanged.

The modal mirrors the server rule through the store's read-only contract rather than new props at the call sites: `RoleModalStore` gains `isOwnRole` and folds it into `isReadOnly`, which the existing `isDisabledOrSystemRole` getter already feeds. Every permission control, the name and description fields, Save and Delete all gate on it automatically. Same shape as the fix in #75.

## Validation

`yarn typecheck`, `yarn lint:check`, `yarn conventions:check` (300 tests), `yarn test` (**308 files, 2587 tests, 51 skipped, 0 failing**), `yarn i18n:audit` and `yarn build` all green.

`yarn conventions:check` caught the new error key missing from `DIRECT_COMMON_ERROR_KEYS` in `tests/conventions/i18n-key-resolution.test.ts`; registered it alongside `roleSystemRequired`, which uses the identical direct-`t()` pattern. All five locales updated.

**The hole was reproduced before it was fixed.** `features/role/__tests__/upsert-role.interactor.test.ts` was written first and run against the unpatched interactor. Exactly one of its four cases failed: a payload setting the caller's own role to `canManage: "yes"` and `readAccess: "all"` on every resource returned `ok: true` and reached `upsertRoleOrThrow`. The other three passed, confirming system-role protection and role creation were never the problem. With the guard, all four pass.

Exercised on a local self-hosted instance. Seed data has one system role (Admin) and two custom roles, so I granted the "Sales Manager" role `users.create/update/delete` in the local database to construct the actual threat model, then signed in as its holder, Sofia Rossi:

- **Her own role (Sales Manager):** warning shown, name and description read-only, **all 74 permission radios disabled**, Save disabled, Delete absent. The "Company / Manage" control, the escalation target, sits at "No" and cannot be moved.
- **A role she does not hold (Customer Success):** no warning, **0 of 74 radios disabled**, fields writable; flipping a radio enabled Save and surfaced Reset. Her legitimate ability to administer other roles is intact. The edit was discarded, not saved.

## Impact and rollback

A user can no longer edit the role assigned to them. Editing any other non-system role, creating roles, and system-role protection are all unchanged and covered by tests. Two new translation keys across five locales; no schema, migration, or API surface change.

One behaviour note for review: a workspace whose *only* user-managing role is a custom role held by its only administrator can no longer edit that role from the app, since there is no second user to do it. The built-in Admin system role is the intended escape hatch, and it stays assignable. Worth confirming that matches your expectation before merge.

Revert the single commit to roll back.


benjiwagner:
### End-to-end server verification (the gap my original evidence left)

The PR body cited the interactor test as the server proof. That test **mocks `getTranslations` to return the key itself**, so it could not have caught a broken runtime translation lookup. The new strings had never actually rendered. Closed that now.

Method: temporarily neutralised the **client** guard (`RoleModalStore.isOwnRole` forced to `false`), which reproduces the pre-fix UI exactly (own role editable, 0 of 74 radios disabled, no warning), leaving the server as the only thing standing. That is the "including by hand-crafted request" acceptance criterion.

Signed in as the custom-role holder, flipped `company.canManage` to "yes" on her own role, and hit Save. The request reached the server action.

- **Refused.** Modal stayed open, nothing saved.
- **Toast rendered the real localized string:** "You cannot change the permissions of the role assigned to you."
- **Database unchanged:** the role still has only the seeded `company:readAll` / `company:readOwn`, with no `company:create/update/delete`, which is what `canManage: "yes"` writes.

Repeated under German (`displayLanguage=de`) to prove locale resolution beyond English: "Du kannst die Berechtigungen deiner eigenen Rolle nicht ändern." Save button read "Speichern", confirming the locale was genuinely active. The second new key renders too: "Diese Rolle ist dir zugewiesen und kann nicht bearbeitet werden."

The temporary patch was reverted with `git checkout`; `git status` and `git diff HEAD` are both empty, so the tree matches `44834291` exactly. Reloaded afterwards to confirm the guard is live again: 74 of 74 radios disabled, name read-only, Save disabled, warning shown. Role tests re-run, 13 passing.

benjiwagner:
### Browser verification of the two remaining cases

Completes the end-to-end pass. Both had been covered by interactor tests only.

**Role creation (no `id`) still works.** As the custom-role holder, created "E2E Verification Role" through the modal with `tasks.canManage = yes`. Saved successfully and the row appeared. Confirmed in the database: `isSystemRole = false` with exactly the intended permission set (`tasks:create/update/delete` plus the defaults). The guard correctly stays out of the way when `data.id` is undefined, and the new-role modal shows no own-role warning and 0 of 74 radios disabled.

**System-role refusal is unchanged, and it refuses on the server too.** Client side, the Admin role shows the pre-existing "This is a system role and cannot be edited or deleted." alert with 74 of 74 radios disabled and Save disabled. To exercise the server, I neutralised the client system-role gating (`isSystemRole` and `isDisabledOrSystemRole`), flipped `tasks.canManage` to "no" on Admin, and saved:

- **Refused.** Server log: `⨯ Error: Cannot update system roles`.
- **Database untouched:** Admin's `updatedAt` is still the seed value `2025-08-06 08:05:00`, and it is still a system role.

One observable difference worth naming, since it is now confirmed rather than predicted: because the system-role path is a raw `throw` rather than a validation result, a bypassed client gets the generic **"An unexpected error occurred..."** toast, whereas the new own-role refusal produces the specific localized message. That asymmetry is pre-existing and deliberately left alone here, since the acceptance criteria require system-role behaviour to stay unchanged. Worth a follow-up issue if you want the two aligned.

All temporary patches reverted with `git checkout`; `git status` and `git diff HEAD` are empty, tree matches `44834291`. Reloaded afterwards and re-confirmed all three states: Admin 74/74 disabled with the system alert, own role 74/74 disabled with the own-role warning, another role 0/74 disabled with no warning. Role tests re-run, 13 passing.

benjiwagner:
### Moved the guard into the write precheck (`12b934f1`)

Review feedback: the check belonged in the validation layer, not the interactor body. It did. `ValidateAssigneeGuardInteractor` is the existing precedent for exactly this shape, an authorization check that compares input against the current user, implemented as a pure validator and wired through a `Write` precheck.

Now `validateOwnRoleGuard` sits beside `validateAssigneeGuard`, emits a `CustomErrorCode` issue on the ctx, and runs from the precheck next to the existing role-id validation.

Three things fell out of the move that make it more than cosmetic:

1. **Prechecks run outside the transaction.** Decorator order is `Validate → Precheck → ValidateOutput → Transaction → invoke`, so a refused write no longer opens a transaction at all. The old placement ran inside one.
2. **`getTranslations` and the hand-built `ZodError` are gone.** `getZodParseContext` builds its message table from `Object.values(CustomErrorCode)`, so the code resolves to a message automatically.
3. **The manual registry entry is gone.** `CUSTOM_ERROR_CODE_KEYS` is enum-derived, so `roleSelfEditForbidden` no longer needs its hand-added line in `DIRECT_COMMON_ERROR_KEYS`. One less thing to keep in sync.

Confirmed `getTenantUser()` is available in a precheck before relying on it: `TenantInteractor` is a class decorator that wraps the fully-decorated `invoke` in `runWithTenant`, so the whole chain including the precheck runs inside tenant context.

**Behaviour is unchanged**, re-verified in the browser rather than assumed. With the client guard neutralised again, saving a self-targeting escalation still produces the same localized toast, "You cannot change the permissions of the role assigned to you.", now routed through the shared error map. The role still has only `company:readOwn` / `company:readAll` afterwards, with no `create/update/delete`.

Four unit tests added for the pure guard (no id, caller holds no role, different role, own role), mirroring the `validateAssigneeGuard` tests. Full suite 2591 passing, up from 2587. `yarn typecheck`, `yarn lint:check`, `yarn conventions:check`, `yarn i18n:audit`, `yarn build` all green.

Note on CI: the Vercel check failed once on `P1001: Can't reach database server` during `prisma migrate deploy` against the preview branch database. Infrastructure, not the diff, since the local build passed and the same code deployed fine on the previous push. A redeploy of the same commit went green. All 8 checks pass.

benjiwagner:
### Alert placement and gating (`7932f44e`)

Review feedback: the alerts sat in two different places, and the own-role alert showed even to viewers who could not edit anything.

**Placement.** The split was across the two modals, not within one. `role-modal.tsx` rendered its alerts as the first children of the card body; `user-modal.tsx` rendered the own-profile warning *last*, below every field and directly above the actions. The member modal's alert moves to the top so both modals explain themselves before the form.

**Gating.** The own-role alert now requires `canManage`. Its copy says "ask another administrator to change it", which is misleading for a viewer who could not edit any role: the modal is already fully read-only for them via the permission path, and `FormActions` renders nothing at all.

Verified locally against a seeded custom role, asserting DOM order with `compareDocumentPosition` rather than eyeballing:

| Case | Result |
| --- | --- |
| Own role, `users:readAll` only | **no alert**, 74/74 radios disabled, Save not rendered |
| Own role, after granting `create/update/delete` | alert present, positioned above `#name` |
| Own user profile | warning positioned above `#email`, settings link resolves to `/en/profile/settings` |

**One deliberate asymmetry for review.** The member modal's warning is intentionally left ungated. It points at `/profile/settings`, a route every user has regardless of user-management rights, so it stays accurate and useful for a viewer without them, unlike "ask another administrator". Happy to gate both if you would rather have strict uniformity.

Gates: `yarn typecheck`, `yarn lint:check`, `yarn conventions:check`, `yarn test` (308 files, 2591 passed, 51 skipped), `yarn build`. All 8 CI checks green.

**Unrelated finding, not addressed here.** A hydration mismatch fires app-wide in dev, including on `/en/dashboard` with no dialog open: Radix `useId` values differ between server and client on the sidebar collapsibles, the sidebar tooltip and dropdown trigger, and the dashboard accordion. Confirmed unrelated to this branch (the changed files contain no sidebar references, and it reproduces on a page that renders neither modal), but it puts a permanent "1 Issue" badge in the dev overlay. Worth its own issue if you want it chased.

benjiwagner:
### Precheck shape and single-use validator (`691d30b6`)

Review feedback: the precheck should follow the house pattern, and the validation could be inlined to remove an abstraction level. Both correct.

**Precheck shape.** The inline multi-statement lambda I introduced was the only one in the repository. Counting every `precheck:` declaration outside tests: ~30 delegate to an injected `*WritePrecheckInteractor` (the five CRM entities), **7 delegate to a named method on the interactor** (`admin-update-user-details`, `upsert-webhook`, `upsert-widget`, `modify-entity-relation`, `upsert-custom-column`, `start-chat`, `resolve-provider-profile`), 6 are one-liners delegating to a single validator, and 1 was the lambda. `upsert-role` was previously in the one-liner group; adding a second rule should have promoted it to the named-method group, which it now is.

**Single-use validator.** `core/validation/` holds rules with many callers: `validateAssigneeGuard` reaches five entities via `ValidateAssigneeGuardInteractor`. The own-role rule had exactly one caller, and the closest precedent for a one-caller bespoke rule is `upsert-widget`, which keeps `validateKind` and `validateActivityFilters` as private methods rather than shared modules. `core/validation/validate-own-role-guard.ts` is deleted and the rule is inline in the precheck.

Net **-54/+9** lines. The emitted `CustomErrorCode`, the resulting validation result and the localized message are all unchanged.

**A test I removed rather than kept.** Moving coverage off the deleted pure function, I wrote a "caller holds no role" case that passed for the wrong reason: `TenantInteractor` re-wraps every call in `runWithTenant(user, ...)` with the DI-resolved user (`tenant-interactor.decorator.ts:65`), so the roleless context the test established was discarded and the caller was still the ordinary mock user. The scenario is also unreachable in production, since a user with no role fails the `users.create` + `users.update` gate before the precheck runs. Deleted rather than repaired. The four remaining interactor cases (own role, other role, system role, create) exercise the real decorator chain.

Gates: `yarn typecheck`, `yarn lint:check`, `yarn conventions:check`, `yarn test` (308 files, 2587 passed, 51 skipped), `yarn build`. All 8 CI checks green.

benjiwagner:
### Accessor sweep across the remaining interactors (`59678b0f`)

Seven interactors imported the free `getTenantUser` while already extending `AuthenticatedInteractor`, and therefore `UserAccessor`, which exposes `user`, `userId` and `companyId`. The accessor is the established shape by **197 uses to 9**, so these were stragglers rather than deliberate choices. All seven now use it.

**Three left alone deliberately.** `UserAccessor.user` and `BaseRepository.user` are the definition sites the pattern is built from. `CreateCheckoutSessionInteractor` extends nothing, so it has no accessor to inherit; giving it one is a wider change than this cleanup and it sits on a payments path in `ee/`.

**Verified end to end on a local instance, not by types alone**, because each of these reads an identity that lands in a persisted row or an outbound email:

| Interactor | Evidence |
| --- | --- |
| `get-user-details` | Profile settings renders the signed-in user |
| `update-user-details` | Audit row `user.updated`, `entityId` = caller's id |
| `update-company-settings` | Audit row `company.updated`, `entityId` = company id |
| `admin-update-user-details` | Renamed another member successfully |
| `admin-update-user-details` guard | Client guard bypassed, server refused with `Cannot update own details.`, record unchanged |
| `send-feedback` | Email composed as "General Feedback from Maximilian Bergmann" with the caller's address |
| `invite-users-by-email` | Invite composed as "Maximilian Bergmann invited you to their team" |

The last two are the strongest: the accessor's value appears literally in the outbound payload. Both were logged local-only (`[EmailService] EMAIL (local only)`) against a placeholder Resend key, so nothing was sent.

**One gap, stated rather than papered over.** `create-support-ticket` has no UI surface at all; it is reachable only from `support.mcp-tools.ts:23`, with no caller under `app/`. It was not exercised in the browser. Its edit is the identical one-line substitution as `invite-users-by-email`, which was.

Gates: `yarn typecheck`, `yarn lint:check`, `yarn conventions:check`, `yarn test` (308 files, 2587 passed, 51 skipped), `yarn build`. All 8 CI checks green.

PR #82, opened 2026-08-18

benjiwagner:
## Summary

Ships the customer-facing in-app AI Assistant as a cloud-only Enterprise Feature under `ee/agent-chat`. The authenticated panel reuses the existing MCP capabilities, requires fresh approval before consequential outbound actions, meters model usage against per-seat credits, and supports all five application locales.

The final hardening round also:

- gives hosted tools and public MCP the same structured interactor-outcome boundary;
- keeps expected authorization, inactive-user, validation, entitlement, and approval outcomes out of Sentry while unexpected failures remain redacted and captured once;
- preserves discovery and completed work across long-turn continuation without replaying tools;
- raises the bounded long-workflow envelope and retains cost-aware, fail-closed admission;
- renders distinct localized activity names instead of generic workspace labels;
- accepts model-emitted preferred/legacy custom-field option aliases only when their parsed values are deeply identical, then invokes the interactor once;
- seeds a deterministic Pro credit anchor and active-seat timestamps so demo data is visually representative and the seeded preview user is entitled in cloud mode. Demo-mode execution remains blocked.

## Context

CUS-125 covers the end-to-end hosted assistant experience.

- Demo mode is a read-only showroom: the seeded panel opens by default, while every mutating assistant action remains blocked.
- Cloud users can open the panel from the sidebar or Cmd+J.
- The provider key is passed explicitly to the model client, and the operational kill switch is enforced at every assistant entry point.
- Credit reservation, live-run fencing, chat admission, and pre-provider cleanup use explicit atomic transactions so a failed admission leaves a durable zero-credit released ledger row without a poisoned turn.
- Support requests send the subject, description, and a sanitized recent chat transcript to the operator email. Success is shown only after the email provider accepts delivery, and replies return to the user's account email.
- Support ticket numbers, persisted lifecycle state, in-app support replies, and idempotency claims are removed.
- The original assistant migration keeps its applied name but now creates only the final schema shape. Two branch-only cleanup migrations were folded into it.

## Durable, model-agnostic, balance-gated agent (steps 1-6)

The assistant is being turned into a durable CRM agent on this same PR, in reviewable commit slices. Landed so far:

- **Background tenant context** (`199180fd`). `runAsBackgroundTenant(userId, fn)` lets a workflow step assume an identity without a request session; `TenantInteractor` reuses an ambient identity when one exists. A convention test pins which five entrypoints may assume an identity, and restricts the new helper to `workflows/`. Reviewed as a security change: this is the only new authorization path.
- **Pinned, tier-aware pricing** (`f100ba34`). `model-pricing.snapshot.ts` pins rates per *endpoint* from the gateway's endpoints API, which is authoritative where the flat model list disagrees by 2x. Pricing is selected by whole-prompt token count with half-open tiers and no tie-break, and computed in exact integer arithmetic. Unpinned or unpriceable models throw at load; `UNKNOWN_MODEL_PRICING` and the prefix family match are gone.
- **AI SDK 7 + Workflow 5 beta** (`6ba57c05`). Exact pins plus `resolutions` for `ai` and `jose`. This is deliberate adoption of a beta package for production durability.
- **Collapsed durable schema** (`5ebf7dfc`). `AgentTurnRequest` becomes the run aggregate; `AgentRunRound` and `AgentToolReceipt` added; `costMicrocents` widened to `BigInt` because a 32-bit column caps a single usage row at $21.47 and would overflow inside the settlement transaction for a premium model.
- **Multi-step context fix** (`77145fed`). AI SDK 7 made `step.response.messages` per-step where 6 was cumulative, so the model never saw its own tool results after the first step and multi-part questions looped and answered nothing. 3,543 tests stayed green throughout, because the fixture hand-built SDK 6 histories: a test that encodes a dependency's contract cannot detect that contract changing.
- **Gateway cutover and model catalog** (`58710c70`). `MODEL_CATALOG` maps an opaque key to a gateway slug, a pinned serving provider and a per-model envelope; adding a price point is config, not surgery. Envelopes moved from serialized bytes to prompt tokens, which removes a guard that compared a byte count against a token boundary and would have refused 24 of 39 tiered catalog models. Models are addressed by id string, never a provider instance, so no API key can reach a durable step argument. `OPENAI_API_KEY` is retired in favour of `AI_GATEWAY_API_KEY` as the single billing source.
- **Measured-cost settlement** (`7f0ec4b8`). Turns are charged what the gateway actually billed, read inline from the response. A call that never billed releases its reservation; a cost that cannot be trusted is quarantined against the pinned estimate rather than throwing, because at settlement the money is already spent. `retainReservation` is no longer reachable from the request-bound runner.

Deleted with no replacement or backfill, because no agent conversation exists outside preview and local: the OpenAI tool-search / deferred-loading / stored-item machinery, the `item_reference` retention, `store: true`, the single-member model-lane union, and the usage-evidence charging path.

## Durable agent, continued (steps 7-13)

The turn now runs as a Vercel Workflow rather than inside the HTTP request. All four harms this issue opened with are fixed.

- **Tool identity registry** (`52518cf9`). Approval rules are keyed by an identity triple rather than a bare tool name, because `search` and `fetch` are ubiquitous on public MCP servers and the first external connector would otherwise shadow an internal rule. `requiresApproval` refuses a non-internal source *before* reading the tool's own `readOnlyHint`, which an external server controls.
- **Durable turn state** (`90fff6d5`). Exactly-once tool receipts around transactional mutations, per-round persistence, a lease heartbeat, and a reservation keyed to the turn rather than the run.
- **The turn runs in the workflow body** (`b2dc5ca0`, `17ef324e`). Driving `WorkflowAgent` from inside a step silently collapses its internal steps into one, giving turn-level rather than round-level durability: a resume would have re-charged every model call. Run from the body, each model round and each tool call is its own durable step. Approvals use the SDK's own pause: the agent returns with the call unexecuted, the body persists the request and suspends on a hook raced against an expiry, then resumes the same stream.
- **The panel no longer owns the work** (`ca562f2f`). The route admits, dispatches, and streams the durable run. Cutting the client off after one event leaves the turn running to completion. A reattach endpoint replays from a chunk index, and the SSE id carries that index so a client resumes exactly.
- **The durable stream is translated, not proxied.** It carries raw tool output, model identifiers, per-call usage and gateway cost. An allowlist emits only the events the client contract defines, so a new chunk type from a future SDK version cannot leak by default. The SDK's own transform would have forwarded tool output verbatim.
- **Interface tools** (`b999ef49`). A panel tool is registered with no execute function, which makes the agent pause and hand the call back; the body writes the command, suspends, and resumes with the result. With no panel attached the command lapses and the model says so.
- **Stop actually stops** (`08af7471`). Cancellation is a durable fact checked at open, at every step boundary, and after any suspension. The turn settles what it used.
- **One run per conversation** (`3c64dd94`). The lease was keyed by user, which meant one turn suspended on a thirty-minute approval locked the assistant out of every other conversation. Engine load is now a counted admission check instead of a unique index.
- **Safety stops** (`c92146ba`). The durable path had none of them; the step ceiling was hiding it. The wall clock is deliberately dropped, because a run that waits for an approval is the point of durability.
- **Credits reserved a few rounds at a time** (`7a07544d`). Forty credits up front became eight, topped up as the turn proceeds, so a user with thirty credits left can start a turn that costs one. This deleted the degradation ladder, which silently gave a user low on credits a *worse* assistant rather than a shorter one.
- **No step ceiling** (`71c55628`). The loop drives one thirty-two round segment at a time, compacting context between segments. A workspace audit that used to stop at twenty rounds now completes in thirty-eight.
- **Model choice** (`e68f9bd2`). An opaque key, localized copy and a server-computed relative cost band; no model identifier, dollars or token counts cross the wire. Scoped to new conversations, because prompt caches are per model.
- **The request-bound runner is deleted** (`d230cba1`). Roughly two and a half thousand lines including its tests, plus request-signal cancellation, both poll loops and the per-turn step ceiling. The eval harness now drives the real endpoint over HTTP, which is why the runner could go.

Two defects worth flagging for review, both money-related and both found by comparing counters that must agree rather than by a test:

- `cancelled ||= await persistRound(...)` short-circuits, so the final round of a cancelled turn was billed but never recorded.
- An exception thrown in `onStepEnd` is swallowed by the agent loop: the turn completed, created a record, answered correctly, and charged nothing. Accounting now precedes policy, and a throw stops the run loudly.

Also fixed outside the assistant: `runAsBackgroundTenant` loaded the DI barrel with a dynamic import on every call, which under load returned a partially initialised namespace (`getUserService is not a function`). It runs on every durable step (`7371e0c1`).

**Deliberately not done.** The `useChat` and `WorkflowChatTransport` rewrite. Its justification was that deleting the request-bound loop would break the panel; that did not happen, because the durable run is served over the existing event contract. Adopting it now would mean writing a UI-chunk transform *with* our own privacy filter (the SDK's forwards tool output verbatim), a history-to-UIMessage converter, and a second state system beside MobX. It is more custom glue, not less.

## Panel and permission work after the durable rewrite

The assistant panel was simplified down to what it needs, and one authorization bug fell out of using it.

**The composer is one line: input, credits ring, send.** A model picker and a per-model speed selector were built and then removed, along with everything that existed only to feed them, so the assistant runs on its configured default and nothing selects a model or an effort level. Two findings from that work are worth keeping even though the code is gone: reasoning effort has to travel in the provider's own namespace, because the gateway option bag accepts any key and silently drops what it does not recognise (`gateway.reasoningEffort` changed nothing; `openai.reasoningEffort` moved a turn from 0 reasoning tokens at `none` to 1552 at `max`); and one live turn on Claude Opus 5 charged 101 credits for a two-part question, because the tool catalog dominates the prompt at that model's rates.

**Reloading mid-turn no longer loses the run.** The durable stream survived a reload from the start, but nothing on the client rejoined it, so the panel showed committed history and no progress until the next visit. The conversation payload now reports whether a turn is in flight, and loading a conversation in that state reopens the durable stream through the reader the send path already uses. Rejoining also restores stop, because the client takes an abort controller for a rejoined run exactly as it does for one it started.

**A permission check could not run without a request session.** `hasPermission` resolved the caller through `getActiveUserOrThrow`, which reads the session, so every write guarded by `hasPermissionOrThrow` threw an auth error inside a durable step. The assistant reported it as "Sign in to use this action" while the user was signed in, and creating a custom field failed that way on preview. It now prefers the ambient tenant identity and falls back to the session, the way `ValidateAssigneeGuard` already did; the rules are unchanged and `getActiveUserOrThrow` keeps its session-only meaning for callers that must prove a request made the call. This is the same class as the step 8 guard failure: reads succeed in the background and only writes fail.

**Every custom field type is reachable, and onboarding hands over cleanly.** Two bugs that only appeared by using the assistant: single-select was the only field type it could create, and the panel stayed unmounted after onboarding until a refresh. Both are covered in the validation section below.

**Interface work.** The activity timeline now carries the live step as its last row rather than floating the indicator below a closed group, its header names the running step instead of showing a settled duration mid-turn, each connector segment grows in with its row, and a label holds for at least 700ms so a tool call that finishes in milliseconds cannot strobe the header. Alongside: relative dates on chat rows, a busy state on the sidebar entry, a destructive stop control, an icon-only scroll return in the lower right, `Cmd+J` toggling rather than only opening, and sidebar links showing navigation progress in their own icon.

## Validation

Current head: `55b803cf5793705324edb6beed004311532bca0c`. All required GitHub checks are green; the full current-head UI matrix is recorded below.

Historical checkpoint: `7f0ec4b832b0ebe180888dfefd0db837af4c8e68`.

Re-run at that head:

- `npx tsc --noEmit`, zero-warning full ESLint, and `yarn build`
- full database-enabled Vitest: 407 files, 3,556 passed, 1 intentional skip
- zero schema drift after applying all 34 migrations to a fresh database from zero state
- GitHub CI green by commit SHA: `ci`, repository policy, rendered hubs, and both CodeQL analyses
- live local end-to-end through the AI Gateway on `openai/gpt-5-nano`: a real multi-tool turn settled at `costSource = measured`, 296,067 microcents, which equals to the microcent what the pinned snapshot independently computes for the same token split; a second turn earlier settled at 331,309 microcents with the same agreement
- a free-tier 429 mid-session proved the previous charging rule wrong in production terms: zero provider attempts and zero tokens still retained the whole reservation. That path now releases, and the regression is pinned by a test
- convention guards: no production file constructs a provider; every model invocation stays behind the metered runner; and no model identifier, dollar or token value appears in the credit summary, an agent-chat client bundle, or the turn's own event stream. Both new guards were mutation-tested by planting real violations

Earlier validation, at head `c8bc98fc32292a9821f32aac1924562f53285b7f`:

- `npx prisma format`, `npx prisma validate`, and `npx prisma generate`
- `npx tsc --noEmit`
- zero-warning full ESLint
- full database-enabled suite before the seed-only delta: 402 files, 3,518 passed, 1 intentional skip
- current-head full Vitest: 402 files, 3,427 passed, 92 database-gated skips
- focused synthetic identity/fixture contract: 13/13 passed
- fresh disposable PostgreSQL: all 34 migrations applied; canonical synthetic seed converged twice; Pro subscription anchor and all three active-seat timestamps persisted
- production build passed before the seed-only delta; the exact current head compiled successfully in Vercel after running migrations and the canonical seed
- isolated localhost provider E2E: the formerly failing single-select custom field was created exactly once with all three options and no retry
- long localhost provider E2E: one discovery plus 17 ordered CRM steps completed in one turn, with 17/17 successful activities, no replay, no attention item, no continuation prompt, and a terminal `completed/completed` result
- database proof for that long turn: exactly 1 field; 3 contacts, organizations, services, deals, tasks, and widgets; exact contact option values and requested widget configurations
- reload proof: conversation, 2% meter, and all three dashboard widgets persisted
- exact preview deployment: `dpl_9vX2fhaiAeWULEBf4hjCNYW1SSRw` (`READY`) at https://feat-cus-125-assistant.customermates.com
- preview provider smoke: seeded Pro user had an enabled composer and 500-credit-period meter; the single-select field succeeded and appeared as a Contacts table column after navigation
- preview observability: clean browser console, no HTTP 500 logs; only the known PostgreSQL SSL-mode deprecation warning
- GitHub CI, repository policy, rendered hubs, CodeQL, and Vercel all passed
- independent read-only architecture and deployment audits found no remaining application-code blocker

### Validation at head `d230cba1`

- `npx tsc --noEmit`, zero-warning full ESLint, `npx next build` (44 steps, 4 workflows compiled)
- full database-enabled suite: 416 files, 3,596 passed, 1 intentional skip
- durability, driven live against the local gateway on the shipped model:
  - client cut off after one event; the turn completed and persisted its full transcript with nothing attached
  - reattach from the next chunk index produced zero duplicate events and assembled the reply text cleanly across the boundary
  - approval approved, rejected and expired: the record was deleted only on approval; the pending request was cleared in every case
  - interface command with the panel attached changed the page; with nothing attached it lapsed and the model said so
  - stop mid-run halted at the next boundary with five rounds recorded against five model calls; stop while suspended on an approval woke the run within two seconds and left the record intact
  - one user held two leases across two conversations at once; both ran and both released
  - a workspace audit ran thirty-eight rounds to `completed`, reservation growing from eight credits to eleven and settling at nine, measured, no policy breach
  - forcing the per-round reserve past the allowance stopped the turn as `partial`, charged the one credit used, and told the user their credits ran out
- browser journeys as the seeded user: record creation, the approval prompt, navigation by interface tool, and stop
- migration verified from zero against a shadow database: the chain reproduces the schema with no drift
- deletion gate: no surviving reference to `runAgentLane`, the poll-interval constants, the step-limit constant, the tool-search machinery, `item_reference`, or `AGENT_MODEL_ID`

### Validation at head `386d7c79`

Asked to prove the previous head end to end, I re-derived the consequences of parking a suspended turn in `awaitingApproval` and found four regressions the whole suite had missed, then found two further defects that only a repeated live run exposed. All six are fixed here.

**The status write was wrong, and the lease extension alone was right.** `awaitingApproval` is an *active* state, but six call sites express "a turn is in flight" as `status: "running"`. Writing it meant **stop did nothing** while a turn waited for an approval, the conversation became **archivable and deletable** underneath a live run, and a duplicate send **threw on an unknown status** instead of answering 409. Extending the lease over the approval window fixes the original bug on its own and leaves every guard working, so the status stays unwritten and the sweeper is untouched. The database tests now pin the lease horizon, which is the invariant that matters, and were mutation-tested.

**A run announced its suspension before it could be resumed.** The resume hook was created *after* the approval request was published, so a user who answered promptly hit a window where the hook did not exist. The resume returned false without complaining and the run slept out its entire window: thirty minutes for an approval, thirty seconds for every interface command the panel answered instantly. The hook is now reserved before the request reaches anyone, which the runtime supports because `createHook` claims its token at call time. The live eval went from one arbitrary scenario stalling per run to ten of ten, and from 550 seconds to 157.

**A fatal step failure left the client waiting forever.** The workflow reported the error and rethrew without closing the durable stream, so the panel waited for a terminal event that never came while the turn stayed running until the sweeper settled it. The failure path now closes the stream before rethrowing.

**One interface-target query could not cover several pages.** The catalog filter matched the whole query as a single substring, so the one focused query the system prompt asks for ("contacts, deals and the dashboard") matched nothing and the assistant answered that the interface exposes no matching targets. Any word may match now, and an unmatched query falls back to the whole catalog instead of stranding the model. Mutation-tested by restoring the all-words rule.

**Stop lost its fallback to the interface hook** when the wake helpers were collapsed, so stopping a turn that was waiting on the panel no longer woke it. Restored, with three cases pinning the fan-out.

Gates: `npx tsc --noEmit`, zero-warning full ESLint, `npx next build` (45 steps, 4 workflows), full database-enabled suite 417 files / 3,603 passed / 1 intentional skip.

`yarn agent:eval` passes **ten of ten**, twice consecutively on this head, in 157 and 164 seconds. The journey covers model pinning across turns, reattach with zero duplicate sequence numbers, an approval surviving a sweep past the ordinary lease horizon, stopping a running turn, and stopping a turn suspended on an approval. Every scenario asserts recorded rounds equal the reported round count and that measured settlement equals the sum of its own rounds.

Live in the browser as the seeded user, on this head:

- delete approval: the lease held 35 minutes of headroom, the turn stayed `running` so every in-flight guard still applied, approving completed the turn immediately, deleted the contact exactly once, and settled `completed` with 3 rounds and 1,016,588 measured microcents equal to the sum of those rounds
- guided tour: a six-step tour across Contacts, Deals and the Dashboard rendered with real highlights and notes, which is the case that previously failed
- an earlier head also proved a nine-minute suspension surviving a real panel close and reopen, and a stop mid-run terminating `cancelled` with 6 rounds and 1,721,100 microcents equal to the sum of those rounds

**Operational requirement found while proving this.** The workflow world's Postgres pool defaults to 10 while its queue concurrency is 50, and the `pg` pool waits indefinitely for a free connection, so durable steps stalled for fifteen to thirty minutes under load. The workflow package warns about this at startup. `WORKFLOW_POSTGRES_MAX_POOL_SIZE` must be set to at least the queue concurrency wherever the workflow world runs; it is now documented in the cloud env template.

**Assessed and deliberately left unchanged.** `runAgentTurn`'s catch has no `isExpectedError` early return, unlike `deliver-webhook`: swallowing an expected error there would complete the run with the turn unfinalized, and `reportFailure` already skips expected errors internally. Segment chaining on `deploymentId: 'latest'` stays out. A resume hook lost to a deploy or restart mid-suspension still costs the remainder of the window, because the wait is a single race rather than a bounded loop; the run recovers correctly, just not promptly.

**Preview, verified after the database reset.** The reset was necessary and its absence was doing real damage: the build before it logged `No pending migrations to apply`, because `prisma migrate deploy` matches by name and the collapsed migration kept its name through two in-place edits, so the preview had been seeding against the old schema all along. Deployment `dpl_F7hcpCUzsDTBPRf3Vt5kCb6dJEmi` on `15fe637a` applied `20260817120000_agent_assistant` and `20260820120000_drop_support_ticketing` fresh, reported all migrations applied, and completed the synthetic seed. The running app then served every public page 200, redirected protected pages 307, answered the agent message, stream and v1 routes 401, and produced zero runtime errors or warnings.

**Not verified, and only Benji can.** Everything behind the preview login. Signing in is outside what an agent may do, so the assistant itself has been proven locally rather than on preview: open the panel as a seeded user, run one multi-tool turn, approve one delete, and press stop once. Approval expiry after the full thirty-minute window was also not re-run on this head; it is unchanged code from the approvals slice and the lease work no longer touches that path.

### Validation at head `b6945819`

Merged `origin/main` (three UI fixes, no conflicts) and re-ran everything against it.

Gates: `npx tsc --noEmit`, zero-warning full ESLint, `npx next build`, full database-enabled suite 418 files / 3,613 passed / 1 intentional skip.

The permission fix was proven by reproducing the reported failure rather than by reading the code:

- two database cases assert a permission check resolves from an established tenant identity with no session, and that a user without the right still gets the permission error rather than an auth error; both fail if the session-only lookup returns
- live locally, the assistant created the exact field that failed on preview: `Lead Source`, `singleSelect` on contacts, with `Website`, `Referral` and `Outbound` all persisted, terminal code `completed`

Live panel checks on this head: the activity header moved between `Looking through Deals` and `Thinking` with a shortest gap of one second where it previously flashed for milliseconds; a mid-run reload rejoined a running turn showing eight completed steps, live progress and stop in place of send, then streamed to completion; a short message renders as a 64px bubble instead of collapsing.

`yarn agent:eval` was last run green at ten of ten, twice consecutively, before the panel work; the changes since are presentational plus the permission fix, which the database cases and the live custom-field creation cover.

**Not verified.** The preview environment on this head. Its database was reset and migrated earlier in this branch and the deployment is green, but everything behind its login is outside what an agent may drive.

### Validation at head `2e9f95c1`

Level with `main` again (two merges, the second with an additive enum conflict and a re-keyed tenant-write allowlist; main was diffed against the merge first to confirm nothing was dropped).

Gates: `npx tsc --noEmit`, zero-warning full ESLint, `npx next build`, full database-enabled suite 420 files / 3,639 passed / 1 intentional skip.

**Only single-select custom fields were reachable through the assistant.** Strict function-calling schemas require the model to emit every property, so a date field arrived carrying a placeholder select option and junk in every other option key, and the tool refused it for holding select choices on a non-select type. Date, currency, plain and email all failed that way, each reported to the user as an undiagnosable validation error. The boundary now projects the type config down to the keys that type actually has, ignores select choices unless the type is single-select, and treats an empty or null config as omitted. This removes a deliberate guard: the type is authoritative and immutable, so choices on a non-select column could never have been stored, and refusing the write cost the whole field type for nothing. The two cases that pinned the refusal now pin the new contract.

**The assistant did not appear after onboarding until a manual refresh.** `protectedEnhancementsAllowed` derives from a server-computed account state, and the wizard's client-side redirect reuses the protected layout, so that state still said `onboarding` and the entire enhancements block, the panel included, stayed unmounted. Completion now leaves the wizard with a document load so the shell re-reads the state once.

Exercised locally against a real workspace rather than reasoned about:

- every custom field type created by the assistant end to end and confirmed in the database: date, currency, plain, email and single-select, on contacts, organizations, deals and tasks
- sixteen assorted turns covering terminology, three dashboard widgets, record creation and linking, search, tasks, webhooks, team and interface navigation: thirteen completed on the first pass, and the three that did not were exactly the field types fixed above
- **zero authentication errors across every turn**, which is the check that matters after the `hasPermission` fix, since that failure mode was invisible to the whole test suite
- timings 10 to 25 seconds per turn with one 51 second outlier on interface navigation, no fatal errors, and no pool-starvation warning

**Not verified.** The onboarding flow end to end in a browser: the local browser pane shares one cookie jar, so a second seeded user cannot be signed in alongside the existing one. That fix rests on three store cases and on the mechanism above. Preview is untouched on this head.

### Validation at head `9862a67c`

A final dead-code sweep, keyed on what the shipped design actually reads. Removed with zero references each: the schema columns only segment chaining and the origin discriminator would have used (`origin` and its enum, `segmentIndex`, `lastRoundIndex`, `accruedCredits`) plus the `awaitingApproval` status value the lease-only suspension deliberately stopped writing (`waitingBudget` and `needsAttention` stay as documented seams); the old runner's tool-input sanitizer and five supporting constants; the widget dependency module nothing imports; the degradation ladder's per-step floors; the model-spec reverse lookup and microcent helpers the gateway cutover orphaned; the workflow's never-passed window overrides; and a store field that survived the picker removal as an unpopulated shell. A running reply also no longer carries its own spinner or a premature timestamp; the activity group owns the live indicator.

The collapsed migration is edited in place as before. An already-migrated database keeps the dropped columns harmlessly, since nothing selects them; a fresh deploy never creates them.

Re-verified end to end after the trim: typecheck, zero-warning lint, `next build`, full suite 420 files / 3,639 passed, `yarn agent:eval` ten of ten in 203s, and a live browser turn that created a record and settled `measured`.

### Validation at head `4f0b1f31`

A final structural review against the codebase's own siblings, executed with approval on all three findings.

- **The panel is filed like the inbox now.** Sixteen components lived in one 1,400 line agent-chat.tsx while the inbox keeps components in their own files; the panel matches that pattern, with the shell and composer staying in agent-chat.tsx and eight extracted files beside it. Two small defects fell out: a lock expression computed in two places and an activity group derived twice per row.
- **The durable vocabulary is retired.** It distinguished two runners; only one has existed for weeks. One limits file, unprefixed constants.
- **Database tests refuse to run while a dev server holds the app port**, ending the phantom fixture timeouts that recurred five times on this branch. The guard ignores non-dev-server listeners, checks once per process, and stands down on CI and for the live eval. Proven by planting a port squatter.

Everything measured as aligned stays untouched and is worth naming so it is not re-litigated: the 1,781 line repository matches the messaging repository's repo-per-aggregate size, every workflow in the repo keeps its steps in-file so agent-turn.ts does too, and the store remains one store for one surface.

Re-verified end to end after the reorganization: typecheck, zero-warning lint, full suite 421 files / 3,644 passed, live eval ten of ten in 204s, and a browser pass touching every extracted surface, with the history list, activity timeline, usage ring and a completed measured turn all confirmed working.

### Validation at head `748c199`

A second line-by-line review of the whole agent surface, this one hunting for code the durable rewrite orphaned without deleting. Everything found was removed and re-gated:

- The client's 409-running path re-posted the same message every 1.5s up to 100 times as a way of waiting for a busy conversation. It now rejoins the running turn through the reattach stream and re-sends exactly once when that turn ends; the reconcile bypass and both poll constants are gone, and the store test pins the new flow.
- `retainReservation` was never passed true, so the retained settlement branch, its guard clauses and repository filters, and the `llm.service` wrapper whose only job was defaulting the flag are deleted; `usageToTokenCounts` moved to `agent-usage-settlement` and `llm.service.ts` is gone. The schema enum keeps the dormant `retained` value until the next migration touch, noted deliberately.
- Step-ceiling fossils: catalog `maxSteps`, `AgentTurnBudget.maxSteps`, `agentTurnWorstCaseUsd`, and `AGENT_CONTEXT_ACCUMULATION_STEPS` fed nothing but tests that pinned each other.
- Smaller remnants: the transcript's invalid-call bookkeeping, unread `removedToolProtocol` relay and caller-less `setApprovalStatus`; the stream reader's write-only tool-input map; gateway-cost's caller-less error reader; two workflow-local aliases of the window constants.

Read and judged aligned, untouched: admission, all responder interactors, both routes, the durable-stream allowlist, budget and credit policy, gateway-cost, and repository guard placement. Left as a deliberate product decision for review: cross-turn context is the last 8 messages at 1,200 chars each (`AGENT_REPLAY_COUNT`/`AGENT_REPLAY_MAX_CHARS`), unchanged since before turns became long-running.

Gates on this head: typecheck clean, zero-warning lint, full suite 421 files / 3,642 passed, `next build` compiled, all checks green. The live eval and a browser approval turn on this exact head are still owed: local file access to the working tree (and its env secrets) was revoked by a macOS permission mid-session, so this commit was produced and verified from a clean clone; the database-backed suite ran against the real local Postgres.

### Validation at head `a3172ffe`

All eleven review threads are answered; six are resolved with fixes, five carry evidence replies awaiting your last word.

- **Fixed as requested**: agent-translator deleted for the house `i18n/get-translator`; `assertInvariant` inlined as plain throws (the error-contract test now exempts plain `new Error` throws, aligning it with "invariants throw, expected outcomes are returned"); every unshared interactor schema is a local `Schema`; the config action collapsed to `serializeResult` with a disabled assistant expressed as data; both model env knobs deleted (the catalog constant is the single config point).
- **Demo mode is a read-only showroom**: entitlement and availability admit demo, the eight mutating interactors are demo-blocked, the seed carries one active conversation answered with a markdown table plus three archived chats, the panel opens by default, and a blocked send restores the draft and shows the standard demo toast. Verified in the browser on `demo.localhost`: auto-open, table rendering, archived list, blocked-submit unwind.
- **Answered with evidence, no change**: ui-control strings are model-consumed tool results; agent-output-safety is a leak-redaction boundary while display uses the prebuilt ai-elements; app-errors keeps its test-pinned cross-bundle identity armor and loses only a caller-less export; the error-model explanation (three lanes, Prisma allowlist included) is on the interactor-failure-server thread.
- **Merge with main resolved**: locale unions, main's FAQ mdx structure with the branch's hosted-AI sentences grafted back, the branch's resurrected `monday-vs-hubspot` pages re-deleted to honor main's route retirement, `InvalidJsonBodyError` folded into the AppErrorCode system with five-locale copy, and two real docs-tool scoring flaws exposed by main's content growth fixed (capped body-frequency so reference pages are not outranked by chatty guides; rarest-term anchoring with bounded context so a focused excerpt keeps the queried detail inside the model-visible prefix).

Gates: typecheck, zero-warning lint, 431 files / 3,785 tests against local Postgres, `next build`, seed converges twice, all checks green, `MERGEABLE`. Still owed once the gateway key is reachable: `yarn agent:eval` and one live approval turn on this head.

### Validation at head `a02e6f1b`

The expected-failure idiom is now a named pattern, decided with Benji in review. Kind-named constructors (`fail`, `failNotFound`, `failConflict`, `failRateLimit`, `failUnavailable`, `failAuthorization`) replace `createInteractorFailure`; the kind is stamped into the issue at the call site and classification prefers it, with the set registry retained only for refinement-born codes. Every hand-rolled site migrated (agent, messaging posts/sales-navigator/outbound, checkout, legal, presets, roles, admin updates, api keys, channel conflicts); provider outcome codes stay kind-neutral deliberately; the entitlement denial passthrough keeps its shape as the feature-gating variant. Two inherited drifts fixed: send-chat-message's uncoded draft guard, and four locale keys promoted to enum codes. Gates: typecheck, zero-warning lint, 431 files / 3,786 tests, build, all checks green. The KB error model is updated in customermates/knowledge-base#89, to merge after this PR.

### Validation at head `55b803cf` — the full UI end-to-end matrix

Driven live in the browser against the real gateway, multiple users, under a $10 ceiling (actual model spend ≈ $2.40 including two eval runs).

- **Gate 0**: `yarn agent:eval` 10/10 (204s) on the final code.
- **A — new company onboarding**: sign-up → wizard → Finish lands on the dashboard with the panel already open (no refresh); trial ring healthy; the full workspace-setup turn created all six custom fields with correct types (the old strict-schema regression case), three widgets and three renames; 17 rounds = ledger to the microcent.
- **B — invite with approval gate + proration**: share-link invite → new user registers → "Waiting for Approval" → admin assigns Customer Success and activates → the seat prorates to exactly the hand-computed 175 of 500 credits, shown in the ring popover; a first turn works at 1 credit. (Also surfaced: saving the member dialog without a role shows a raw "Invalid UUID" toast — small UX finding.)
- **C — permission boundaries**: Elena (deals read-only) asking for a deal creation gets a structured authorization failure the model explains gracefully — turn completed, nothing created, UUIDs redacted; her deals-page suggestions offer no creation; Sofia gets no workspace-setup suggestion; conversation history is per-user.
- **D — credits**: mid-run exhaustion stops honestly with the localized message, charge capped at the reservation, balance exactly 0 and never negative; blocked composer swaps to the credits notice and recovers; admission at exactly 1 credit succeeds; a turn admitted before a simulated period shift settles into its original bucket while the new period reads fresh.
- **E — durability**: the cross-tab rejoin scenario **found a real bug** — losing the lease race with a new message returned a prettified string 409 without the running disposition, so no client could ever rejoin cross-tab (`e8f2867b` fixes it; verified live: one 409, reattach watches the running turn, exactly one auto-resend). Reload mid-turn reattaches without duplicates; `kill -9` mid-write resumed via the queue with the task created exactly once, one settled receipt, no double charge; an approval survived 8.5 minutes suspended across a panel close and a sweep, then approved into exactly one deletion; Stop woke a suspended run in ~1s; the queued prompt fired automatically.
- **F — eight-question MCP battery** (search, aggregation vs widgets, custom-field writes, task triage, inbox, relations, docs, terminology): all completed, every turn's rounds-sum equals its settled measured cost exactly; spot-checked facts match the database to the euro; the docs walkthrough used the new excerpt scoring and the headless interface tour lapsed honestly.
- **Merges**: main moved twice during the run; both merges resolved (the second aligned main's LinkedIn contract rewrite with this branch's structured-failure and approval-context conventions, including profileType-aware verification).
- **Known boundaries reconfirmed**: sweeping audits can hit the repeated-activity safety guard (`partial` with the localized notice) — the open product tuning question from the plan; the temporary gateway key should now be rotated.

### Validation at head `6fef0896` — Enterprise credit floor

An Enterprise subscription whose contracted allowance was never filled in previously blocked its users outright, so an existing Enterprise customer would have met a "contact us to enable requests" notice on release day for a plan they already pay for. Such a seat now falls back to the lowest allowance any plan defines, read from `PLAN_CATALOG` via `lowestPlanHostedAiCreditsPerActiveUser()` rather than hardcoded, so the floor follows the catalog if pricing moves. A contracted figure still wins whenever one is set.

The fallback makes `enterprise_allowance_missing` unreachable, so that blocked reason is removed along with its copy in all five locales and its entry in the i18n key registry.

Migration-day picture for existing users, with the create migration's `UPDATE "User" SET "agentCreditActivatedAt" = "createdAt" WHERE "status" = 'active'` backfill: active trial seats get the flat 500, Starter 200, Pro 500, Business 1,200, Enterprise its contracted figure or 200 as the floor, all unprorated because the stamped seat predates the current period. Pending and inactive users stay unstamped and prorate from the moment an admin activates them. This also retires the pre-merge task of configuring Enterprise allowances before deploying.

Gates: typecheck, zero-warning lint, 435 files / 3,826 tests, `next build`. Checks on this commit were first swallowed by a GitHub Actions major outage (jobs reported "not acquired by Runner of type hosted", zero steps, and `main` failed identically); re-triggered once Actions returned to operational.

### Validation at head `c4096467` — adjacent-surface regression sweep

Benji asked whether webhooks, event retrieval and audit logs still work, since this branch touched files well beyond the agent. All three were driven live, and the sweep found one real regression, now fixed.

- **Webhooks end to end**: a webhook created through the UI form (URL, description, secret, event selection), then a contact created through the contacts UI, produced a real delivery at a local receiver carrying the full payload and an `x-webhook-signature` header, recorded as `contact.created | 200 | success`.
- **Event retrieval**: the deliveries page renders the live delivery with URL, event, resolved element name, status and code, alongside historical rows including a failed 404.
- **Audit logs**: both actions captured in order with correct actor attribution.
- **The changed webhook MCP tool** (`manage_webhooks`, migrated to the failure constructors) listed both webhooks correctly with UUIDs redacted. That turn also drifted into an unrelated destructive `delete_records`, which the approval gate caught: run suspended with its lease extended, rejected in the UI, activity settled `cancelled`, records verified intact.
- **463 tests across all 44 webhook/audit/event files** pass.

**Regression found and fixed**: pruning the failure-kind registry to refinement-born codes missed `threadNotFound`, because `validate-thread-ids` passes the code as an argument to `checkIds` rather than writing it inline. Unstamped and unregistered, it classified as `validation`, so any request naming a missing thread answered **400 instead of 404**. The entry is restored and a convention test now derives the expectation from the validators themselves; removing it again fails by name, which is how it was caught.

Also verified: self-hosted still excludes the agent while cloud and demo include it (that predicate changed for demo mode), and the Enterprise floor resolves live across every tier — no contract 200, contract 4000, invalid contract 200, starter 200, pro 500, business 1,200.

Two pre-existing REST quirks were observed and confirmed **not** to originate here (this branch never touches `app/api/v1/webhooks/`): DELETE of a missing webhook returns 400 rather than 404, and GET of a missing webhook returns 200 with `null`.

Gates: typecheck, zero-warning lint, 435 files / 3,827 tests, `next build`; all six checks green on the preceding head.

### Live eval at head `c4096467`

`yarn agent:eval` **10/10 in 180s** against the real gateway on the exact merge head, with zero authentication errors in the server log. This closes the last verification gap: every earlier eval had run on an older head, and one attempt in between failed only because the temporary gateway key had been revoked (every model call returned "AI Gateway authentication failed", turns died with no frames, and the first failure cascaded into the rest). That episode incidentally exercised the fatal-error path correctly: turns closed their streams instead of hanging, and nothing was charged.

Full picture at this head: typecheck, zero-warning lint, 435 files / 3,827 tests, `next build`, all six CI checks green, live eval 10/10.

## Impact and rollback

Set `AGENT_CHAT_DISABLED=true` to disable the hosted Assistant without reverting the deployment.

**New environment requirement.** `AI_GATEWAY_API_KEY` must be set wherever the assistant runs; `OPENAI_API_KEY` is no longer read by any assistant path. Model selection is code-configured through the catalog; the removed `AGENT_MODEL_KEYS` and `AGENT_MODEL_DEFAULT` variables must not be provisioned. A missing gateway key fails the turn, not the deployment.

**Preview reset completed.** The isolated branch preview database was reset and migrated after the final in-place migration edits. Production still requires the human destructive-migration checkpoint described below; application rollback alone cannot restore deleted support-ticket data.

**Behaviour changes a reviewer should expect.** A turn survives the panel closing and keeps spending until it finishes, is stopped, or runs out of credits, so stop is now a durable request rather than a local abort. A turn is no longer capped at twenty rounds. A user may hold three concurrent runs, one per conversation. Credits are reserved incrementally, so a turn can begin that the balance cannot finish; it stops and says so rather than being refused up front or silently degraded.

The approved `20260820120000_drop_support_ticketing` forward migration drops the already-shipped `SupportTicket` table and its two enum types. Existing support-ticket rows are permanently removed. An application rollback does not restore that data; restoration requires a database backup plus a new forward schema migration.

Email-only support intentionally has no ticket number, persisted status, reply thread, idempotency record, or failed-attempt record. Replaying a successful request can send a duplicate email.

The two removed `20260819` migration names existed only on this feature branch. A branch preview that previously recorded them may require deletion of exactly those two rows from its isolated `_prisma_migrations` metadata before forward deployment can continue.


app/components/agent-chat/actions.ts
```diff
@@ -0,0 +1,91 @@
...
+
+export async function getAgentConfigAction() {
+  if (!isAgentChatEnabled()) return { enabled: false as const };
+
+  const result = await getGetAgentConfigInteractor().invoke();
+  return { enabled: true as const, config: result.ok ? result.data : null };
+}
+
+export async function getAgentConversationAction(conversationId: string, before?: string | null) {
+  if (!isAgentChatEnabled()) return null;
```
benjiwagner: Shouldn't these checks instead be within the interactors, where we gate for the entitlements usually? This is the same for all the other actions.

env.ts
```diff
@@ -20,6 +22,10 @@ export const env = {
   CI: process.env.CI,
 
   APP_MODE: resolveAppMode(process.env),
+  // Kill switch for the hosted in-app Assistant. On by default wherever it is supported, so
+  // forgetting it can never silently ship no assistant. Reached for under pressure, so it accepts
+  // any of false, 0, no or off in any case rather than only one exact spelling.
+  AGENT_CHAT_ENABLED: !AGENT_CHAT_OFF_VALUES.has((process.env.AGENT_CHAT_ENABLED ?? "").trim().toLowerCase()),
```
benjiwagner: Isn't it simpler to instead just use a Boolean value that is gonna be the negative, so disabled, and once we set it, it gets disabled instead. So, for example, on VSL, we don't need to add this flag by default, and only have to add the flag if we want to disable it.

.env.selfhost.template
```diff
@@ -17,3 +17,6 @@ RESEND_OPERATOR_EMAIL=mail@your-company.com
 # GOOGLE_CLIENT_SECRET=
 # AZURE_AD_CLIENT_ID=
 # AZURE_AD_CLIENT_SECRET=
+
+# The hosted in-app Assistant is cloud only and stays off here.
```
benjiwagner: Remove the comments

.env.cloud.template
```diff
@@ -37,3 +37,8 @@ LEMONSQUEEZY_VARIANT_ID_BUSINESS="XXX"
 UNIPILE_API_KEY="XXX"
 UNIPILE_WEBHOOK_SECRET="XXX"
 UNIPILE_HOSTED_AUTH_DOMAIN="auth.customermates.com"
+
+# Hosted in-app Assistant. Runs whenever APP_MODE is cloud; everything else about it is a constant.
```
benjiwagner: Remove the comments

app/components/agent-chat/agent-form-ops.ts
```diff
@@ -0,0 +1,111 @@
+import type { BaseFormStore } from "@/core/base/base-form.store";
+import type { RootStore } from "@/core/stores/root.store";
+
+export type AgentFormStore = BaseFormStore<Record<string, unknown>> & {
+  customColumns?: { id: string; label: string; type: string; options?: unknown }[];
+};
+
+export function resolveFormStore(root: RootStore, form: string): AgentFormStore | null {
```
benjiwagner: I thought the UI is gonna be used by the AI agent by using the IDs and knowing about the page structure and not by actually doing something with the forms. Would my approach that I just recommended be easier and also more maintainable? Also, because the page structure is already described in the docs, so the AI agent is going to be aware of the docs and can search for them and discover on their own which application or functionality is available.

app/components/agent-chat/actions.ts
benjiwagner: Moved into the interactors. There is now a RequiresAgentChat guard in features/agent-chat/agent-availability.ts, applied to all nine agent interactors, so no caller can reach the assistant on a deployment that does not offer it. The actions no longer repeat the check; getAgentConfigAction keeps the single graceful one, because that action exists to answer "is the assistant available", and the panel needs enabled:false rather than an error to stay hidden. The protected layout and the streaming route now call the same isAgentChatAvailable helper.

env.ts
benjiwagner: Inverted. AGENT_CHAT_ENABLED is gone; the flag is now AGENT_CHAT_DISABLED and a cloud deployment runs the assistant without setting anything. Only an explicit opt-out turns it off, so nothing has to be added on Vercel by default. The value parsing also goes away with it, since presence is enough for a disable switch.

.env.selfhost.template
benjiwagner: Removed.

.env.cloud.template
benjiwagner: Removed.

app/components/agent-chat/agent-form-ops.ts
benjiwagner: Fair challenge, and the ID-and-docs route is what navigate, highlight_element and start_tour already do: they only take documented anchor ids, and Mate discovers the page structure from the docs pages.

Where it stops working is typing. Every field in the app is a controlled MobX input, so a synthetic DOM event does not change state: React ignores the value unless you go through the native setter, and comboboxes, single-selects and date pickers need bespoke event sequences each. That path is the fragile one to maintain, and it breaks silently whenever a field switches component.

fill_form is not a second addressing scheme though. The DOM input id IS the form-store path, so it uses the same ids the docs describe, and just applies them through the store instead of synthesizing keystrokes. That is also what makes the dirty-form refusal reliable: the store knows whether the user has unsaved edits, the DOM does not.

Happy to drop it if you would rather keep the surface smaller. It is about 110 lines in agent-form-ops.ts plus the fillForm method, and removing it costs the "fill this in for me" behaviour but leaves navigation, tours and view control untouched. Say the word and I will take it out.

benjiwagner: Done, removed. fill_form, its schema, the client resolver and the mounted-form accessor it needed are gone, and the assistant no longer touches form stores at all.

Interface work now runs the way tours do: Mate opens the right form and points at the fields with highlight_element or start_tour, and the user types and saves. Verified live: asking it to start a new contact with a first and last name now opens the form and answers "The new contact form is open. Please enter: First name Nora, Last name Vogel. Nothing has been saved yet."

configure_view stays, per your call on the scope: layout, grouping, sorting and filters are view state rather than user data, so there is nothing to overwrite and no controlled-input problem.

Net 416 lines deleted. Docs in both languages and the system prompt updated to match.

app/[locale]/(protected)/layout.tsx
```diff
@@ -1,102 +1,9 @@
...
-          <TimelineDetailModal />
-
-          <WebhookDeliveryModal />
-
-          <WebhookModal />
-        </>
-      ) : null}
-    </>
-  );
+  return <ProtectedShell agentChatEnabled={isAgentChatAvailable()}>{children}</ProtectedShell>;
```
Comment (benjiwagner) [inline app/[locale]/(protected)/layout.tsx]: isAgentChatAvailable: Why isn't this variable on the root store, like for other global things?

app/components/agent-chat/actions.ts
```diff
@@ -0,0 +1,58 @@
...
+import { isAgentChatAvailable } from "@/features/agent-chat/agent-availability";
+
+import type { RespondToApprovalData } from "@/features/agent-chat/respond-to-approval.interactor";
+import type { RespondToUiCommandData } from "@/features/agent-chat/respond-to-ui-command.interactor";
+import type { ArchiveAgentConversationData } from "@/features/agent-chat/archive-agent-conversation.interactor";
+import type { ListAgentConversationsData } from "@/features/agent-chat/agent-history";
+import type { DeleteAgentConversationData } from "@/features/agent-chat/delete-agent-conversation.interactor";
+
+export async function getAgentConfigAction() {
+  if (!isAgentChatAvailable()) return { enabled: false as const };
```
benjiwagner: move this to the interactor like we did with the other interactors

features/agent-chat/list-agent-conversations.interactor.ts
```diff
@@ -0,0 +1,40 @@
...
+@TenantInteractor()
+export class ListAgentConversationsInteractor extends AuthenticatedInteractor<
+  ListAgentConversationsData,
+  AgentConversationHistoryResult
+> {
+  constructor(private repo: PrismaAgentChatRepo) {
+    super();
+  }
+
+  @RequiresAgentChat
```
benjiwagner: Why do we have this new decorator which requires agent chat? Isn't this something that we make part of the entitlements, so we don't need a separate decorator?

prisma/migrations/20260819120000_drop_agent_workspace_setup/migration.sql
```diff
@@ -0,0 +1,17 @@
+-- DropTable
+DROP TABLE "AgentWorkspaceSetupResource";
```
benjiwagner: Why do we drop things here? We should only need one migration file. Due to anything that was adapting while working on this feature, we should not rely on backwards compatibility within the change set. So everything that is purely for backwards compatibility within this change set, for example the second migration file, should be merged and simplified so we only have one migration file.

env.ts
```diff
@@ -20,6 +20,7 @@ export const env = {
   CI: process.env.CI,
 
   APP_MODE: resolveAppMode(process.env),
+  AGENT_CHAT_DISABLED: Boolean(process.env.AGENT_CHAT_DISABLED),
```
benjiwagner, line 25: Is there a specific reason why we have not added the OpenAPI key also in the env.ts file and used it via this?

features/user/prisma-user.repository.ts
```diff
@@ -283,6 +293,7 @@ export class PrismaUserRepo
         avatarUrl: args.avatarUrl,
         country: args.country,
         roleId: args.roleId,
+        agentCreditActivatedAt: activatedAt,
```
benjiwagner: Why do we update the agentCreditActivatedAt in the details here? Can you also explain the reasoning behind the timestamp?

features/support/prisma-support.repository.ts
```diff
@@ -1,18 +1,66 @@
-import type { SupportTicketSource } from "@/generated/prisma";
+import { SupportTicketSource } from "@/generated/prisma";
 
 import { BaseRepository } from "@/core/base/base-repository";
 
 export class PrismaSupportRepo extends BaseRepository {
```
benjiwagner: So regarding the support ticket functionality, we changed the features that we want. So if a user asks for any specific support, the idea is that we send this as a feedback to the mail at customermates.com domain, like we already do for the feedback button in the sidebar. It should basically be the same endpoint and give also the stack trace of the chat in the details, so that we don't need the Prisma support repository functionality anymore and also don't need the models.

features/agent-chat/send-agent-message.interactor.ts
```diff
@@ -0,0 +1,282 @@
...
+    private repo: PrismaAgentChatRepo,
+    private usageService: AgentUsageService,
+  ) {
+    super();
+  }
+
+  @RequiresAgentChat
+  @Write({ input: SendAgentMessageSchema })
+  async invoke(data: SendAgentMessageData): Validated<SendAgentMessageResult> {
+    const user = getTenantUser();
```
benjiwagner: So the interactors usually can follow the user accessor on the authenticated interactor, so we don't need to call the getTenantUser. Double check also the other interactors in this change set. Make sure that all the interactors we are using functionally and structurally are aligned with the other interactors or general patterns. Don't invent new patterns and be critical about all the interactors you have created or touched.

core/errors/app-errors.ts
```diff
@@ -40,6 +40,18 @@ export class DemoModeError extends AppError {
   }
 }
 
+export class AgentLimitExceededError extends AppError {
```
benjiwagner: Why have we added those two app error? The app errors are currently only for specified claims that we handle explicitly different. Everywhere else in the interactors, we usually just simply throw a normal error. So please do the same for the agent limit exceeded as well as for the agent session unavailable error. Although, for example, if the agent limit exceeded, the front end should not even allow a limit exceeded error. So it should guard against that and display this gracefully in the user interface even before it happens.

app/components/agent-chat/actions.ts
benjiwagner: Implemented in f32e31da. getAgentConfigAction is now a thin adapter over the interactor and hand-serializes the denial code so the store can distinguish a genuine availability denial from a retryable validation failure. The tautological enabled literal was removed. The store latches off only for the explicit unavailable code and retains its retry behavior for other failures. Focused action, availability, and store tests pass.

features/agent-chat/list-agent-conversations.interactor.ts
benjiwagner: Implemented in f32e31da. agentChat is now an entitlement on all four plans, while the demo exclusion is scoped specifically to agentChat so messaging and sharedAccounts retain their prior behavior. Self-hosted entitlement behavior is unchanged. The operational kill switch is enforced on every exported assistant action and at the messages route boundary. Browser acceptance confirms the demo launcher remains visible but cannot mount or shortcut-open the panel.

app/[locale]/(protected)/layout.tsx
Comment (benjiwagner) [reply app/[locale]/(protected)/layout.tsx]: Implemented in d45e087c. The root server layout now computes assistant availability once and passes agentChatEnabled through Providers into RootStore beside appMode. The protected layout is back to its client shape, the branch-only force-dynamic wrapper is gone, and both mounting and Cmd+J use the root flag while AgentChatStore.enabled remains nullable for the config retry loop. The production build confirms protected routes remain dynamic. Local browser acceptance also passed for the cloud sidebar and Cmd+J, while demo keeps Ask AI visible and shows the unavailable notice.

features/support/prisma-support.repository.ts
benjiwagner: Implemented in 1cf68afd with one important correction to the original premise: SupportTicket already shipped on main in 20260722085017_support_ticketing, so this is an explicitly approved destructive forward migration, not cleanup of branch-only schema. The new migration drops the table, both enum types, and any existing rows. A shared strict FeedbackCreator now backs sidebar feedback plus MCP and chat support email. Chat support includes a sanitized recent transcript, and the UI reports success only after the email provider accepts delivery; provider rejection remains an error. The intentional losses are no ticket number, persisted lifecycle, in-app reply thread, idempotency, or failed-attempt record, and replay can send a duplicate email. Fresh migration replay, empty schema diff, strict delivery tests, all locale copy, and en/de docs pass.

features/user/prisma-user.repository.ts
benjiwagner: Implemented in 5837a93e. agentCreditActivatedAt now changes inside AdminUpdateUserDetailsInteractor beside the user activation transition and subscription quantity update. The timestamp marks when an assistant seat became billable, which is the anchor needed for correct mid-cycle credit proration. Named tenant-scoped repository methods keep the user transition, billing quantity, and credit anchor in the same transaction so those states cannot drift. Focused interactor, repository, and tenant-write convention tests pass.

core/errors/app-errors.ts
benjiwagner: Implemented in 5c0ec87e. AgentSessionUnavailableError was removed and its call sites now use normal errors. AgentLimitExceededError remains only for the 429 contract because the agent chat store explicitly branches on HTTP 429 to show the localized limit-reached toast; turning it into a generic 500 would regress that user path. The exception no longer exposes blockedReason or reset timing, and submitDraft now has the missing store-level blockedReason guard. Route and store tests cover the 429 and blocked paths.

env.ts
benjiwagner: Implemented in fbe0d78c. OPENAI_API_KEY is optional at global environment parsing and is passed explicitly into the OpenAI client at the hosted assistant boundary. Missing credentials therefore fail only when the cloud-only assistant is instantiated, without breaking self-hosted or assistant-disabled builds. Environment and model-boundary tests pass.

prisma/migrations/20260819120000_drop_agent_workspace_setup/migration.sql
benjiwagner: Implemented in 1884656f. The applied folder name 20260817120000_agent_assistant is unchanged and now creates the final assistant schema directly. The two branch-only cleanup migrations were deleted, including the temporary workspace setup tables and AgentMessage.searchText. The later 20260820120000_drop_support_ticketing migration remains intact because SupportTicket was already shipped on main. A fresh disposable database applied all 34 migrations, prisma migrate diff reported no difference, and the migration database suite passed. If the isolated preview recorded the deleted branch-only names, only those two metadata rows may need removal.

features/agent-chat/send-agent-message.interactor.ts
benjiwagner: Implemented in a188ed77 and 1e8dea85. The interactors now use inherited user, userId, and companyId accessors consistently, const output shapes are explicit, and demo decorators are aligned. The transaction bug received its own fix: the outer write runs with tx disabled, phase one atomically claims the lease and reserves credit, phase two re-fences the exact live lease and reserved pre-provider usage before atomically admitting all chat state, and cleanup is one commit-ambiguity-safe transaction. The real PostgreSQL regression proves rollback of partial admission, a durable zero-credit released ledger row, no lease or poisoned turn, and successful retry with the same client request id.

app/[locale]/(protected)/company/components/feedback/feedback-modal.store.ts
```diff
@@ -31,6 +31,8 @@ export class FeedbackModalStore extends BaseModalStore<SendFeedbackData> {
         this.onInitOrRefresh({ feedback: "", type: this.form.type });
         this.toastSuccess("feedback.success");
       } else this.setError(res.error);
+    } catch {
+      this.toastError("Common.notifications.unexpectedError");
```
Comment (benjiwagner) [inline app/[locale]/(protected)/company/components/feedback/feedback-modal.store.ts]: Just to double check, do we usually do it like this, that we try to catch the error and throw it manually with this toast error? Or is there a global behavior for it?

app/components/agent-chat/actions.ts
```diff
@@ -0,0 +1,77 @@
...
+    };
+  }
+
+  const result = await getGetAgentConfigInteractor().invoke();
+  if (result.ok) return result;
+  return { ok: false as const, error: z.treeifyError(result.error), code: result.code };
+}
+
+export async function getAgentConversationAction(conversationId: string, before?: string | null) {
+  requireAgentChatEnabled();
```
benjiwagner: Please double check all the usages in this file. I was expecting the actions file to just call the interactors and not have other usages like the require agent chat enabled because this should instead be part of the interactors logic.

app/components/agent-chat/agent-view-ops.ts
```diff
@@ -0,0 +1,115 @@
+import { FilterOperatorKey } from "@/core/base/base-query-builder";
+
+import type { Filter, SortDescriptor } from "@/core/base/base-get.schema";
+import type { BaseDataViewStore, HasId } from "@/core/base/base-data-view.store";
+import type { RootStore } from "@/core/stores/root.store";
+import type { ConfigureViewData } from "@/ee/agent-chat/ui-operations";
+
+export type AgentDataViewStore = BaseDataViewStore<HasId>;
+
+export function resolveDataViewStore(root: RootStore, view: string): AgentDataViewStore | null {
```
benjiwagner: Why do we still have the stores here?

app/components/agent-chat/actions.ts
benjiwagner: Rechecked the whole file. `actions.ts` is now 63 lines and every export is a single pass-through of the shape `serializeResult(getXInteractor().invoke(data))` — no availability check, no env read, no store access. The entitlement gate lives where you said it belongs: each interactor opens with `this.entitlements.require("agentChat")`, so a caller that bypasses the action still cannot get past it. The kill switch is separate and enforced at the HTTP entry points rather than per action.

app/components/agent-chat/agent-view-ops.ts
benjiwagner: Gone — `agent-view-ops.ts` no longer exists. What remains under `app/components/agent-chat/` is the panel component, the two stores it needs for presentation (chat and UI control), the activity grouping hook, the tour overlay, and the actions file. Nothing in that directory reaches into a store from a server action any more.

app/[locale]/(protected)/company/components/feedback/feedback-modal.store.ts
Comment (benjiwagner) [reply app/[locale]/(protected)/company/components/feedback/feedback-modal.store.ts]: There is a global behaviour and this no longer bypasses it. `runUserAction` wraps a gesture and routes anything thrown to `reportApplicationError`, so a store does not need its own catch. The submit path here now has no `catch` at all: it uses `try/finally` purely to clear the loading flag, and branches on the structured interactor result — `res.ok` closes and toasts success, otherwise `setError(res.error)` renders the validation error inline on the form. A manual error toast would have been the wrong shape for a form field error anyway.

app/components/agent-chat/actions.ts
```diff
@@ -0,0 +1,63 @@
...
+import type { RespondToApprovalData } from "@/ee/agent-chat/respond-to-approval.interactor";
+import type { RespondToUiCommandData } from "@/ee/agent-chat/respond-to-ui-command.interactor";
+import type { ArchiveAgentConversationData } from "@/ee/agent-chat/archive-agent-conversation.interactor";
+import type { ListAgentConversationsData } from "@/ee/agent-chat/agent-history";
+import type { DeleteAgentConversationData } from "@/ee/agent-chat/delete-agent-conversation.interactor";
+
+export async function getAgentConfigAction() {
+  const result = await getGetAgentConfigInteractor().invoke();
+  if (result.ok) return result;
+  return { ok: false as const, error: z.treeifyError(result.error), code: result.code };
```
benjiwagner: shouldnt this be moved to the interactor?

app/components/agent-chat/ui-control.store.ts
```diff
@@ -0,0 +1,326 @@
...
+      };
+    }
+    if (!this.navigateCallback) return { ok: false, result: "Navigation is not available right now." };
+
+    const outcome = await this.navigateCallback(target.route);
+    if (outcome === "navigated") return { ok: true, result: `Navigated to ${target.route}.` };
+    if (outcome === "blocked") {
+      return {
+        ok: false,
+        result: "Navigation requires the user to resolve unsaved changes.",
```
benjiwagner, line 124: are we sure this does not need to be translated

.env.cloud.template
```diff
@@ -37,3 +37,13 @@ LEMONSQUEEZY_VARIANT_ID_BUSINESS="XXX"
...
 UNIPILE_WEBHOOK_SECRET="XXX"
 UNIPILE_HOSTED_AUTH_DOMAIN="auth.customermates.com"
+
+# The workflow world's Postgres pool must be at least its queue concurrency (50).
+# Leaving it at the pg default of 10 makes durable steps wait for a free connection,
+# which surfaces as agent turns hanging for tens of minutes under load.
+WORKFLOW_POSTGRES_MAX_POOL_SIZE="50"
+
+AI_GATEWAY_API_KEY="XXX"
+# AGENT_MODEL_KEYS="fast,balanced"
```
benjiwagner: AGENT_MODEL_KEYS and AGENT_MODEL_DEFAULT are not needed IMO

ee/agent-chat/cancel-agent-turn.interactor.ts
```diff
@@ -0,0 +1,54 @@
...
+import { agentUiCommandHookToken } from "./agent-ui-command";
+import type { BackgroundTaskService } from "@/core/utils/background-task.service";
+
+export const CancelAgentTurnSchema = z.object({ conversationId: z.uuid() }).strict();
+
+export type CancelAgentTurnData = Data<typeof CancelAgentTurnSchema>;
+
+const OutputSchema = z.object({ cancelling: z.boolean() });
+
+@AllowInDemoMode
```
benjiwagner: Double check to change that. We only want to allow the reading of the previous and current chats. We don't want any other actions in the demo mode. The seat user Max Bergmann should also have a couple of archived chats, as well as a current chat in the demo mode. We always want the chat to be opened by default, and it should display beautifully answered question about the data of the CRM, where we also display, like, the CS Power export or visualization in the tabular format. Something that is easy to understand for Users

ee/agent-chat/delete-agent-conversation.interactor.ts
```diff
@@ -0,0 +1,46 @@
...
+    input: DeleteAgentConversationSchema,
+    output: DeleteAgentConversationResultSchema,
+    tx: false,
+  })
+  async invoke(data: DeleteAgentConversationData): Validated<{ deleted: true }> {
+    const denied = await this.entitlements.require("agentChat");
+    if (denied) return denied;
+
+    const deleted = await this.repo.deleteArchivedConversation(data.conversationId);
+    if (!deleted) return createInteractorFailure(CustomErrorCode.agentConversationNotFound, ["conversationId"]);
```
benjiwagner: Why exactly have we invented a new pattern, create interactor failure?

ee/agent-chat/get-agent-conversation.interactor.ts
```diff
@@ -0,0 +1,86 @@
...
+import type { EntitlementService } from "@/ee/subscription/entitlement.service";
+
+import type { PrismaAgentChatRepo } from "./prisma-agent-chat.repository";
+import { clientSafeAgentMessageParts } from "./agent-chat.schema";
+import { sanitizeAgentConversationTitle } from "./agent-output-safety";
+import { AgentMessagePageSchema, type AgentMessagePageData } from "./agent-history";
+import { createInteractorFailure } from "@/core/validation/interactor-failure-server";
+import { CustomErrorCode } from "@/core/validation/validation.types";
+
+export const GetAgentConversationSchema = AgentMessagePageSchema;
```
benjiwagner: Double check all the interactors. We usually name the input schema only schema and also only export it if it's actually used, exported anywhere. Else rename it to schema, keep it local. Double check the other interactors as well.

ee/subscription/refresh-subscription.interactor.ts
```diff
@@ -24,7 +25,7 @@ export class RefreshSubscriptionInteractor extends AuthenticatedInteractor<void,
     const subscription = await this.repo.getSubscriptionOrThrow();
 
     if (subscription.plan === SubscriptionPlan.enterprise) return { ok: true as const, data: null };
-    if (!subscription.lemonSqueezyId) throw new Error("Subscription does not have a LemonSqueezy ID");
+    assertInvariant(subscription.lemonSqueezyId, "Subscription does not have a LemonSqueezy ID");
```
benjiwagner: Like in the previous comments, why have we created a new wrapper method with, for example, assert invariant? I think this makes it harder to read instead of plainly throwing an error. Is there a specific reasoning behind that? Is it really needed as a new pattern?

core/errors/app-errors.ts
```diff
@@ -80,19 +110,85 @@ export class UnmappableWebhookPayloadError extends Error {
...
   const message = (err as { message?: unknown } | null | undefined)?.message;
   if (typeof message !== "string") return false;
 
   return message.includes(WEBHOOK_FAILURE_MESSAGE_PREFIX) || message.startsWith(DEMO_MODE_MESSAGE);
 }
 
-export function appErrorResponse(err: unknown): { message: string; statusCode: number } | null {
+const ERROR_CAUSE_DEPTH_LIMIT = 8;
+
+function errorCause(error: unknown): unknown {
```
benjiwagner, line 131: The changes in the app arrows look pretty complicated and also I don't quite get the reasoning. Can you explain this deeply to me and whether we actually need that?

core/validation/interactor-failure-server.ts
```diff
@@ -0,0 +1,14 @@
+import { getTranslations } from "next-intl/server";
+
+import type { CustomErrorCode } from "./validation.types";
+import { createZodError, type InteractorFailure } from "./validation.utils";
+
+export async function createInteractorFailure(
```
benjiwagner: So, as far as I understand it, we have created this wrapper function to gracefully throw within interactors instead of plainly throwing new errors. But the core idea was that the pre-check is gracefully checking for errors. Everything else is considered a failure in the business logic, which we don't want. So generally speaking, a new error thrown is something where we want to be informed by Sentry and don't want to gracefully return something to the user aside from, Hey, 500 bad request, this does not work. This was what we thought about it. So it's super easy. If something is not expected in the business logic of an interactor and within the invoke method, we can simply throw an error, which we would not expect to happen. Therefore, by now creating this interactor failure, it sort of looks like that we are re-validating things that should not happen, which could be quite over-engineered, to be honest.

ee/agent-chat/agent-output-safety.ts
```diff
@@ -0,0 +1,330 @@
+const INTERNAL_REFERENCE = "[internal reference]";
```
benjiwagner, line 1: All the output safety stuff and display things for our custom-built chat in the front end seem to be quite extensive. Are we completely certain that there isn't anything pre-built that fits our framework and stack, visually speaking, so we can simplify that quite a bit, so we don't need to reinvent everything ourselves?

ee/agent-chat/agent-translator.ts
```diff
@@ -0,0 +1,19 @@
...
+import es from "@/i18n/locales/es.json";
+import fr from "@/i18n/locales/fr.json";
+import it from "@/i18n/locales/it.json";
+import { appLocaleOrDefault } from "@/i18n/locale-registry";
+
+export type AgentTranslator = (key: string, values?: Record<string, string | number>) => string;
+
+const CATALOGS = { de, en, es, fr, it };
+
+export function agentTranslator(locale: string): AgentTranslator {
```
benjiwagner: Why do we need to create a separate agent translator? Why can't we use the normal translation way that we have everywhere else? This seems to be like an duplication of existing patterns

app/components/agent-chat/actions.ts
benjiwagner: Agreed and fixed in 19b6d34. The action was hand-rolled because the interactor smuggled an entitlement code through a custom result type. A disabled assistant is now data, not an error: the interactor returns `{ enabled: false }` through the standard `Validated<>` contract, the action is a plain `serializeResult(...)`, and the store maps `data.enabled === false` to the disabled state and any failure to retry.

app/components/agent-chat/ui-control.store.ts
benjiwagner: Deliberately untranslated: these strings are tool results consumed by the model, never rendered to the user. The whole tool surface speaks English to the model (all server tool results do the same), and the model answers the user in their locale. Translating them would localize the model's input, not the UI.

.env.cloud.template
benjiwagner: Agreed and went further in 19b6d34: both env vars are deleted entirely (env.ts, template, catalog plumbing). `SHIPPED_AGENT_MODEL_KEY` in the catalog is now the single configuration point; changing models is a one-line code change.

ee/agent-chat/cancel-agent-turn.interactor.ts
benjiwagner: Built in 19b6d34. Demo mode is now a read-only showroom: availability and entitlement admit demo, the eight mutating interactors (send, cancel, both responders, archive, delete, restore, support ticket) lost their demo allowance so the tenant decorator blocks them, and the reads stay open. The seed gives Max one active conversation (an opportunities question answered with a markdown table plus an insight) and three archived chats (duplicates, task triage, pipeline health). The panel opens by default in demo, and a blocked send unwinds cleanly: draft restored, no phantom message, the standard demo toast with the signup link.

ee/agent-chat/delete-agent-conversation.interactor.ts
benjiwagner: It is the graceful-failure factory for outcomes a consumer branches on, using the vocabulary the codebase already had: a zod issue carrying params.error with a CustomErrorCode, resolved against the same Common.errors catalog that zod-error-map-server already resolves for schema parsing. Not a parallel system, the same lane made available after parsing. The deeper reasoning is on the interactor-failure-server thread; the naming half of this comment (schemas) is fixed in 19b6d34.

ee/agent-chat/get-agent-conversation.interactor.ts
benjiwagner: Agreed and fixed in 19b6d34: every interactor schema nobody imports is now a local `Schema`, un-exported. One exception kept deliberately: `ArchiveAgentConversationSchema` stays exported because the restore interactor genuinely shares it, which is exactly the rule you stated.

ee/subscription/refresh-subscription.interactor.ts
benjiwagner: Agreed and fixed in 19b6d34: assertInvariant is gone, all six sites are plain `if (!x) throw new Error(...)` (an if-throw narrows types identically, so the helper bought nothing). The honest reason it existed: the interactor error-contract test forbade throw statements inside tenant interactors, and a call expression slipped past it. The test now exempts plain `new Error` throws, which are invariants by definition, while still forbidding thrown access errors and business failures, which matches your philosophy: invariants throw and reach Sentry, expected outcomes are returned.

core/errors/app-errors.ts
benjiwagner: The complexity has one concrete cause: the compiled workflow bundle under .well-known/workflow contains a second copy of this module, so `instanceof` lies across that boundary (you can see the duplicated functions inside the built route.js). Main had already solved this for the base class with the Symbol.for brand; this branch extended the same trick one level down because the agent tool boundary must know which error occurred (permission vs auth vs demo) to hand the model a structured failure instead of killing the turn. The cause-chain walking exists because runInTransaction wraps errors. Each remaining piece is pinned by a test that names its scenario (foreign bundle copy, dev-HMR module mix, refusing to Sentry-suppress malformed branded objects); the one export nothing called is deleted in 19b6d34.

core/validation/interactor-failure-server.ts
benjiwagner: The layering, in full. Lane 1, validation: schema refinements add issues carrying params.error codes; messages resolve via zod-error-map-server against Common.errors, client toasts them. Lane 2, thrown errors: everything thrown propagates to handleError, which was already on main mapping AppErrors (401/403) and five known Prisma codes (P2025 to 404, P2002 to 409, ...) to clean HTTP without Sentry; everything else rethrows to Sentry as a 500. So even on main the rule was never 'all throws are incidents' but 'throws are incidents unless the boundary recognizes them'. Lane 3, what the agent added: expected business outcomes discovered mid-transaction (lease already claimed, approval expired, credits exhausted mid-admission) that physically cannot live in a precheck and that the client branches on: 409 drives the rejoin flow, 429 the credits toast, approval-unavailable its own state. Throwing those would 500-and-Sentry a user closing a laptop. createInteractorFailure builds them with the same code vocabulary and the same Common.errors catalog Lane 1 already uses; it lives in a separate file only because validation.utils is client-imported and message resolution needs next-intl/server. Two properties keep it honest: a graceful failure returned inside a transaction still rolls it back, and the model boundary gets the same treatment (a mid-tool ForbiddenError or P2025 becomes structured data the model adapts to instead of a dead turn). The working rule, now also enforced by the amended contract test: return a failure only when a consumer branches on it; throw everything else.

ee/agent-chat/agent-output-safety.ts
benjiwagner: The display side already is prebuilt: rendering goes through the vendored ai-elements MessageResponse (markdown, tables, code). This file is not display, it is a leak-redaction boundary applied to model-visible text before it reaches the browser: UUIDs, secrets and JWTs, cookie and authorization headers, model identifiers, token counts and internal costs, leaked reasoning blocks, and our page_context markup. Nothing prebuilt can do this because it encodes our private vocabulary, and it is what the agent-client-payload-boundary guarantee rests on at the text level. It is fully covered by its own test suite, including the streaming sanitizer's chunk-boundary cases.

ee/agent-chat/agent-translator.ts
benjiwagner: Agreed and fixed in 19b6d34: agent-translator is deleted. It duplicated i18n/get-translator (the house background translator the lifecycle emails already use) and was strictly worse, statically bundling all five catalogs where the house one imports per locale. The workflow's runner messages now use getTranslator directly.

core/validation/interactor-failure-server.ts
benjiwagner: Follow-up from our discussion, landed in a02e6f1b: the factory became kind-named constructors (failNotFound, failConflict, failRateLimit, failUnavailable, failAuthorization, plus neutral fail), so the HTTP class is visible where the outcome is born and misclassification is structurally impossible. Every hand-rolled site across messaging and the other features migrated onto them, including the send-chat-message draft guard whose forgotten code had misclassified it. The KB error model documents the pattern in knowledge-base#89.

PR #83, opened 2026-08-18
fix: gate the registration database suite and give the checkout interactor its accessor

benjiwagner:
Closes [CUS-244](https://linear.app/customermates/issue/CUS-244/gate-the-registration-database-test-behind-run-database-tests).

## Summary

`features/user/__tests__/registration-real-db.test.ts` read `process.env.DATABASE_URL` directly and ran unconditionally, unlike every other database-backed suite. Since `vitest.config.ts` injects only `APP_MODE` and `BASE_URL` and never loads `.env`, a plain `yarn test` on a machine without PostgreSQL on the default port failed five tests with `connect ECONNREFUSED 127.0.0.1:5432`.

It now uses the shared `getLocalDatabaseTestUrl()` helper and the `describeDatabase` pattern, and is renamed to the `*.database.test.ts` convention the other gated suites follow.

## Context

Found while running the suite for [CUS-219](https://linear.app/customermates/issue/CUS-219/make-the-current-users-assigned-user-modal-fully-read-only): five failures in a file that change had not touched, initially indistinguishable from a regression in it.

**CI is unaffected.** `.github/workflows/test.yml:17-18` already sets `DATABASE_URL` and `RUN_DATABASE_TESTS: "true"`, so the suite runs in CI exactly as before. This only changes what happens on a machine that has not opted in.

## Validation

The risk with a change like this is trading a false failure for lost coverage, so I checked all three states rather than just the quiet one:

| State | Result |
| --- | --- |
| No database, no flag | `yarn test`: **306 files passed, 6 skipped, 0 failures** (previously 5 failures) |
| `RUN_DATABASE_TESTS=true` + local database | Suite **executes, all 5 tests pass** |
| `RUN_DATABASE_TESTS=true`, database unreachable | Fails loudly with `ECONNREFUSED`, which is correct |

`yarn typecheck`, `yarn lint:check`, `yarn conventions:check` (300 tests) and `yarn build` all pass.

## Impact and rollback

One test file: four lines added and a rename. No production code, no CI configuration, no schema change. The suite is skipped rather than deleted, and still runs wherever it ran before. Revert the single commit to roll back.

### DIFF
--- a/ee/subscription/create-checkout-session.interactor.ts
+++ b/ee/subscription/create-checkout-session.interactor.ts
@@ -10,8 +10,8 @@ import type { Subscription } from "@/generated/prisma";
 import type { CountActiveUsersRepo } from "@/features/user/count-active-users.repo";
 
 import { TenantInteractor } from "@/core/decorators/tenant-interactor.decorator";
+import { UserAccessor } from "@/core/base/user-accessor";
 import { Validate } from "@/core/decorators/validate.decorator";
-import { getTenantUser } from "@/core/decorators/tenant-context";
 import { resolveRequestOrigin } from "@/core/config/environment";
 import { redirectTo } from "@/features/auth/auth-outcome";
 import { env } from "@/env";
@@ -28,12 +28,14 @@ export abstract class CreateCheckoutCompanyRepo {
 }
 
 @TenantInteractor({ resource: Resource.company, action: Action.update })
-export class CreateCheckoutSessionInteractor {
+export class CreateCheckoutSessionInteractor extends UserAccessor {
   constructor(
     private lemonSqueezyService: SubscriptionService,
     private repo: CreateCheckoutCompanyRepo,
     private userRepo: CountActiveUsersRepo,
-  ) {}
+  ) {
+    super();
+  }
 
   @Validate(Schema)
   async invoke(data: CreateCheckoutSessionData): Promise<Redirect | { ok: false; error: z.ZodError }> {
@@ -52,7 +54,7 @@ export class CreateCheckoutSessionInteractor {
       offer,
       quantity: activeUsersCount,
       custom: {
-        company_id: getTenantUser().companyId,
+        company_id: this.companyId,
       },
       redirectUrl,
     });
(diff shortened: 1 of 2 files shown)


benjiwagner:
### Added: the checkout interactor now inherits the accessor (`71bb800f`)

Scope added on request, so the PR title is updated. CUS-244's stated non-goal "no production code change" is superseded deliberately.

**Why this belongs to the accessor work.** Cross-tabulating base class against decorator across every `*Interactor` shows the convention is not "most interactors extend something", it is conditional on tenant scope:

| Base class | Decorator | Count |
| --- | --- | --- |
| `AuthenticatedInteractor` | `@TenantInteractor` | 115 |
| `BaseGetInteractor` | `@TenantInteractor` | 14 |
| `BaseGetConfigurationInteractor` | `@TenantInteractor` | 5 |
| `UserAccessor` | `@TenantInteractor` | 3 |
| **(none)** | **`@TenantInteractor`** | **1** |
| (none) | `@SystemInteractor` | 50 |
| (none) | (none) | 24 |

**137 of 138 tenant-scoped interactors extend something.** `CreateCheckoutSessionInteractor` was the one exception, which is exactly why it was the last caller of the free `getTenantUser`. The 74 that extend nothing are system interactors and undecorated validators/prechecks, none of which has a tenant user to read.

**It extends `UserAccessor`, not `AuthenticatedInteractor`.** `invoke` returns `Promise<Redirect | { ok: false; error: ZodError }>`, and `Validated<T>` is `Promise<{ ok: true; data: T } | { ok: false; error: ZodError }>` — a `Redirect` is `{ redirect: string }` and satisfies neither branch. `ReconnectConnectedAccountInteractor` has that identical signature and resolves it the same way, including the `super()` call in a parameter-property constructor.

**Verification, given this is a payments path.** No provider call was made. The pre-existing `create-checkout-session.interactor.test.ts` already asserts `custom: { company_id: COMPANY_ID }` against the mock user's company, so it covers the accessor directly. Because a vacuous test would prove nothing, I mutation-checked it: swapping `this.companyId` for `this.userId` fails the assertion with

```
-  "company_id": "00000000-0000-4000-8000-000000000001",
+  "company_id": "test-user-id",
```

and reverting restores 4/4. Behaviour is otherwise identical, since `UserAccessor.companyId` reads `this.user`, which delegates to `getTenantUser` after a tenant-bypass check.

Rebased onto the updated branch after #77 merged; all gates re-run on the new base: typecheck, lint, conventions, 307 files / 2582 tests, build.

PR #87, opened 2026-08-19
fix(messaging): stop probing offset pagination against cursor-only providers

benjiwagner:
## Summary

`paginateStep` inferred a provider's pagination style from whether the first page carried `next_cursor`. Stop inferring, and treat the provider's own rejection as the authoritative answer.

## Context

Sentry `CUSTOMERMATES-3X` (50 events) and `CUSTOMERMATES-4G` (6 events), both on `POST /.well-known/workflow/v1/step`:

> `UnipileRequestError: Unipile v2 request failed: 400 ... "detail":"Google Calendar use cursor for pagination."`

`paginate.ts` decided the style with `if (!mode) mode = result.next_cursor ? "cursor" : "offset"`. A missing `next_cursor` is ambiguous. It means either the provider paginates by offset, or that page was the last one. The code always resolved it to the first reading, so a cursor-only provider whose first page came back exactly full (`>= UNIPILE_MAX_LIMIT`, 25) was classified as offset-paginated. The offset branch then declined to terminate, advanced `offset`, and the next request carried `offset=25` to Google Calendar's cursor-only events endpoint.

Inference cannot be made correct here. The Unipile SDK documents both parameters on the same endpoint as provider-dependent:

> "An offset used for pagination, if supported by the provider, else use `cursor`."

So "offset-provider page 1, exactly full" and "cursor-provider last page, exactly full" are indistinguishable from one response. Rather than guess, keep the existing walk and let the provider answer: if a request in offset mode is rejected specifically because the provider requires cursors, there are no further pages to fetch, so end the walk cleanly instead of reporting the rejection.

This is deliberately narrow. `isUnipileCursorPaginationRequired` matches only a 400 whose error type ends in `/invalid_parameters` and whose body names cursor pagination, so any other bad request still surfaces. Offset-paginated providers are untouched, which matters because terminating on a missing `next_cursor` instead would have silently truncated their backfills after one page.

## Validation

- New `ee/messaging/ingest/backfill/__tests__/paginate.test.ts`, 4 cases: cursor walk, offset walk, the cursor-only rejection, and an unrelated 400 that must still throw.
- Confirmed the test is a real regression test: with the guard removed, "stops cleanly when a cursor-only provider rejects the offset probe" fails and the other three still pass.
- `yarn vitest run ee/messaging` 355 passed / 40 files, `yarn lint` clean, `yarn typecheck` clean.

The `ECONNREFUSED` errors in the local vitest run are the local `.env` pointing at a Postgres container that is not running, not a regression from this change.

## Impact and rollback

Affects the Unipile backfill and resync walk only. The change is additive: one new exported predicate and one catch clause that ends the walk on a specific provider rejection. No behaviour change for any request that succeeds, and no change to offset-paginated providers. Rollback is a straight revert of the commit.

### DIFF
--- a/ee/messaging/ingest/backfill/__tests__/paginate.test.ts
+++ b/ee/messaging/ingest/backfill/__tests__/paginate.test.ts
@@ -1,0 +1,90 @@
+import { describe, it, expect, vi } from "vitest";
+import { MOCK_ENV_MODULE } from "@/tests/helpers/interactor-test-setup";
+
+vi.mock("@/env", () => MOCK_ENV_MODULE);
+
+import { paginateStep, UNIPILE_MAX_LIMIT } from "../paginate";
+import { UnipileRequestError } from "../../../messaging.service";
+
+function page(count: number, nextCursor?: string) {
+  return { data: Array.from({ length: count }, (_, index) => ({ id: `${index}` })), next_cursor: nextCursor };
+}
+
+function cursorPaginationError() {
+  return new UnipileRequestError(
+    400,
+    "provider/invalid_parameters",
+    '{"object":"Error","status":400,"type":"provider/invalid_parameters","title":"Invalid parameters","detail":"Google Calendar use cursor for pagination.","req_id":"req-9jg"}',
+  );
+}
+
+describe("paginateStep", () => {
+  it("follows next_cursor while the provider keeps returning one", async () => {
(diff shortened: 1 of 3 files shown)


benjiwagner:
Superseded by #91, which carries both commits unchanged. These two branches each added a predicate to `ee/messaging/messaging.service.ts` at the same insertion point, so they conflicted with each other and whichever merged second would have gone CONFLICTING (which silently drops the `ci` job). Combining them resolves that as a pure union of the two functions.

PR #93, opened 2026-08-19
feat(data-view): apply and persist shared filter changes automatically

benjiwagner:
## Summary

Shared data-view filter popovers now apply and persist complete filter changes as the user makes them. The Apply action is gone from normal filtering; named presets still require an explicit Save, and that Save no longer doubles as the only way to use a preset.

Closes [CUS-186](https://linear.app/customermates/issue/CUS-186/auto-apply-and-persist-shared-data-view-filters).

- Operator, select, date, date-range and preset changes commit immediately. Text and numeric entry debounces by 300 ms and flushes on blur, overlay dismissal, navigation and preset save.
- Only configurations that pass `hasValidFilterConfiguration` reach a query or persistence, so a half-built filter is never sent.
- Auto-applied refreshes route through the existing `executeRefresh("background")` mode instead of the visible one, so results update without the full-screen loading overlay.
- Selecting a saved preset in the popover dropdown now applies it immediately, matching what the preset chips in the active-filters bar already did.
- Editing a filter while a preset is selected updates the view but leaves the stored preset untouched until Save, so a temporary adjustment to a saved preset is finally usable.

## Context

The popover previously edited `EditFiltersModalStore.form.filters` as a draft and committed to `BaseDataViewStore` only on Apply. Four things recorded on the issue shaped the implementation:

- **Apply was not the only commit path.** The popover body is wrapped in `<AppForm store={modalStore}>`, so Enter submits. `onSubmit` now flushes pending edits and only writes a preset when one is being created or edited.
- **The entity activity timeline is the hard-failure surface.** `ActivitiesStore.fetchPage` runs `ActivityFiltersSchema.parse` (not `safeParse`) over `.strict()` branches, so any mid-interaction shape throws and puts the timeline into refresh-error. Gating every commit on `hasValidFilterConfiguration` is what keeps that from happening; verified live by selecting an operator with no value and seeing no request and no error.
- **Newest-wins and a background mode already existed.** The gap was that `setQueryOptions` routed to the visible mode. `setQueryOptions` now takes an optional `refreshMode`, defaulting to today's behaviour for every existing caller. No second queue was introduced.
- **`Common.filters.apply`** is removed from all five locale catalogues, as the i18n parity gate requires.

Two behaviours were changed deliberately beyond the literal scope, both to avoid regressions the auto-apply path would otherwise introduce:

- **A superseded response could revert a newer edit.** A completed request commits `setItems`, which resets `store.filters` from the server's echoed filters. With rapid edits, request A's response could overwrite filters the user had already changed to C, and the queued re-run would then query A. The background drain now passes the store's existing `shouldCommit` guard so a superseded response is discarded rather than committed. A focused test covers this.
- **Removing a chip while the popover is open** left a stale draft that the next auto-apply would resurrect. `syncDraftFromTable` re-syncs the draft on chip removal, clear-all and preset-chip selection, and only when the popover is open for that store.

`mergeFiltersWithFilterableFields` is also guarded against a stored preset whose `filters` is not a usable array. That path is latent today (the write path validates), but it throws inside a MobX reaction, so the failure is swallowed and preset selection silently does nothing. Malformed entries are now skipped rather than throwing; presets remain repaired-not-dropped, consistent with the direction of [CUS-210](https://linear.app/customermates/issue/CUS-210/saved-filter-presets-are-dropped-instead-of-repaired-when-one-filter-fails-validation).

Selecting "No preset" in the dropdown keeps the current filters rather than clearing them, preserving today's behaviour and keeping Cancel out of preset creation from wiping the view.

## Validation

Gates, all green on this branch: `yarn typecheck`, `yarn lint`, `yarn conventions:check` (41 files, 303 tests), `yarn test` (321 files, 2722 tests), `yarn build`.

Focused tests added: `edit-filters-modal.store.test.ts` (13 fake-timer cases: debounce, flush, incomplete configurations, cleared values, guard re-baselining, preset apply, malformed preset, dismissal flush, submit behaviour), `base-data-view-background-refresh.test.ts` (5 deferred-request cases: no visible state, single-flight coalescing with the newest filters, unchanged-filter no-op, failure toast with prior rows, visible mode unchanged), `filter-popover-actions.test.ts` (3 render cases for the footer in normal, editing and creating modes).

Local browser acceptance against disposable PostgreSQL with seeded synthetic data, English and German, desktop (1280) and mobile (390):

- Popover footer offers only Clear; Save appears only in preset mode. Confirmed in English desktop, German desktop and the German mobile drawer.
- Opening the popover fired zero requests.
- Selecting `contains` with no value fired zero requests; typing then produced results.
- Eight rapid keystrokes produced exactly one request after the pause, with the final value in the URL.
- A select change applied immediately: 30 rows to 4, one request, URL updated, popover stayed open.
- Navigating to a bare `/en/contacts` with no query string restored the auto-applied filters from P13n and rewrote the URL, with Apply never pressed.
- Clear applied immediately with no loading overlay.
- Preset selection from the dropdown applied immediately (30 rows to 4). Editing a filter with that preset selected updated the view while the stored `savedFilterPresets` row stayed byte-identical; pressing Save then persisted both filters.
- Entity activity timeline: `Type in Messages` applied cleanly, no console error, no refresh-error state.
- Navigating away with the popover open and an auto-applied filter produced no unsaved-changes guard, and the navigation stayed client-side.
- Failure path, forced by stopping the database mid-edit: prior rows and the entered value were both preserved, nothing reverted, and the only report was the localized toast ("Etwas ist schiefgelaufen. Bitte versuche es erneut."). The next edit after restarting the database recovered.

One dev-only console warning was observed: a Radix `useId` hydration mismatch in the sidebar collapsibles and tooltips. No file in this diff is involved.

## Impact and rollback

Behavioural change to every `FilterPopover` consumer: the five entity tables, entity activity histories, inbox threads, users, audit logs, webhooks and webhook deliveries. The dashboard widget editor is untouched: it reuses the filter leaves but never mounts the popover or the modal store, and `flushPendingChanges` is an optional member that its form store does not implement, so those inputs no-op.

`setQueryOptions` gains an optional `refreshMode` that defaults to the existing visible path, so chip removal, clear-all, search and sorting are unchanged.

Rollback is a straight revert of this commit. Existing URLs, P13n rows, named presets, records and database schema stay compatible and need no repair.

### DIFF
--- a/components/data-view/filter-modal/__tests__/edit-filters-modal.store.test.ts
+++ b/components/data-view/filter-modal/__tests__/edit-filters-modal.store.test.ts
@@ -1,0 +1,252 @@
+import { beforeEach, afterEach, describe, expect, it, vi } from "vitest";
+
+import type { BaseDataViewStore, HasId } from "@/core/base/base-data-view.store";
+import type { RootStore } from "@/core/stores/root.store";
+import type { Filter, FilterableField } from "@/core/base/base-get.schema";
+
+const { deleteFilterPresetAction, upsertFilterPresetAction } = vi.hoisted(() => ({
+  deleteFilterPresetAction: vi.fn(),
+  upsertFilterPresetAction: vi.fn(),
+}));
+
+vi.mock("sonner", () => ({ toast: { error: vi.fn(), success: vi.fn() } }));
+vi.mock("@/app/actions", () => ({ deleteFilterPresetAction, upsertFilterPresetAction }));
+
+import { EditFiltersModalStore, FILTER_AUTO_APPLY_DELAY_MS } from "../edit-filters-modal.store";
+
+type SavedPreset = { id: string; name: string; filters: unknown };
+
+const FILTERABLE_FIELDS: FilterableField[] = [
+  { field: "name", operators: ["contains", "equals", "isNull"] },
+  { field: "userIds", operators: ["in", "hasSome"] },
+] as unknown as FilterableField[];
+
+function tableStore(overrides: { filters?: Filter[]; savedFilterPresets?: SavedPreset[] } = {}) {
+  return {
+    customColumns: [],
+    filterableFields: FILTERABLE_FIELDS,
+    filters: overrides.filters ?? [],
+    p13nId: "contacts",
+    savedFilterPresets: overrides.savedFilterPresets ?? [],
+    setQueryOptions: vi.fn(),
+  };
+}
+
+function modalStore() {
+  const root = { registerModalStore: vi.fn(), localeStore: { getTranslation: (key: string) => key } };
+  return new EditFiltersModalStore(root as unknown as RootStore);
+}
+
+function openedOn(table: ReturnType<typeof tableStore>) {
+  const store = modalStore();
+  store.openFor(table as unknown as BaseDataViewStore<HasId>);
+  return store;
+}
+
+function committedFilters(table: ReturnType<typeof tableStore>, call = 0) {
+  return table.setQueryOptions.mock.calls[call]?.[0];
+}
+
+describe("EditFiltersModalStore auto-apply", () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+    upsertFilterPresetAction.mockReset();
+    deleteFilterPresetAction.mockReset();
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it("commits a typed edit once, in the background, after the debounce elapses", () => {
+    const table = tableStore();
+    const store = openedOn(table);
+
+    store.onChange("filters[0].operator", "contains");
+    store.onChange("filters[0].value", "a");
+    store.onChange("filters[0].value", "ac");
+    store.onChange("filters[0].value", "acme");
+
+    expect(table.setQueryOptions).not.toHaveBeenCalled();
+
+    vi.advanceTimersByTime(FILTER_AUTO_APPLY_DELAY_MS);
+
+    expect(table.setQueryOptions).toHaveBeenCalledTimes(1);
+    expect(committedFilters(table)).toEqual({
+      filters: [{ field: "name", operator: "contains", value: "acme" }],
+      refreshMode: "background",
+    });
+  });
+
+  it("does not commit a typed edit before the debounce elapses", () => {
+    const table = tableStore();
+    const store = openedOn(table);
+
+    store.onChange("filters[0].value", "acme");
+    vi.advanceTimersByTime(FILTER_AUTO_APPLY_DELAY_MS - 1);
+
+    expect(table.setQueryOptions).not.toHaveBeenCalled();
+  });
+
+  it("applies an operator change immediately through the flush the inputs call", () => {
+    const table = tableStore();
+    const store = openedOn(table);
+
+    store.onChange("filters[0].operator", "isNull");
+    store.flushPendingChanges();
+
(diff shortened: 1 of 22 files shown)


benjiwagner:
## Follow-up: operator changes no longer discard a usable value

`31c8e7d1`. Feedback from testing the preview of this branch: switching a relation filter from `in` to `not in` lost the selection. That clearing was unconditional before this branch, but auto-apply makes it much more visible, because the filter now drops out of the query the instant the operator is touched rather than only on Apply.

### The rule

A value survives an operator change when the old and new operator render the **same input for that field** and the held value is usable by it. Everything else still clears.

Preserved: `in` ↔ `not in` on relation, scalar-select and single-select fields; `equals` ↔ `contains` on text fields; every pairing among `equals`/`gt`/`gte`/`lt`/`lte` on currency; every pairing among `gt`/`gte`/`lt`/`lte` on dates, plus `contains` on the range column types, where it means "the range contains this date" and renders a date picker rather than a text box.

Cleared: any transition into or out of a value-less operator (`isNull`, `isNotNull`, `hasNone`, `hasSome`, `hasUnset`, `allSet`); either side of `between`; either side of `inLastDays`; and any surviving case where the value would not satisfy `hasValidFilterConfiguration` for the new operator.

### Why the class also drives rendering

`resolveFilterValueClass` is now the single dispatch used both by `FilterField` to choose the input and by the preservation check. A separate mirror of that dispatch would drift the first time a custom column type or a date field is added, and the failure would be silent: the wrong preserve decision, not a compile error.

Two pre-existing dispatch gaps make the extra guards load-bearing rather than redundant:

- `startsAt` (calendar events) is absent from the `dateFields` list in `filter-field.tsx`, so it renders a text input for all six date operators. The explicit `between`/`inLastDays` guards catch what the class comparison alone would miss.
- `calendarId` is absent from the relation list, so its class is `text` on both `in` and `not in` while the wire wants `string[]`. The closing `hasValidFilterConfiguration` check forces the clear.

Neither gap is introduced here and neither is fixed here; the guards just stop the new rule from trusting them.

### One related hole closed

`validDraftFilters` now strips the value from `hasSome`/`hasNone` before committing. A saved preset or a copied URL can carry one (`decodeFilterToken` and `normalizeFilterInput` both retain the value while rewriting the operator), and the entity activity timeline runs `ActivityFiltersSchema.parse` — not `safeParse` — over `.strict()` branches that reject it, which is the hard-failure surface this issue set out to protect. `widget-modal.store.ts` and `use-filter-select-items.tsx` already stripped; the data-view draft was the odd one out.

### Validation

`yarn typecheck`, `yarn lint` and `yarn test` pass (322 files, 2739 tests, 17 new).

New `filter-value-class.test.ts` walks the operator sets from `FILTER_FIELD_DEFAULT_OPERATORS` and `operatorsByType` rather than a hand-written list, and asserts every ordered transition on a custom column preserves exactly when the classes match. Added to the store test: a hydrated `hasSome` filter carrying an array commits as `{field, operator}` and the real `ActivityFiltersSchema` accepts that shape while throwing on the unstripped one.

Verified live on the local instance: `Assigned in Max Bergmann` → `not in` kept the selection and rewrote the URL from `userIds:in:<id>` to `userIds:notIn:<id>` in one request; → `has some` dropped the value, removed the input, and put a bare `userIds:hasSome` on the URL with nothing riding along.

The local production build was not re-run for this commit, to avoid tearing down the local instance in use at the time; the Vercel preview build and `rendered-hubs` cover it.

PR #94, opened 2026-08-19
fix(page-state): render list and detail pages without a second skeleton

benjiwagner:
## Summary

The contacts list and the entity detail pages rendered their loading skeleton twice: the route-level `loading.tsx` skeleton, then the page's own loading branch. Because those are separate React subtrees, the first was destroyed and the second mounted fresh, replaying the staggered `page-skeleton-build` entry animation from `opacity: 0.08`, which reads as a flicker and a second skeleton.

Two independent causes, fixed separately:

- **List routes.** `useDataViewSync` applied the server result from a `useLayoutEffect`, which never runs during SSR. The store therefore stayed `uninitialized`, and `resolveDataViewPageState` maps that to `loading` before it checks `itemCount`, so the server-rendered page was a skeleton even though the rows were already in the RSC payload. It now applies the result during render, keyed on `initialResult` identity so authoritative server props still win on a later render.
- **Detail routes.** The five `[id]/page.tsx` routes never fetched the entity, only the activity timeline; `EntityDetailLayout` loaded it from a `useEffect` after paint, holding the second skeleton for a full server round trip on every visit. The entity is now fetched server-side alongside the timeline and hydrated before the first paint, and the redundant initial `loadById` is skipped.

Hydrating a store during render is the pattern already used by `use-owned-activities-store.ts:26`.

## Context

Linear: [CUS-247](https://linear.app/customermates/issue/CUS-247/stop-list-and-detail-pages-rendering-the-loading-skeleton-twice)
Base commit: `529939e2`
Page state: current.

Authorities read: `core/base/base-data-view.store.ts`, `components/data-view/data-view-state.ts`, `components/entity-detail/entity-detail-page-state.ts`, `core/base/base-custom-column-entity-modal.store.ts`, `styles/globals.css`.

Scope is the page-state seeding path only. The `error | loading | filtered-empty | true-empty | content` contract, the skeleton artwork, the entry animation, the drawer load path, and every route-level `loading.tsx` are unchanged. `EntityDetailLayout` is used only by `EntityDetailPageView`; the drawer has its own load path and is untouched.

## Validation

Verified against a local instance on a disposable Postgres container.

Server-rendered HTML, counting `data-page-state="loading"` sections and whether records reach the DOM rather than only the RSC flight payload:

| Surface | Loading sections before | after | Records in DOM before | after |
| --- | --- | --- | --- | --- |
| `/en/contacts` | 2 | 1 (route fallback only) | 0 | 30 rows |
| `/en/contacts/[id]` | 3 | 2 (route fallbacks only) | 0 | contact name present |

In both cases the remaining sections carry `loading.tsx`'s `h-[calc(100svh-4rem)]` class, so the page's own loading branch no longer renders at all.

Live in a fresh browser tab, after hydration: `/en/contacts` shows 30 rows, 0 loading sections, 0 skeleton shapes; the detail page shows the contact name, 0 loading sections, 0 skeleton shapes. No console errors on either.

Recorded DOM mount sequence before the fix, tagging each `[data-page-state]` element by identity on a soft navigation into a contact:

```
t=342ms   #5 loading (route skeleton)
t=1475ms  #6 loading (page's own branch, a different element)
t=1756ms  content
```

Commands:

- `yarn typecheck` passed
- `yarn lint:check` passed
- `yarn conventions:check` passed, 303 passed / 1 skipped
- `npx vitest run` passed, 2704 passed / 66 skipped across 319 files

New `components/data-view/__tests__/use-data-view-sync.test.ts` guards the contract: the server result is applied during render so a server render needs no effect to leave its loading state, an unchanged result is not re-applied on re-render, and a new result identity is applied. Registered in the `dom` vitest project next to the existing hook test.

## Impact and rollback

User-visible change is that the second skeleton no longer appears; the route-level skeleton still covers the server wait. Detail routes now do one extra server-side query per request, replacing a client round trip that ran on every visit anyway.

Both changes are additive and confined to the page-state seeding path, so this PR can be reverted whole with no data, schema, or migration impact; reverting restores the previous loading behavior.

### DIFF
--- a/app/[locale]/(protected)/contacts/[id]/page.tsx
+++ b/app/[locale]/(protected)/contacts/[id]/page.tsx
@@ -3,7 +3,7 @@ import { EntityType, Resource } from "@/generated/prisma";
 
 import { EntityDetailPageView } from "@/components/entity-detail/entity-detail-page-view";
 
-import { getGetActivitiesInteractor } from "@/core/di";
+import { getGetActivitiesInteractor, getGetContactByIdInteractor } from "@/core/di";
 import { requireAccess } from "@/features/auth/next/require";
 import { ACTIVITIES_P13N_ID } from "@/features/messaging/activities/activities.store";
 
@@ -16,6 +16,9 @@ export default async function ContactDetailPage({ params }: Props) {
 
   const { id } = await params;
 
+  const entityResult = await getGetContactByIdInteractor().invoke({ id });
+  const entity = entityResult.ok ? entityResult.data.contact : null;
+
   const timelineResult = await getGetActivitiesInteractor().invoke({
     scope: activityScopeForRecord(EntityType.contact, id),
     pagination: { page: 1, pageSize: 25 },
@@ -24,6 +27,7 @@ export default async function ContactDetailPage({ params }: Props) {
 
   return (
     <EntityDetailPageView
+      entityInitial={entity && entityResult.ok ? { entity, customColumns: entityResult.data.customColumns } : null}
       entityType={EntityType.contact}
       id={id}
       timelineInitial={
--- a/app/[locale]/(protected)/deals/[id]/page.tsx
+++ b/app/[locale]/(protected)/deals/[id]/page.tsx
@@ -3,7 +3,7 @@ import { EntityType, Resource } from "@/generated/prisma";
 
 import { EntityDetailPageView } from "@/components/entity-detail/entity-detail-page-view";
 
-import { getGetActivitiesInteractor } from "@/core/di";
+import { getGetActivitiesInteractor, getGetDealByIdInteractor } from "@/core/di";
 import { requireAccess } from "@/features/auth/next/require";
 import { ACTIVITIES_P13N_ID } from "@/features/messaging/activities/activities.store";
 
@@ -16,6 +16,9 @@ export default async function DealDetailPage({ params }: Props) {
 
   const { id } = await params;
 
+  const entityResult = await getGetDealByIdInteractor().invoke({ id });
+  const entity = entityResult.ok ? entityResult.data.deal : null;
+
   const timelineResult = await getGetActivitiesInteractor().invoke({
     scope: activityScopeForRecord(EntityType.deal, id),
     pagination: { page: 1, pageSize: 25 },
@@ -24,6 +27,7 @@ export default async function DealDetailPage({ params }: Props) {
 
   return (
     <EntityDetailPageView
+      entityInitial={entity && entityResult.ok ? { entity, customColumns: entityResult.data.customColumns } : null}
       entityType={EntityType.deal}
       id={id}
       timelineInitial={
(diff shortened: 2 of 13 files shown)


benjiwagner:
## Regression found on the preview and fixed: React #418 hydration mismatch

Reported on the preview at `/en/organizations` with sort, `pageSize=100` and two filters. The console showed `Uncaught Error: Minified React error #418`, a hydration mismatch.

**This PR caused it, and the cause is a latent bug it exposed.**

Every date and time formatter in `core/stores/intl.store.ts` builds an `Intl.DateTimeFormat` with no `timeZone`, so it uses the ambient zone. Vercel runs UTC and the browser runs the viewer's zone. That never mattered before, because list pages served a skeleton and no date ever reached the server HTML. Making the server render real rows put those cells into the server HTML for the first time, where they cannot agree with the client.

Nothing stores a viewer timezone: the only `timezone` columns in the schema belong to `Calendar` and `CalendarEvent`, so the server genuinely cannot render the viewer's local time.

### Reproduction

It does not reproduce on a normal local machine, because the dev server and the browser share one timezone. Forcing the production split reproduces it exactly:

```
TZ=UTC yarn next dev --turbopack -p 4002    # server in UTC, browser in Europe/Berlin
```

Same row, same field:

| | value |
| --- | --- |
| Server HTML (UTC) | `09/09/25, 03:00 PM` |
| Browser (Europe/Berlin) | `09/09/25, 05:00 PM` |

Two hours apart on every date cell, which is the mismatch React reports as #418 and recovers from by regenerating the tree.

### Fix

`551be2ba` holds every timezone-dependent formatter until the client has mounted, flipped once from `RootStoreProvider`. Timezone-independent formatting, currency and numbers, is untouched. This is central rather than per-cell on purpose: the exposed surface is the shared tail columns, the webhook, delivery, audit-log and member columns, the `date` and `dateTime` custom-column cells, and the filter chips, so a per-call-site fix would miss cases and would not hold for new ones.

After the fix, on the same UTC-versus-Berlin split:

- server HTML contains **0** timezone-dependent date strings, so there is nothing left to mismatch
- server HTML still contains **12** `data-slot="table-row"` elements and **1** loading section, so the skeleton fix this PR exists for is intact
- the browser renders the dates in local time after hydration

Note the dev console in my browser pane did not surface React's hydration error even before the fix, so the proof here is the server-versus-client HTML, not console output.

`core/stores/__tests__/intl.store.test.ts` gains three cases covering the gate. `yarn typecheck`, `yarn lint:check`, `yarn conventions:check` and the full `vitest run` (2704 passed) are green.

### Separately: this class of error never reaches Sentry

Worth its own issue, not fixed here.

- A recovered hydration error is not reported. `instrumentation-client.ts` initialises Sentry but wires no `onRecoverableError`, and `app/global-error.tsx` only captures errors that reach the error boundary. `instrumentation.ts:40` wires `onRequestError`, so the server side is covered and the client side is not.
- A failed server action returns HTTP 200 with an error payload, so Vercel's runtime logs show a clean 200 and nothing is captured.
- `toastZodErrorTree` and `BaseStore.toastError` show the user a toast with no `Sentry.captureException` anywhere on that path.

So a user-visible client error can occur with no trace in Sentry or Vercel, which is how this one stayed invisible.

PR #102, opened 2026-08-24
fix(seo): technical SEO foundations - status codes, canonicals, linking, and FAQ rendering

benjiwagner:
## Summary

Technical SEO foundations for customermates.com (CUS-70). The site loses ranking signal in ways
nothing in CI notices, because every defect below returns HTTP 200. This fixes the plumbing: real
status codes, one canonical URL per page, a sitemap that matches what the server serves, metadata on
every submitted URL, an internal link graph that reaches every page, and a locale switcher a crawler
can follow.

Content and keyword strategy are deliberately out of scope. This is the measurement layer the
content work has to sit on.

## Context

Search Console, 90 days: **237 clicks on 55,541 impressions, average position 54.8.** 167 of those
clicks are the brand name. The two primary markets produce **one non-branded click each**. 92.5% of
impressions sit at position 21-100 and only 2.9% are on page one, so this is not a snippet problem:
almost nothing is on a page where a snippet would be read.

Every defect was verified individually against live HTTP, Search Console and the SE Ranking audit
before being touched.

**Silent 200s**

- Every bogus slug returned **200** with a noindex and a blank title, and `/en/blog?page=1` returned
  200 carrying `<meta http-equiv="refresh">` where a 308 belongs. `app/[locale]/loading.tsx` put a
  Suspense boundary above every locale route, so the shell flushed before the body could throw and
  `notFound()` degraded to a 404 card inside a 200. The test suite had encoded the tolerance
  (`[200, 404]`, `[200, 308]`) rather than the cause.
- **`/docs` and `/docs/intro-page`** both returned 200, rendered the same MDX, and each
  canonicalised to itself.
- **29 deleted docs URLs** served soft 404s while still earning impressions.
- **Ten reciprocal `X-vs-Y` / `Y-vs-X` comparison pages** competed with themselves. For "zoho vs
  hubspot" one sat at position 71 and its mirror at 55.5, on the same query.

**Missing signal**

- **206 `/docs/openapi/*` URLs and `/contact` had no metadata at all** - bare "Customermates" title,
  no canonical, no hreflang.
- **Eight auth URLs in the sitemap with no noindex.** `/en/auth/signup` listed `sitemap.xml` among
  its own referring URLs.
- **All 404 sitemap URLs carried a `lastmod` and 322 shared one value**, because the plugin reads
  file mtimes and a shallow CI checkout writes every file at clone time.
- **`isSubdomainHost` counts dot-separated labels**, so `www.customermates.com` scored as a subdomain
  and `robots.ts` answered it `Disallow: /`. Only the Vercel edge redirect kept that from serving a
  sitewide disallow on the apex hostname.

**Nothing linked to anything**

- **No crawlable link existed between the English and German trees.** `/en/features`, `/en` and
  `/en/pricing` each emitted 43 internal anchors, **zero** pointing at `/de*`. The language selector
  navigated via `window.location` inside a Radix dropdown, which renders no anchor server-side. This
  is the mechanism behind `/de` sitting at "Crawled - currently not indexed".
- **Related posts picked the 3 newest on every post**, so 3 posts received 71 inbound links each and
  the other 68 received none. Compare, industry and feature pages had no related block at all.
- **Sixteen docs pages shipped one-word titles** (Inbox, Company, Profile, n8n) and eight were
  byte-identical in English and German.

**Two defects found while verifying, both live before this branch**

- **Closed FAQ answers never reached the server HTML.** Radix `AccordionContent` renders a closed
  item as `<div hidden></div>` with no children, so 6 of 7 answers on `/en/pricing` and 4 of 5 on the
  homepage were absent from the rendered page.
- **A raw i18n key rendered as an `<h2>` on 186 sitemap pages.** All three detail collections use
  `getTranslations("StructuredData.breadcrumb")`, which prefixes every key, so a `Common.*` key
  resolved to `StructuredData.breadcrumb.Common.relatedPages`. next-intl renders unresolved keys as
  visible text rather than throwing.

Two corrections to the original plan, both of which would have made things worse: the "~20 docs URLs
missing from the sitemap" are **deleted pages**, so adding them would have submitted 20 soft 404s;
and a blanket 307-to-308 on `localeRedirect` would have let browsers and CDNs pin visitors to a
locale, because the target is negotiated per visitor.

## Validation

`yarn typecheck`, `yarn lint`, `yarn build` and `yarn test` all pass: **2,923 tests across 364
files.** Both end-to-end suites run against a local production build, which is the only place
canonical and hreflang are checked end to end.

- **Every one of the 404 submitted URLs returns 200 and canonicalises to itself.** Nothing in the
  sitemap is a redirect, a duplicate, an auth page or an openapi child.
- **2,005 FAQ answers are present in the visible HTML** with `<script>` tags excluded, up from one
  per page.
- Every page carries exactly one crawlable anchor to its own twin in the other locale, pointing at
  the same page rather than the home page. 396 of 404 carry it; the 8 that do not are paginated hubs
  whose menu links to the hub root, which pagination hreflang already resolves.
- All 71 English blog posts have exactly **3 inbound and 3 outbound** related links, no orphans, no
  self-links. Compare, industry and feature pages carry their related links inline in MDX as
  `<RelatedPages>` blocks: 184 pages, 4 links each, **0 orphans against 74 before**, with every
  hand-curated cluster preserved.
- **Every submitted title is within 60 characters and every description within 160**, measured on the
  rendered string across all 404 URLs, with **no duplicate title and no duplicate description** in
  either locale. Paginated hubs now carry a "Page N" suffix on the description as well as the title.
- **Both hosts were crawled in full and diffed**, live against this branch's preview deployment, so
  each finding is a measured before and after. On real Vercel the preview answers 404 to a bogus
  slug and 308 to `?page=1` where production answers 200 with a meta refresh.
- **Eight dead external citations repointed** to verified live sources. One of them, a German post's
  "Bitkom-Studie zu Vertriebssoftware", has no Wayback snapshot at any point in time and no matching
  path in a CDX scan of bitkom.org, so the sentence now cites the same Salesforce research its
  English twin already used rather than keeping an unverifiable attribution.
- `lastmod` went from 404 URLs sharing one timestamp to **142 with real dates across 59 distinct
  days**, and 262 that honestly declare nothing.
- `/fr/pricing`, `/it/blog/*` and `/` still answer **307**; bogus slugs answer **404**; retired URLs
  answer **308** to their survivors.

Every new guard was verified by planting the defect it exists to catch, not by trusting a green run.
Removing one `generateMetadata` fails the coverage check by name; restoring the three-newest rule
fails the ring properties with "post-001 is linked from 142 pages, not 3"; re-wrapping the locale
menu in a Portal fails with "a force-mounted menu must render inline or SSR drops it"; putting a
`Common.*` key back under a scoped translator names the file, key and scope.

Four findings are **not** acted on, deliberately:

- The 214 "images without alt text" are decorative `alt=""` wave SVGs. Correct practice; the counter
  should not move.
- "408 internal links to redirects" is stale. PR #26 fixed it eight days after that audit ran.
- **Locale-less deep paths still answer 307, not 308.** A blanket permanent redirect would pin
  visitors to a locale, and the scoped version (308 on deep content paths, negotiation kept at `/`)
  is a separate change: those redirects are client-cached, so it needs its own commit and its own
  rollback note rather than riding along here.
- **Two external links still answer 404.** Both are bare `account.microsoft.com` URLs inside the
  privacy policy, which resolve only for signed-in Microsoft users. Legal copy is not edited here
  without review. A further 31 external links answer 403 to automated clients (Gartner, Capterra,
  G2, OpenAI) and load normally in a browser; those are not defects.

## Impact and rollback

The permanent redirects are the risk and are individually revertable, but they will be client-cached,
so undoing them means shipping the reverse mapping rather than a plain revert, and expecting a tail.

`/en/docs/intro-page` holds 507 impressions at position 3.9 and `/en/docs` has none. The 308 should
carry that across, but **expect a one to three week dip**; that is the redirect working, not failing.
The same applies to the retired comparison slugs.

App behaviour is unchanged: `(protected)` keeps its own loader and its 25 per-route loaders; only the
public and static trees lost the shared one. Titles and descriptions changed in frontmatter only, so
no visible heading moved.

The metric to watch at week 4-8 is not clicks. It is the count of URLs with zero clicks (385 today)
and the average position of the 132 pages in the 21-50 band.


app/[locale]/(public)/contact/page.tsx
```diff
@@ -1,9 +1,38 @@
...
 
 import { ContactForm } from "./contact-form";
 
+import { env } from "@/env";
+import { buildAlternateLanguages } from "@/core/seo/alternates";
 import { WaveDecoration } from "@/components/marketing/wave-decoration";
+import { Footer } from "@/app/components/footer";
+import { CONTENT_LOCALES, buildLocalePath } from "@/i18n/locale-registry";
+
+export async function generateMetadata({ params }: { params: Promise<{ locale: string }> }): Promise<Metadata> {
```
Comment (benjiwagner) [inline app/[locale]/(public)/contact/page.tsx:12]: This looks weird. Why cant we use the standard for generate metadata with the mdx file path.

app/[locale]/(static)/docs/openapi/[slug]/page.tsx
```diff
@@ -8,8 +10,19 @@ import { getMDXComponents } from "@/core/fumadocs/mdx-components";
...
+import { Footer } from "@/app/components/footer";
+import { generateMetadataFromMeta } from "@/core/fumadocs/metadata";
 import { contentLocaleOrDefault, formattingTagFor, stripLocalePrefix } from "@/i18n/locale-registry";
 
+export async function generateMetadata({
+  params,
+}: {
+  params: Promise<{ locale: string; slug: string }>;
+}): Promise<Metadata> {
+  const { locale, slug } = await params;
```
Comment (benjiwagner) [inline app/[locale]/(static)/docs/openapi/[slug]/page.tsx:22]: This looks weird. Why cant we use the standard for generate metadata with the mdx file path.

content/compare-pages/de/hubspot-vs-pipedrive.mdx
```diff
@@ -234,6 +234,32 @@ Customermates skaliert anders. Fügen Sie User zum gleichen Tarifpreis hinzu, oh
 
 HubSpot und Pipedrive erzwingen beide Kompromisse. HubSpot bietet Breite auf Kosten von Komplexität und eskalierenden Preisen. Pipedrive bietet Einfachheit auf Kosten von Tiefe und Funktionssperrung. Customermates bietet eine moderne Alternative: vollständige CRM-Kernfunktionalität ab [[commercial.price.starter.monthly]]/User/Monat, EU-gehostet, Open-Source und in Minuten einsatzbereit. Für Teams, die effektives CRM ohne Marketing-Ballast oder Pipeline-only-Einschränkungen wollen, ist es die klare Wahl.
 
+## Häufig gestellte Fragen
```
benjiwagner, line 237: For the FAQ, do you think it is worth it to create a base component where we have the typical accordion style that includes this content, so it's a bit more visually pleasing?

benjiwagner: So when we have added it, please double check all the other MDX files, whether we can add this base component there as well to the same sections.

core/fumadocs/schemas/automation.ts
```diff
@@ -33,6 +33,7 @@ export const pricingTitleSchema = z.object({
 export type PricingTitle = z.infer<typeof pricingTitleSchema>;
 
 export const automationSchema = frontmatterSchema.extend({
+  ...seoOverrideFields,
```
benjiwagner: Can you explain the SEO overrides for me? Do you think we really need this? And if so, explain it to me.

app/[locale]/(public)/contact/page.tsx
Comment (benjiwagner) [reply app/[locale]/(public)/contact/page.tsx:12]: Fixed in ea06b96e. You were right, and it did not need to be bespoke.

The reason it was hand-rolled: `/contact` is a code-rendered page, so there was no MDX file for `generateMetadataFromMeta` to read a title and description from, and I wrote the metadata assembly by hand instead of asking why.

The answer was already in the repo. `content/api-overview/en/openapi.mdx` is a five-line frontmatter-only MDX backing `/docs/openapi`, which is also fully code-rendered. `/contact` now has the same thing at `content/contact/{en,de}/contact.mdx`, a `contact` collection in `source.config.ts`, an entry in `ROUTE_SOURCE_MAP`, and the page is down to:

```ts
const { locale } = await params;
return generateMetadataFromMeta({ locale, route: "/contact" });
```

That also deletes the hand-built canonical and `buildAlternateLanguages` call, and the two `ContactPage` meta keys I had added to all five locale bundles, since the copy now lives with the content rather than in i18n. Verified on a production build: `/en/contact` and `/de/contact` both emit a canonical plus en/de/x-default alternates, same as every other page.

-- Claude

app/[locale]/(static)/docs/openapi/[slug]/page.tsx
Comment (benjiwagner) [reply app/[locale]/(static)/docs/openapi/[slug]/page.tsx:22]: Fixed in ea06b96e, though the answer here is different from the `/contact` one.

This file did already use the standard builder. What was non-standard was a fourth argument I had invented:

```ts
generateMetadataFromMeta({ locale, noindex: true, route: "/docs/openapi/:slug", params: { slug } })
```

Every other noindex route on the site gets that from a registry (`NOINDEX_PUBLIC_ROUTES` in `i18n/routing.ts`), so the auth pages never pass a flag. This route was the one exception, purely because it is not a member of `PUBLIC_ROUTES_SEO` and so was not eligible for that list.

`/docs/openapi/:slug` now joins the noindex set, the `noindex` parameter is deleted from `generateMetadataFromMeta` entirely, and the call is byte-identical in shape to `/docs/:slug`. The 206 generated endpoint pages still emit `noindex, follow` with a self-canonical, verified on a production build.

-- Claude

core/fumadocs/schemas/automation.ts
benjiwagner: Fair challenge, and checking usage says you were two thirds right. Changed in ea06b96e.

`seoOverrideFields` was three fields. Counting what actually uses them:

| field | files using it |
|---|---|
| `metaTitle` | 33 |
| `metaDescription` | 0 |
| `updated` | 0 |

`metaDescription` and `updated` are removed. Neither had a single user, and keeping schema on the argument that something might want it later is how schemas rot. `updated` was wired into `app/sitemap.ts` as a lastmod override, so the sitemap now derives lastmod from a post publication date alone, which is exactly what it did in practice anyway since nothing ever set the field.

`metaTitle` earns its place. Sixteen docs pages shipped one-word document titles, `Inbox`, `Company`, `Profile`, `Records`, `n8n`, and eight of those were byte-identical between English and German because the German frontmatter carried the English string. `metaTitle` overrides the document title only, so the visible `<h1>` on `/docs/app-inbox` still reads `Inbox` while the tab and the SERP read `Unified CRM Inbox: Email, WhatsApp and LinkedIn`. That fixed the short-title and title-equals-h1 findings without touching a single visible heading.

On why it is spread into all fifteen schemas rather than the three that use it: `generateMetadataFromMeta` reads `page.data.metaTitle` from a union of every source in `ROUTE_SOURCE_MAP`, and TypeScript requires the property to exist on every member of that union. Narrowing it to the docs, api-overview and feature-pages schemas fails `tsc`. That is the whole reason, and it is checkable by deleting one spread.

-- Claude

content/compare-pages/de/hubspot-vs-pipedrive.mdx
benjiwagner: I looked into this properly and my recommendation is **not to do it**, at least not as a site-wide change. Here is the evidence, because the answer was not obvious to me either.

**The base component already exists and already works.** `components/marketing/faq-section.tsx` renders a Radix Accordion and is driven by `faq:` frontmatter. Ten content files use it across five route families: homepage, pricing, help-and-feedback, affiliate and n8n-crm. I checked the built HTML and every answer is server-rendered, so it is not hiding content from crawlers.

So the real question is only about the *prose* FAQ sections, the ones written as `## Häufig gestellte Fragen` followed by `### question`. There are two routes to make those accordions and both have a problem.

**Route 1, move the prose into `faq:` frontmatter.** Blocked. `faqItemSchema.content` is `z.string()` and `FAQSection` renders it as plain text, so markdown stops rendering. There are **1,060 markdown links inside FAQ sections across 209 files**. Converting would silently flatten every one of them into literal `[text](url)`. On a PR whose entire purpose is strengthening internal linking, that is the wrong trade.

**Route 2, a build-time rehype plugin** that rewrites the section during MDX compilation. No content edits, links preserved, and SEO-neutral since the answers still ship in the HTML. But it has to run *after* fumadocs own `rehypeToc` to avoid emptying the table of contents, and the only lever for that is the function form of `mdxOptions.rehypePlugins`, which depends on private ordering inside `fumadocs-mdx` (currently 14.2.9, `dist/preset-Y6J6llIV.js`). A patch bump that reorders it would empty roughly 1,400 ToC entries across ~200 pages **with a green build**. That is a silent failure mode on undocumented internals, in exchange for a cosmetic gain.

There are also two boundary hazards: 74 files end the FAQ with a `### Related ...` link block that is not a question, and 22 for-pages files glue a page-closing CTA to the last answer behind an `---` rule, so naive grouping would reorder published content.

**Net:** the gain is visual, the risk is a custom MDX compilation step on private APIs, and the site currently averages position 54.8 with 52 referring domains, so this cannot move rankings either way.

If you want it on a specific high-value page, say `/en/blog/pipedrive-pricing`, 70 of the 279 files have no links in their FAQ at all and could move to `faq:` frontmatter cleanly today, one file at a time, using the component that already exists. Happy to do that for any page you name.

-- Claude

benjiwagner: Surveyed them all, and the numbers argue against a blanket rollout more than for it.

**279 MDX files** carry a level-2 FAQ heading, not the ~200 I first counted; my initial grep missed the `Frequently Asked Questions` variants and the topic-suffixed headings. By collection: blog-posts 141, for-pages 84, feature-pages 30, compare-pages 24. `content/docs` and `content/api` have none. This is all pre-existing content, origin/main has 281, so it is not something this PR introduced.

They are not one shape, they are two:

| shape | files |
|---|---|
| `### Question` | 201 |
| `**Question?**` with the answer in the same paragraph | 78 |

No file mixes them, so any transform needs two branches, and one covering only the `###` shape would silently handle 72% while looking finished.

Three more things that make a blanket pass risky:

- **Heading text is not two strings.** There are about 30 distinct FAQ heading literals. A prefix match on `Häufig` is wrong, because `Häufige Fehler` means "common mistakes" and `Häufige Anwendungsfälle` means "common use cases". Four files carry both a false positive and a real FAQ heading, including `content/blog-posts/de/how-to-choose-crm.mdx` and `content/blog-posts/en/client-qualification.mdx`.
- **74 files** end the FAQ with a `### Related Industries` / `Verwandte Vergleiche` link block that is not a question.
- **22 for-pages files** put a closing CTA between the last answer and that Related block, behind an `---` rule, so grouping by heading alone would reorder what is published.

Combined with the 1,060 markdown links in the other thread, my recommendation is to leave the prose FAQs as they are. They render as plain headings and paragraphs today, which is fully visible to both readers and crawlers, and the accordion is available per page via `faq:` frontmatter where a page actually wants it.

-- Claude

benjiwagner: Correcting myself: you were right and my earlier reply was wrong in the parts that mattered. Built in 5be131c9, on the 24 comparison pages first.

**What I got wrong.** I assessed two routes and wrote a verdict as if it covered all of them. Neither was the one you meant:

- moving the prose into `faq:` frontmatter, where answers flatten because `faqItemSchema.content` is `z.string()`
- a rehype plugin hooking private fumadocs ordering

The component route has neither problem. Children of an MDX component **are** MDX, so the 1,060 links keep working, and `commercialTokens` rewrites `file.value` before parsing, so `[[commercial.*]]` resolves inside JSX exactly as in prose. I also kept repeating an SEO concern I had already disproved with my own measurement: the existing accordion server-renders every answer. And 252 content files already use JSX components with eight registered in `getMDXComponents`, so this is the house pattern, not an exception.

**Shape.** The heading stays in markdown, so the section keeps its anchor and its single ToC entry while the questions leave it, one entry per page instead of seven:

```mdx
## Frequently asked questions

<Faq>
<FaqItem question="Is Pipedrive easier to use than HubSpot?">
Yes, for pure sales pipeline management. See [the comparison](/compare/hubspot-vs-pipedrive).
</FaqItem>
</Faq>
```

First item open, rest collapsed, matching the existing `FAQSection` on homepage and pricing so both FAQ styles look identical.

**Conversion was a script, not hand edits**, because the transform is mechanical and a script can prove itself: every word of every converted file preserved, checked by comparing the normalised token multiset before and after across all 52 files in the collection. The four pilot files whose FAQ ends with a `### Related Comparisons` block keep it outside the accordion, in source order.

**FAQPage structured data** is derived from the same children that render, so it cannot describe a question the page does not show. Honest about its worth: Google restricted FAQ rich results to gov and health sites in 2023, so this is for answer extraction, not snippets.

Verified on a production build: item counts match question counts, every answer is in the server HTML, tokens resolved, links inside answers render.

**Remaining 255 files** are ready when you are: 189 use `### Question`, 66 use `**Question?**`, and the converter already handles both. The only extra care needed is the 22 for-pages files that glue a closing CTA behind a `---` rule inside the section, which the pilot did not contain. Say the word and I will run the rest.

-- Claude

core/fumadocs/schemas/automation.ts
benjiwagner: Replaced in 2db355a9. You were right that this was reinventing something the repo already had, and checking the other schemas found the pattern.

**Every other collection already separates the two titles.** `title` is the document title; the visible heading is a separate field:

| file | `title` | visible `h1` |
|---|---|---|
| `compare-pages/en/folk` | Folk CRM vs Customermates: Honest Comparison 2026 | Folk CRM vs Customermates |
| `feature-pages/en/api` | CRM API - REST API, Webhooks, MCP &amp; OpenAPI | A CRM API Built for Developers Who Ship Fast |
| `for-pages/en/healthcare` | CRM for Healthcare: Patient &amp; Referral Management | The Open-Source CRM for Healthcare |

Docs was the single collection where `title` was also the `h1`, and `metaTitle` patched around that instead of fixing it. That is exactly why it read as an odd extra concept.

**Docs now follows the same rule, and it needed no new field.** `title` is the document title, and the `h1` comes from `DOC_NAV_GROUPS`, the registry the sidebar and breadcrumb already read. I checked coverage before relying on it: 19 nav slugs against 17 doc files plus the empty slug and openapi, nothing missing. So `seoOverrideFields` is deleted from `common.ts` and from all fifteen schemas it had been spread across, and the union problem that forced the spread disappears with it, because only `DocsPageHeader` reads the label and it is docs-specific.

**Reusing the nav label also fixed something `metaTitle` left broken.** The German headings came from German frontmatter carrying the English string, so `/de/docs/app-inbox` rendered "Inbox", `app-company` rendered "Company" and `app-profile` rendered "Profile". They now render **Posteingang**, **Unternehmen** and **Profil** - the registry label is translated, and the sidebar was already showing those words two inches to the left of the wrong heading.

One visible consequence worth flagging: `llms.txt` lists the document title, so an entry now reads "Customermates CRM Inbox: Threads and Replies" where it read "Inbox". I think that is the better string for an AI discovery surface, but it is a change to a published file.

2,909 tests, both end-to-end suites, and a production build where the h1 is the short localised label and the tab is the long title.

-- Claude

PR #125, opened 2026-08-29
feat(operator): add cross-workspace operator console

benjiwagner:
## Summary

A super-admin operator console inside the normal protected platform, built on the same `DataView` plus modal composition every other list in the product uses. Four pages, reachable from an expandable **Operator console** section in the sidebar, beside My Profile and My Company:

- **Overview** — fleet widgets: hosted-AI spend against the monthly cap, revenue at risk, workspaces and users, activity and growth.
- **Users** — cross-workspace directory with search and account, plan, subscription and operator filters. Account status, platform operator access, plan and subscription are edited **in the row**: the chip opens a dropdown of the available values, and a confirmation names the record and the change. Credits sit behind a compact popover on the row, holding the current position, a signed correction and the two calculated resets.
- **Workspaces** — every workspace with its derived label, owner, members, plan and allowance, with plan, subscription and the Enterprise allowance edited in the row the same way.
- **Audit log** — product `AuditLog` and `OperatorAuditEvent` merged into one filterable list with a source discriminator, showing **only changes**.

List state lives in the URL exactly like Contacts or Deals, so filters, sort and pagination are shareable and survive back and forward. Breadcrumbs route through a sibling dropdown like the profile and company sections. Mass actions are off on operator lists: every operator write is confirmed one record at a time. There are no operator modals and no typed reason; the confirmation is the deliberate step, and the audit record carries what changed.

Every layer follows the pre-existing analogue rather than a shape of its own. Operator interactors return `Validated<T>` under `@Enforce`/`@ValidateOutput` over zod DTO schemas, the repository's throws map once at the boundary to `failConflict`/`failNotFound`/`failUnavailable`, and every server action is a `serializeResult` or `unwrapValidated` one-liner. Store mutations follow `ApiKeysStore.delete` (`withLoading` + `toastZodErrorTree` + `refresh`), chip options derive from `USER_STATUS_COLORS_MAP` and the shared subscription colour map, dates go through `useHydratedIntlStore`, each list has its own page view shaped like `contacts-page-view.tsx`, and every loader renders `PageState` over a `DataViewSkeleton` exactly like Contacts. The four-class list repository is split one class per file, matching all 22 existing `prisma-*.repository.ts` files.

There is deliberately **no settings page**. The fleet-wide provider pause and monthly spend cap are edited directly in the database; the console carries no global switches.

## Context

Implements Linear **CUS-126**.

Authorization requires cloud application mode, an interactive cookie session, a fresh unexpired database session, a verified authentication user, normalized email equality, exactly one matching domain user, that user active, a matching company, and persisted `User.isPlatformOperator`. Sidebar visibility is discoverability only; every cross-tenant operation re-authorizes server-side through `@OperatorInteractor`, and page metadata is gated on the same check so a denied route is indistinguishable from any other 404.

Cross-tenant reads reuse `BaseGetInteractor`/`BaseGetRepo`; tenant scoping comes from the repository choosing not to pass a `companyId`.

### Defects found and fixed

Each carries a regression test that was confirmed to fail against the original bug before being kept:

1. Six of fourteen audited actions had no label and rendered as a generic "Operator activity", including all three mutating user-console actions.
2. `generateMetadata` resolves independently of rendering, so a denied route returned a correct 404 body while `<title>` still leaked "User operations". The operator routes now declare no per-page metadata at all, matching every other protected route; the layout keeps only a static `robots` noindex.
3. Operator access was a one-way door: bootstrap granted the first operator then closed permanently, and nothing else wrote `isPlatformOperator`, so a deactivated sole operator locked everyone out with no recovery.
4. Operator list dates rendered blank, because the shared intl store returns an empty string until it hydrates. Operator columns format through `next-intl`.
5. Every operator column logged a missing message on render: the data view resolves its column menu through `Common.table.columns.<uid>`, independently of the header text, so correct-looking headers hid thirteen failed lookups.
6. The workspace subscription correction guarded on the member's `updatedAt` while the repository compares the subscription's own timestamp, so every attempt returned a conflict and the correction could never be applied. Found by exercising the form in a browser, not by the type checker.
7. The global audit log rendered raw action codes such as `operator.user_status.update`. All fifteen audited actions now have copy in five locales, pinned by an exhaustiveness test.
8. **Every chip in a DataView row silently swallowed the row click.** `AppChip` wraps its `Badge` in a `TooltipTrigger`, and Radix's Slot merges the child element's props last; because `Badge` wrote `data-slot="badge"` before spreading props, the rendered chip carried `data-slot="tooltip-trigger"`, which `is-interactive-click` treats as an interactive control, so `data-table` skipped the row handler. Clicking a status or role chip did nothing while clicking the padding around it worked, on members, roles, webhooks, webhook deliveries and audit logs. `AppChip` now sets `data-slot` on the `Badge` element itself, the way `avatar.tsx` already does; fixing it there rather than in `badge.tsx` keeps `Badge` consistent with the other `components/ui` tags and avoids a `Badge` slotted straight into a Radix trigger overwriting that trigger's identity. The regression test renders the real components with no tooltip mock, which is exactly what hid this: the existing `app-chip` test asserts `data-slot="badge"` against a mocked `TooltipTrigger` that never exercises the Slot merge.

### Pattern convergence

A late review found the console had drifted into conventions of its own. The final commit removes that drift rather than adding to it: five dead interactors, two dead action endpoints and their repository methods and schemas are deleted; the bespoke error union, its error-code map and the operator toast hook give way to the shared `CustomErrorCode` contract; DI factories adopt the `get<ClassName>` convention; DTO timestamps become `Date` (ISO strings survive only as optimistic-concurrency tokens); the route `error.tsx`, the dead navigation fields and the crumbs special-casing are removed; and client code moves under `operator/components/{users,workspaces,audit,overview}` like every sibling route.

Three divergences are deliberate and remain. Event publication happens inside the repository transaction, because a cross-tenant write has no ambient tenant to publish under afterwards. No `p13nId` is threaded, because personalisation is per-workspace and these lists are not. Authorization is enforced by the layout plus `@OperatorInteractor` rather than `requireAccess`, because the permission model is per-workspace and operator access is deliberately outside it.

### Review feedback and the repository-pattern pass

The three review comments are addressed: the operator index page is removed (company and profile
have none, and the sidebar links to the first subroute), first-operator bootstrap is removed
entirely since the initial operator is set directly in the database, and `PrismaOperatorAccessRepo`
now lives in its own repository file. Only four files in the repository defined a `BaseRepository`
subclass outside a `prisma-*.repository.ts`; two were these, and both are gone.

A pattern audit against the 22 pre-existing repositories then found two defects rather than style
issues. The audit log translated `inLastDays` as a rolling 24-hour window while `BaseQueryBuilder`
uses `startOfDay(subDays(now, N))`, so the same filter meant different things on different lists, and
`gt`/`lt` collapsed into `gte`/`lte` and wrongly included boundary rows. Separately the workspace
label was computed three ways, one of which read at most 2000 user rows and counted in memory,
deriving the label from an arbitrary subset beyond that; there is now a single shared SQL
implementation and both label tests prove it equivalent.

The remaining convergence: five overrides restated the base class's own defaults, the workspace list
hand-rolled a relation search and passed `searchTerm: undefined` to defeat the base builder (it now
declares `users.email`), `mapAdjustment` was an identity function, `buildQueryArgs` results are
spread as eleven siblings do, three pure reads were wrapped in transactions left over from when reads
were audited, and the `Audited` infix is gone from three methods that write no audit event. Four
further suggestions were rejected with counts rather than accepted: `keyof typeof` appears twenty
times elsewhere, both schema layouts exist in the codebase, `@Transaction` cannot pass `companyId`,
and `operator-list-filters.ts` is used by all three list repositories.

### One inconsistency found by driving the console, not by reading it

The Enterprise allowance popover was the only operator mutation that committed without confirming.
Every inline chip and both credit resets call `showConfirmation` and name the record; Save on the
allowance wrote straight through, on the widest-reaching write in the console, since the allowance is
granted per user across an entire workspace. It now raises the same confirmation, naming the
workspace and the value and disclosing the workspace-wide scope, in all five locales.

### Security defect found and removed before merge

An adversarial audit of the authorization boundary found one issue, and this branch introduced it
rather than inheriting it. `shouldIncludePreviewOperatorAccess` granted `isPlatformOperator` to the
synthetic seed user on any deployment with `VERCEL_ENV=preview`, keeping none of the guards its local
counterpart enforces: no explicit opt-in, no `!VERCEL`, no `!CI`, and no proof the database is
loopback-local. `scripts/vercel-build.sh` runs the seed on every preview, and the build log for this
branch's own preview confirms it executed against the remote Neon database.

The account it flagged is `max.bergmann@customermates.com`, whose credential row the seed rewrites on
every run with the compile-time constant `SYNTHETIC_SHARED_USER_PASSWORD` and `emailVerified: true`.
The project has password protection, Vercel SSO and trusted IPs all disabled, and branch preview URLs
are deterministic. The grant therefore put full cross-workspace operator read *and* write behind a
password that lives in the source tree.

One qualification, stated rather than glossed: whether the grant actually fired depends on `NODE_ENV`
inside Vercel's build shell at the moment the seed runs, which could not be determined without either
signing in to the preview — which is the exploit — or adding an echo and redeploying. So this is a
confirmed-reachable code path of uncertain live exploitability, not a confirmed live compromise. It is
removed either way: gating a cross-tenant super-admin grant on an undocumented platform variable is
wrong even when that variable currently happens to save you, and the test the branch shipped asserted
the grant *did* fire.

`shouldIncludeOperatorSeedAccess` now defers entirely to the local opt-in, which already requires a
loopback database and the absence of any build environment. The test that pinned the preview grant was
asserting the vulnerable behaviour, so it is replaced by one pinning its absence plus each guard the
opt-in enforces; re-planting the old function makes that test fail, which was verified.

### Deliberate changes to the original contract

- **`OPERATOR_CONSOLE_ENABLED` is removed.** The console ships always; the boundary is cloud mode plus a persisted `isPlatformOperator` plus fresh authorization. The flag added little security while silently disabling the console in every new environment. `HOSTED_AI_OPERATOR_CONTROLS_ENABLED` **stays**, because it gates the fleet-wide pause and spend cap that can stop hosted AI for every customer.
- **Reads are neither recorded nor listed.** Only mutations reach the audit log. The action constants and labels remain so any rows already written still render.
- **Operator access is never seeded on any Vercel environment, preview included.** An earlier commit on this branch did seed it on previews so a pushed branch would be reviewable; that was removed as a security defect (see below). Preview reviewers are granted operator access out of band.

## Validation

All gates run locally against the merged tree (this branch contains current `main`), on a disposable local PostgreSQL 17 container. Production and the hosted demo were never used.

- `RUN_DATABASE_TESTS=true yarn test` — 508 files passed, 1 skipped; **4,456 tests passed, 2 skipped**
- `yarn lint` — clean at `--max-warnings 0`
- `yarn typecheck` — clean
- `yarn i18n:audit` — clean
- `yarn build` — succeeds; all four operator routes emit as dynamic (`ƒ`), so every request re-authorizes
- `git diff --check` — clean

The cross-tenant query layer is proven against a real database: listing across workspaces, subscription and boolean filters, account-status filtering, search, derived workspace labels, workspace member aggregates, and the merged audit log including source filtering, ordering, pagination across both tables, the unfiltered first page the route actually requests, and the exclusion of read events. The platform-access guards are proven inside a deliberately rolled-back transaction, which makes the platform-wide operator count deterministic.

Exercised against a local production build (`next start`) on a disposable local PostgreSQL 17 container, signed in as the seeded operator. All four pages render in the protected shell with cross-workspace data, the sidebar shows the expandable **Operator console** section, and pagination reads `1-25 of 37` and stays correct after an inline write. One inline edit was driven fully through the real UI: the status chip opened its dropdown (Active / Inactive / Waiting for approval), the confirmation named the record and the change ("Set the account status of Inactive Pro Active 32 to Inactive?"), confirming produced an "Updated" toast, the chip re-rendered, the row persisted to the database, and exactly one audit row was appended recording `active` to `inactive`.

The remaining nine actions were driven directly against the running server with real payloads, each verified in the database: platform-access grant and revoke, a signed credit correction, a calculated credit reset (which correctly appends a compensating adjustment rather than editing in place), an Enterprise allowance change, a subscription snapshot correction, and the three list reads. Across five such mutations the audit table grew by exactly five rows. Over roughly fifty read operations in the same window, **zero** read events were written, while the pre-existing read rows still render — confirming the reads-are-not-audited decision. Both new credit adjustments carry a `NULL` reason, exercising the migration.

**Authorization was proven by forged request, not by inspection.** Sessions were minted for four identities, one per guard clause in `findAuthorizedActorUnscoped`: the real operator; a user with no operator flag; a user *with* the flag but `inactive`; and a user with the flag, `pendingAuthorization` and an unverified email. Anonymous requests to all five operator routes redirect to sign-in (307). The three denied identities receive no cross-tenant data on any of the four pages and no sidebar entry, while the operator's page carries 31 emails and 53 record identifiers. All ten operator server actions were then invoked directly by ID, extracted from `server-reference-manifest.json` and the client bundle, under each denied cookie: **all thirty calls were rejected with zero data returned**. A fully valid deactivation payload — one that succeeds for the operator — was replayed by each denied identity; every one was refused and the target's `status` and `updatedAt` were both unchanged, so not even a timestamp moved. The same payload as the operator returned `{"ok":true}` and persisted, which is the positive control proving the test measures something.

One limitation, stated plainly: the automated browser reports `document.hidden === true`, so `ResizeObserver` never fires and table cells report zero width after a re-navigation. Interactive clicking could therefore be driven through the complete journey once but not repeatedly. The plan, subscription and operator chips use the identical `OperatorChipSelect` component as the status chip that was driven end to end, and all four of their commit paths are proven above at the action layer with real writes and audit rows.

**Preview:** https://feat-hosted-ai-operator-console.customermates.com — READY. Anonymous visitors are redirected to sign-in on the operator routes exactly as on any other protected route; behind-login verification was done locally.

## Impact and rollback

**Migration** `20260828120000_hosted_ai_operator_control` is additive and backward compatible: adds `User.isPlatformOperator` (default `false`), creates `AgentCreditAdjustment`, `HostedAiGlobalControl` and `OperatorAuditEvent` with their indexes and check constraints, and inserts the singleton `HostedAiGlobalControl` row fail-closed (`hostedProviderWorkPaused = true`).

**Environment.** One switch, `HOSTED_AI_OPERATOR_CONTROLS_ENABLED`, accepting only `"true"` or `"false"` and defaulting to `false`. **Enable order matters:** because the seeded control row is fail-closed, turning it on will immediately block new hosted-AI provider work until the row is unpaused and given a finite monthly cap. Since there is no settings page, unpause it in the database in the same maintenance step.

**Operator access.** No user is granted operator access by deployment. The first production operator is set deliberately (`yarn operator:bootstrap`, or directly in the database); every operator after that is granted in the console by an existing operator, audited, and guarded against self-change and against removing the last active operator.

**Rollback.** Revert this pull request. The schema is additive and the application tolerates the extra column and tables, so no down migration is required. To disable hosted-AI enforcement without reverting, set `HOSTED_AI_OPERATOR_CONTROLS_ENABLED` to `"false"`; to remove operator access entirely, clear `isPlatformOperator`.


app/[locale]/(protected)/operator/page.tsx
```diff
@@ -0,0 +1,5 @@
+import { redirect } from "next/navigation";
+
+export default function OperatorIndexPage() {
```
Comment (benjiwagner) [inline app/[locale]/(protected)/operator/page.tsx]: Why do we need this?

ee/operator/bootstrap-first-operator.ts
```diff
@@ -0,0 +1,22 @@
+import { createInterface } from "node:readline/promises";
```
benjiwagner: We don't need this because the initial operator I will set myself in the database.

ee/operator/operator-access.service.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export function normalizeOperatorEmail(email: string): string {
+  return email.trim().toLocaleLowerCase("en-US");
+}
+
+export abstract class OperatorAccessRepo {
+  abstract findAuthorizedActorUnscoped(session: InteractiveSession): Promise<OperatorActor | null>;
+}
+
+export class PrismaOperatorAccessRepo extends BaseRepository implements OperatorAccessRepo {
```
benjiwagner: Prisma repositories usually have their own repository files.

app/[locale]/(protected)/operator/page.tsx
Comment (benjiwagner) [reply app/[locale]/(protected)/operator/page.tsx]: Removed. Nothing linked to it: the sidebar builds its href from the first subroute (`/operator/${OPERATOR_SUBROUTES[0]?.slug}`), exactly as company and profile do. Neither of those sections has an index page either, so this was the odd one out. Its `loading.tsx` went with it, since the page-state convention test asserts every loader has a sibling page.

ee/operator/bootstrap-first-operator.ts
benjiwagner: Removed entirely: the script, `operator-bootstrap.service.ts`, the `operator:bootstrap` package script, the `operatorBootstrap` audit action and its five locale labels, and the two database tests that covered it. `AuditActionLabel` falls back to the raw action string for unknown keys, so any row already written still renders rather than crashing.

ee/operator/operator-access.service.ts
benjiwagner: You are right, and this was the pattern break you flagged. `PrismaOperatorAccessRepo` now lives in `ee/operator/prisma-operator-access.repository.ts`. The abstract `OperatorAccessRepo` stays in the service file, which matches the convention: `PrismaAuditLogRepo` is alone in its prisma file while `GetAuditLogsRepo` sits with its consuming interactor. Only four files in the repo defined a `BaseRepository` subclass outside a `prisma-*.repository.ts`; two were mine, and both are now gone.

ee/agent-chat/hosted-ai-admission.ts
```diff
@@ -0,0 +1,20 @@
+export type HostedAiAdmissionBlockedReason =
+  | "configuration_unavailable"
+  | "credits_exhausted"
+  | "global_spend_cap"
+  | "operator_paused";
+
+const HOSTED_AI_ADMISSION_BLOCKED = "HOSTED_AI_ADMISSION_BLOCKED";
+
+export class HostedAiAdmissionBlockedError extends Error {
```
benjiwagner: It seems untypical. Why do we have an extra error defined over here? Why don't we use the normal zod errors or plain thown Errors?

benjiwagner: You were right, and the file is gone.

I checked what main does and it already had the answer in the same two methods: a refused admission returns `null`/`false` and the interactor maps it with `fail*`; plain `Error` is reserved for invariants. `fail*` is never called from a repository, and no repository carries a failure kind to its interactor through an error class.

The class was also doing less than it looked. Its `reason` union was never read at any call site, and all five catches discarded the error and returned the `null`/`false`/`failRateLimit` main already returns directly.

The part I had missed until you pushed: it had **replaced two throws that predate this branch**. Main throws a plain `Error` for a blocked entitlement and for `reservation exceeds the current allowance`, and this branch had swapped both for the class and edited main's own test assertions to match (`/exceeds the current allowance/` → `/credits_exhausted/`). Both throws and both assertions are restored verbatim — main's concurrency test caught it when my first attempt got this wrong.

What's left is only the genuinely new global control: `lockHostedAiGlobalControlForAdmission` returns a boolean, so the three call sites that already returned `false`/`null` for the neighbouring condition now read identically one line apart, and the interactor maps it with `failRateLimit(CustomErrorCode.agentLimitReached)` exactly as it does twenty lines earlier. `openTurn` dropped its try/catch entirely.

One exception, called out so it isn't mistaken for the old pattern: `prepareStep` keeps a turn-local sentinel compared by reference, because the SDK callback has no return channel and returning would let the very provider call we're blocking proceed.

Gate: lint, typecheck, 510 files / 4,467 tests, build — all green. I also planted a broken global control to confirm the converted assertions still fail (6 did).

PR #126, opened 2026-08-30
fix(errors): handle malformed invite tokens and docs copy failures

benjiwagner:
## Summary

Fixes the two open production Sentry issues unrelated to the Unipile integration, and rebuilds the clipboard helper on the current cross-browser contract.

- **CUSTOMERMATES-5F** — `GET /[locale]/invitation/[token]` returned 500 for a malformed token, because the format lived in a throwing `Enforce` schema and fired before the interactor's own `valid: false` branch. It now uses the non-throwing `Validate` decorator and returns the failure to its callers.
- **CUSTOMERMATES-5E** — `NotAllowedError` from `navigator.clipboard.writeText` on the docs pages. Two distinct causes, both fixed: the fumadocs code-block button fired the write as an uncaught floating promise, and our own "Copy Page" button called the Clipboard API *after* awaiting a fetch, which Safari refuses.

Linear: [CUS-287](https://linear.app/customermates/issue/CUS-287/stop-the-unhandled-invitation-and-docs-copy-errors-reaching-sentry)

### Invite tokens

`Enforce` throws on a schema violation, which suits a system interactor called by our own code and is the dominant choice there (27 of 42 uses). But this interactor's token arrives from a URL segment or a cookie. Every system interactor that already takes untrusted input uses `Validate`: `sign-up-with-email`, `sign-in-with-email`, `reset-password`, `request-password-reset`, `continue-with-socials`, `decide-mcp-consent`, `send-contact-inquiry`, `register-user`. This one was the outlier.

Declaring `Validated<ValidatedInviteToken>` also makes the compiler require every caller to narrow on `ok`, so the crash cannot be reintroduced silently. Verified by reverting one caller: `tsc` fails with `Property 'data' does not exist on type 'InteractorOutcome<...>'`. Only `onboarding/wizard` already narrowed correctly; the other four now match it.

### Clipboard

Safari expires the user activation the Clipboard API requires as soon as the calling code awaits anything, and rejects with `NotAllowedError`. Our "Copy Page" button fetched the markdown and only then called `writeText`, so the **first** click on any docs page failed and showed the failure toast, while the **second** succeeded from the in-memory cache. That cache is what made it look intermittent rather than broken.

The documented remedy is to invert the flow: call the Clipboard API synchronously from the handler and hand it the promise, letting it resolve the content while the activation is still valid. `copyToClipboard` now accepts `string | Promise<string>` and writes an unresolved promise through `ClipboardItem`, falling back to `writeText` where `ClipboardItem` is unavailable (Firefox still keeps it behind the disabled `dom.events.asyncClipboard.clipboardItem` preference), and finally to the deprecated selection copy for insecure contexts. The selection fallback now also restores the caller's prior selection and uses `setSelectionRange`, which is what iOS needs.

The docs code block additionally stops using fumadocs' built-in copy button, whose `navigator.clipboard.writeText(...)` at `fumadocs-ui/dist/components/codeblock.js:73` is an uncaught floating promise. We supply our own through the component's supported `allowCopy` and `Actions` props.

Left out deliberately: CUSTOMERMATES-J is a generic-message group spanning about fifteen transactions with no stack trace and needs source maps first (CUS-144); CUSTOMERMATES-3S was already fixed on 5 August by migration `20260805093749_entity_terminology_drop_sales_type`.

## Validation

Gates: `yarn typecheck`, `yarn lint`, `yarn test` (4185 passed / 123 skipped / 0 failed), `yarn build`, all clean after merging `origin/main`.

**Real Safari 18.5, driven over WebDriver with trusted clicks:**

| Pattern | Promise | Copied |
| --- | --- | --- |
| Old: `await fetch(...)` then `writeText` | `REJECTED: NotAllowedError` | nothing |
| New: `ClipboardItem` holding the promise | `RESOLVED` | 7694 bytes |

- "Copy Page", cold cache, first click: success toast, full document on the clipboard. Previously this was the failing case.
- Code-block copy, first and second click: copies, no toast, zero unhandled rejections.
- The old fumadocs floating-promise pattern reproduces `['NotAllowedError']` as an unhandled rejection; ours produces none.

**Invitation route**, end to end against a locally provisioned PostgreSQL 17 with migrations and the synthetic seed:

| Token | Before | After |
| --- | --- | --- |
| `{token}`, `has spaces`, `slash/token`, `a+b` | **500** | 307 → `invalidInviteLink` |
| unknown but well-formed | 307 → `invalidInviteLink` | unchanged |
| expired | 307 → `inviteLinkExpired` | unchanged |
| valid | 307 → `auth/signup` + cookie | unchanged |

Sign-in and sign-up render the invited variant only for a valid token. Registration through the better-auth `user.create.before` hook associates the company only for a valid invite, and a malformed cookie degrades to "not invited" instead of throwing. `onboarding/wizard` verified with a real session across all cookie states. All eight MDX surfaces load. Zero ZodErrors and zero 500s across the sweep.

**Unit**: a new `core/utils/__tests__/clipboard.test.ts` covers the tier order, the Firefox path, both fallbacks, and selection restoration. Its key case asserts that a still-pending promise reaches the Clipboard API *before* resolving; reintroducing the `await`-first defect makes exactly that test fail.

## Impact and rollback

No schema, migration, persistence, or configuration change; no new dependency. `copyToClipboard` widens its parameter type, which is backwards compatible for the existing string callers (copyable chips and code, contact channels, custom field values, agent chat).

One allowlist entry in `tests/conventions/tenant-write-scoping.test.ts` moves from `better-auth.ts:100` to `:103`, because that gate is keyed by `file:line` and the added guard shifts it. The `updateMany` it exempts is unchanged.

The docs copy control is shared by all MDX pages; one incidental effect is that its `aria-label` moves from fumadocs' hardcoded English "Copy Text" to our existing `Common.actions.copy` key, so it is now localized.

Rollback is a straight revert of this pull request.


features/company/invite-token-validation.interactor.ts
```diff
@@ -34,7 +33,7 @@ export class InviteTokenValidationInteractor {
   @Enforce(Schema)
   @ValidateOutput(OutputSchema)
   async invoke(data: InviteTokenData): Promise<{ ok: true; data: ValidatedInviteToken }> {
-    if (!data.token) {
+    if (!data.token || !TOKEN_FORMAT.test(data.token)) {
```
benjiwagner: Doesn't it maybe make more sense to instead of use the enforced decorator use the validate decorator then? Doesn't this more naturally fit our use case?

PR #142, opened 2026-09-02
fix(messaging): stop one throttled source truncating a whole account

benjiwagner:
## Summary

Three defects found during a live end-to-end check, all in our code rather than the provider's. They are cause, symptom and adjacent, so they travel together.

| Fix | Defect |
| --- | --- |
| Scope the give-up to the source | A long `Retry-After` on any one source ended the drain for the whole account |
| Report completeness honestly | A truncated drain still stamped `lastSyncedAt` and looked fully synced |
| Two error mappings | A timeout read as a rejection, and a permission refusal read as an outage |

## Context

Refs CUS-301, CUS-302, CUS-303.

**This branch is stacked on `fix/unipile-live-defects` (PR #140)** and targets it, not `main`. Both fixes touch `drainSources`, which #140 already rewrites, so branching off `main` would have guaranteed a conflict. Merge #140 first.

The truncation was found with a real LinkedIn account. The same account, same code, three runs:

| Run | Threads | Messages |
| --- | --- | --- |
| Aborted by `Retry-After: 52005` | 1 | 0 |
| Partial re-run | 9 | 40 |
| Clean run | 140 | 523 |

The header was not a genuine block. Enumerating all 15 inboxes directly showed 6 readable, 9 returning `403 provider/insufficient_permissions`, zero rate limiting, and `X-Ratelimit-Remaining` at 273 of 300. One header cost 139 threads, and CUS-302 is why nobody would have noticed.

## What changed

`drainSources` now defers a throttled source with a warning and continues to the next, returning whether anything was left undone. Abandoning a source after repeated stalls counts as incomplete too. Resweep behaviour is unchanged: an incomplete pass still skips resweeps, so a throttled account does not get hammered.

`releaseBackfillClaimUnscoped` takes a `complete` flag and only stamps `lastSyncedAt` when every source finished. The claim is still released either way, so nothing can wedge. The schema requires the flag rather than defaulting it, so a future caller cannot silently claim success by omission.

`unipileErrorCode` gains a `status === 0` branch mapping to `unipileRequestTimeout`, whose copy already existed in all five locales with no producer. `provider/insufficient_permissions` maps to `unipileResourceNotFound` instead of falling through to the generic provider-error copy.

## Validation

Every fix carries a test that was mutation-tested: removing the timeout mapping, removing the permission mapping, and making the completeness flag default to true each fail a test, and all pass again on restore.

The repository behaviour was verified against the real local database rather than a mock: releasing with `complete: false` leaves `lastSyncedAt` untouched while clearing `syncing`, and `complete: true` stamps it.

Full suite 4496 passing, typecheck and lint clean.

## Impact and rollback

No schema migration. The only behavioural change visible to a user is that an account whose backfill did not finish no longer shows a fresh Last synced time, which is the point.

`releaseBackfillClaimUnscoped` gained a required parameter, so both implementers were updated; `refresh-inbox` passes `true`, matching its previous behaviour exactly.

Rollback is a straight revert.


benjiwagner:
Folded into #140 and closing this.

Keeping them apart stopped paying off: this repo's CI is `branches: [main]`, so a stacked PR ran nothing but Vercel, and retargeting at main made this PR a superset of #140 rather than the focused diff it was meant to be. It had also fallen 2 commits behind.

The commit is cherry-picked onto #140 cleanly, no conflicts, and the combined tree differs from this branch only by the two commits it was missing. 4496 tests passing.

One thing carried over that a reviewer should know: the CUS-301 fix, deferring a throttled source instead of abandoning the account, has no automated test. `drainSources` cannot be imported in a test (the workflow module times out on import) and a stub Unipile could not reproduce the provider's source topology faithfully enough to trigger the throttle. It rests on review plus the deferred flag reaching `releaseClaim`, which is tested. Extracting the per-source decision into a pure helper would close that.

PR #149, opened 2026-09-02
feat(data-transfer): record an audit entry for every record export

benjiwagner:
## Summary

Adds the `records.exported` audit event that CUS-285 required and #127 shipped without, and corrects the self-hosting docs that still deny the feature exists.

## Context

CUS-285 lists "An export writes exactly one audit entry naming actor, entity type, scope and row count" as an acceptance criterion, and under Constraints and safety: "Because export is deliberately ungated, an audit event is part of the definition of done." Export authorization is existing read permission and nothing more, so the audit log is the only record a tenant has that data left. After #127 merged, `RECORDS_EXPORTED` did not exist anywhere on main and no file under `features/data-transfer/` published anything.

Where the publish goes was the whole problem. Inside the export interactor it fires once per page, up to a hundred rows for one export. In the route body it throws `Tenant context missing`, since `runWithTenant` spans only the interactor's own `invoke`. At `skip === 0` it would fire once but record the matching total rather than the delivered count, and would claim an export happened even when a later page fails and the caller gets a 400 with no file.

The answer is a small interactor called once, after the workbook is built and after the failure branch. It re-establishes tenant context, re-checks `readAll OR readOwn` for that entity type, and records what was actually delivered.

Two deliberate choices:

- **Not in the webhook registry.** `EventService.publish` returns before touching the webhook repo or background dispatch for unregistered events, which keeps delivery failures off the export path entirely. `LEGAL_NOTICE_SENT` is the precedent.
- **No catch around the publish.** `publish` is awaited under `Promise.all`, so a failed audit fails the export. For an ungated export that is the safer direction: delivering records we cannot record is the thing the criterion exists to prevent. The remaining throw source is the audit insert itself, which correlates with a database already broken enough to have failed the export's own reads.

The payload carries entity type, row count, truncation and whether the caller exported a selection or the current view. Filters, search terms and selected ids stay out: a row count is not a reason to copy someone's segment definition into a durable log.

The second commit fixes `content/docs/{en,de}/self-hosting.mdx`, which still told self-hosters there is no import or export workflow.

## Validation

- `yarn typecheck`, `yarn lint`, `yarn i18n:audit` clean; `yarn build` succeeds; generators leave the tree clean.
- Full suite with `RUN_DATABASE_TESTS=true`: 4781 passed, 2 skipped, no failures.
- `export-audit.database.test.ts` against real Postgres: exactly one row, correct actor, company as entity id, delivered count and truncation flag, and the serialised row contains no `filters`, `searchTerm`, `selectedIds` or `sortDescriptor`.
- Every new test was proved to fail before being trusted: removing the publish, removing the permission gate, attributing the row to the user instead of the company, and pointing the privacy assertion at a field that is present each produce a failure.
- The corrected docs pass the `non-xlsx-data-transfer` and `whole-account-data-export` claim rules.

Two convention tests rejected the first version, which hand-threw `ForbiddenError`; the throw now comes from `hasPermissionOrThrow` in access infrastructure.

## Impact and rollback

Additive. One new `DomainEvent` member, its payload type, `null` in `AUDIT_EVENT_ENTITY_TYPE` so exports stay out of per-record activity feeds, a name extractor, one i18n key in five locales, one interactor, one DI registration and one call site. No migration: `AuditLog.event` is a plain `String`.

Existing exports gain an audit row and one extra round trip. Nothing else changes behaviour. Revert the merge; the event disappears and historical rows become unlabelled strings in the audit list, which is the same state as before this PR.


features/data-transfer/export/record-export-audit.interactor.ts
```diff
@@ -0,0 +1,57 @@
...
+export const RecordExportAuditSchema = z.object({
+  entityType: z.enum(EntityType),
+  rowCount: z.number().int().min(0).max(EXPORT_ROW_LIMIT),
+  truncated: z.boolean(),
+  scope: z.enum(["selection", "view"]),
+});
+
+export type RecordExportAuditData = Data<typeof RecordExportAuditSchema>;
+
+@AllowInDemoMode
```
benjiwagner: The exports should not be available in the demo env, nor should the record

benjiwagner: Done, and the file this comment sits on is gone.

Demo is now closed on both directions: the five export interactors and `get-import-relation-index` lose `@AllowInDemoMode`, and the dry-run and commit interactors never had it. Nothing in the feature is demo-allowed.

On the record: rather than decorate a separate interactor, the publish moved into `BaseExportRecordsPageInteractor` at `skip === 0`, so the interactor doing the work publishes its own event like everywhere else. That also collapses your two requirements into one switch — blocking the export now blocks the record, instead of them being two decorators that could drift apart. The second permission gate and the DI entry are gone with it.

The trade worth your eye: it records what the export covers when it is authorised, not what the file finally contained, so an export that dies on a later page still leaves a row. I think that is the right direction for an ungated export, since those pages were already read. Making it delivery-accurate means moving the whole export into one interactor, which would hold 50,000 DTOs in memory instead of one page.

One thing your comment implied that I would have missed: closing the server side alone left demo users reading "The file could not be created", and on import a file-rejected message about a file that was never at fault. The export toast now stands aside in a demo environment so the existing upsell toaster owns the message, and the wizard shows `Common.errors.demoMode`, which already exists in all five locales. The hostname check only picks wording — the refusal stays server-side in `APP_MODE`, so a host that merely contains "demo" cannot lose a working feature.

PR #154, opened 2026-09-04
fix(onboarding): preserve invites and require workspace choice

benjiwagner:
## Summary

- Preserve validated invitations as signed, per-tab onboarding intent through authentication, verification and password recovery.
- Require an explicit Create or Join choice for first-run users without an invitation. Reuse the existing option cards and numbered Steps in all five locales.
- Make existing-identity registration and operator deletion safe against obsolete company references, duplicate submissions, conflicting tabs and deleted identities.
- Align all identified consistency cases: auth intent and sign-in callbacks belong to form data; invitation, inbox linking and import orchestration belong to interactors; Company owns weighting configuration reads; the service quick-create default has one interactor owner.

## Context

[CUS-312](https://linear.app/customermates/issue/CUS-312/let-an-existing-account-accept-a-workspace-invitation) covers a signed-in visitor losing an invitation before validation and being sent into unintended workspace creation. Invitation binding previously only ran when an AuthUser was first inserted.

Direct Company deletion can leave an identity with an obsolete company reference. This is distinct from the current operator action, which deletes actual workspace-member identities and their sessions. The operator hardening also clears deleted-company references on unrelated tenantless identities without deleting those identities.

The current explicit invitation is authoritative until membership exists. Join creates no records and explains how to request an invitation. Create requires a deliberate, account-bound decision. Existing workspace members are never silently moved to another workspace.

The owner requested that the PR review principles be applied to every occurrence identified in the consistency audit, including the related inbox, import, weighting and service paths. These changes reuse the existing decorators, validation, repository ports, transactions and child interactors rather than introducing a new framework. All five existing review threads are resolved.

## Validation

- Node 24: typecheck, lint with zero warnings and production build passed. Build and dev never ran concurrently.
- Full database-enabled Vitest run: **5,199 passed, 2 skipped, 0 failed**. Conventions: 638 passed, 2 skipped. The two skips are rendered-output contracts covered by the separate rendered-hubs check.
- The full run uses one worker to avoid unrelated filesystem-scan/email-preview timeouts under heavy parallel host load. The previous parallel run had six such timeouts; the complete serial rerun passed. JSON reporter counts were checked directly.
- Added real DOM reset-form lifecycle tests, nullish-input validation tests, auth context refresh/clear tests, invitation interactor tests, inbox action/unit/DB tests, all-five-target import dispatch tests and a real import roundtrip, weighting isolation/rollback/deletion DB tests, and service quick-create validation/default tests.
- After the final test-only fixture color correction, weighting DB tests passed again: 6/6. No production code changed after the full build/test gates.
- Three independent final read-only code reviews found no blockers. Independent focused runs passed 189 and 56 tests. Final screenshot review found no blocking visual issue.
- Fresh UI verification used multiple synthetic accounts, invitations, separate operator/member sessions and competing same-session tabs on localhost, with database assertions for actual persistence:

| Local UI journey | Verified result |
| --- | --- |
| Existing identity follows invite; header Login and Continue setup | Invited profile, then Waiting for Approval in the correct workspace |
| New signup with and without an invitation | Invite bypasses chooser; ordinary signup shows Create/Join |
| Join, direct wizard entry, explicit Create | Join writes nothing; direct entry returns to choices; deliberate Create writes exactly one workspace after profile submission |
| Creator profile, invite and AI steps, Back, Next, Skip, Finish | Setup progress remains visible and completion reaches dashboard |
| Overdue unverified invitee | Real local verification callback retains the invitation and joins the invited workspace |
| Verification deadline crosses while creator profile is open | Submit requires verification; local callback returns to creator profile; submission creates one workspace |
| Deleted-company reference or different live pre-tenant binding | Current invitation wins; pending member in invited workspace |
| Existing active, pending or inactive membership opens another invite | Account-switch screen; no rebinding |
| Two invitation forms in the same session | Each retains its inviter; after the second joins, returning to the first reaches the pending guard without changing membership. Late/double-write races are additionally covered by DB tests |
| Operator deletes a workspace while owner browser remains live | Company, member, identity and sessions removed; audit recorded; old browser reaches fresh invitation signup |
| Deleted owner signs up again using the same email | Signup succeeds and joins invited workspace as pending; no extra workspace |
| Separate tenantless identity points at deleted workspace | Identity preserved, reference cleared; subsequent sign-in reaches choices |
| Expired invite, tampered intent, invite revoked after profile render | Explicit error; no accidental membership or workspace |
| Password recovery through invitation | Actual locally generated reset callback and final new-password submission succeed; subsequent sign-in retains invitation |
| Invalid reset token and invalid onboarding intent on recovery | Invalid-token retry retains valid intent; malformed intent does not block ordinary recovery |
| Five locales, narrow viewport, light/dark | Choice cards and shared Steps inspected; Join footer reachable, email wraps safely, no horizontal overflow |
| Inbox thread contact linking and unlinking | Actual modal actions persist; reopened settings retain link; unlink removes identifier |
| XLSX create and update for contacts, organizations, deals, services and tasks | All ten UI import runs succeed, one record per entity, updates do not duplicate records |
| Invalid service amount import | Row-level error shown; Add records disabled; no service created |
| Deal-specific and generic relation service quick-create | Both use amount 100; relations persist; ordinary imported amount 125.5 remains unchanged |
| Weighted deal and custom-field changes | 25% gives 50 on a 200 deal, 60% gives 120; unrelated field deletion preserves it; selected field deletion clears configuration and weighted value, preserving total 200 |

- Auth/onboarding/recovery ran in the cloud-mode dev server with local-only email capture. Under heavy host filesystem pressure, the protected inbox/operator/import/service/weighting tail also ran against the exact completed production build on localhost with outgoing email disabled. The dev server was restored afterwards.
- Final DB readbacks asserted imports, relation defaults, weighting cleanup, invitation memberships, no-write branches, operator cleanup and expected workspace counts.
- Restored the exact pre-test disposable database snapshot: 35 companies, 38 tenant users, 38 auth identities and zero new review fixtures. Before/after snapshots and screenshots remain local in `/tmp/pr154-consistency-e2e.hNBr7C`. Preview remains on port 4010.
- Earlier branch verification also covered shared primary/inverse controls and application-locale metadata boundaries. The shared primary foreground contrast is 4.92:1; its 97% hover remains at least 4.68:1.

Verification limits: real Google/Microsoft OAuth provider round trips were not performed; callback and continuation logic is covered by automated tests. Responsive checks used a browser viewport, not a physical device. No claim is made that every theoretical application scenario has been exhausted. Development-only warnings and automation/host timeouts were distinguished from completed UI results; the protected production-build run reported no browser errors.

## Impact and rollback

New users explicitly choose whether to create or join a workspace. Invitees retain their destination and join as Waiting for Approval; an administrator must assign both a role and active status. Existing workspace members are prompted to switch accounts.

The consistency changes preserve existing import targets, permissions, transactions, data shapes and the existing service default. They move ownership to the established architectural layers and add regression coverage.

No schema migration, dependency change, environment change or production backfill. Roll back by reverting this PR. Cleared references only point to deleted workspaces and do not require restoration. The owner alone merges and performs any production customer repair.

### DIFF
--- a/app/[locale]/(protected)/onboarding/wizard/actions.ts
+++ b/app/[locale]/(protected)/onboarding/wizard/actions.ts
@@ -1,61 +1,27 @@
 "use server";
 
-import type { RegisterUserData, RegistrationTarget } from "@/features/user/register/register-user.interactor";
+import type { RegisterOnboardingProfileData } from "@/features/user/register/register-onboarding-profile.interactor";
 
 import { refresh } from "next/cache";
 import { redirect } from "next/navigation";
 import { getLocale } from "next-intl/server";
 
-import { getAuthService, getCompleteOnboardingWizardInteractor, getRegisterUserInteractor } from "@/core/di";
+import { getCompleteOnboardingWizardInteractor, getRegisterOnboardingProfileInteractor } from "@/core/di";
 import { serializeResult } from "@/core/utils/action-result";
 import { isRedirect } from "@/features/auth/auth-outcome";
 import {
   clearRegisteredAdClicksFromCookie,
   readRegistrationAdAttribution,
 } from "@/features/acquisition/next/ad-attribution-cookie";
-import { resolveOnboardingIntent } from "@/features/company/next/onboarding-intent";
-import { clearInviteTokenCookie } from "@/features/company/next/invite-token-cookie";
-import { pathWithOnboardingIntent } from "@/features/company/onboarding-intent-url";
 import { buildLocalePath } from "@/i18n/locale-registry";
 
-export async function registerProfileAction(data: RegisterUserData, onboardingIntentValue?: string) {
+export async function registerProfileAction(data: RegisterOnboardingProfileData) {
   const adAttribution = await readRegistrationAdAttribution();
-  const onboardingIntent = await resolveOnboardingIntent(onboardingIntentValue);
-  if (onboardingIntent.status === "invalid") {
-    await clearInviteTokenCookie();
-    if (onboardingIntent.source === "explicit")
-      redirect(buildLocalePath(await getLocale(), `/auth/error?type=${onboardingIntent.errorMessage}`));
-  }
-
-  let target: RegistrationTarget = { type: "legacyAuthBinding" };
-  if (onboardingIntent.status === "valid" && onboardingIntent.type === "invitation")
-    target = { type: "invitation", companyId: onboardingIntent.companyId } as const;
-  if (onboardingIntent.status === "valid" && onboardingIntent.type === "createCompany") {
-    const session = await getAuthService().getSession();
-    if (!session)
-      redirect(buildLocalePath(await getLocale(), pathWithOnboardingIntent("/auth/signin", onboardingIntent.intent)));
-
-    if (onboardingIntent.authUserId !== session?.user.id) {
-      await clearInviteTokenCookie();
-      redirect(buildLocalePath(await getLocale(), "/auth/error?type=invalidOnboardingIntent"));
-    }
-    target = { type: "createCompany" } as const;
-  }
-  const registrationResult = await getRegisterUserInteractor().invoke(data, {
-    adAttribution,
-    target,
-  });
-  if (isRedirect(registrationResult) && onboardingIntent.status === "valid") {
-    if (registrationResult.redirect === "/auth/signin")
-      redirect(buildLocalePath(await getLocale(), pathWithOnboardingIntent("/auth/signin", onboardingIntent.intent)));
-
-    if (registrationResult.redirect === "/auth/signup" && onboardingIntent.type === "invitation")
-      redirect(buildLocalePath(await getLocale(), pathWithOnboardingIntent("/auth/signup", onboardingIntent.intent)));
-  }
+  const registrationResult = await getRegisterOnboardingProfileInteractor().invoke(data, { adAttribution });
+  if (isRedirect(registrationResult)) redirect(buildLocalePath(await getLocale(), registrationResult.redirect));
   const result = await serializeResult(registrationResult);
   if (result.ok) {
     await clearRegisteredAdClicksFromCookie();
-    await clearInviteTokenCookie();
     refresh();
     redirect(result.data.redirectTo);
   }
--- a/app/[locale]/(protected)/onboarding/wizard/actions.ts
+++ b/app/[locale]/(protected)/onboarding/wizard/actions.ts
@@ -1,25 +1,61 @@
 "use server";
 
-import type { RegisterUserData } from "@/features/user/register/register-user.interactor";
+import type { RegisterUserData, RegistrationTarget } from "@/features/user/register/register-user.interactor";
 
 import { refresh } from "next/cache";
-import { cookies } from "next/headers";
 import { redirect } from "next/navigation";
+import { getLocale } from "next-intl/server";
 
-import { getCompleteOnboardingWizardInteractor, getRegisterUserInteractor } from "@/core/di";
+import { getAuthService, getCompleteOnboardingWizardInteractor, getRegisterUserInteractor } from "@/core/di";
 import { serializeResult } from "@/core/utils/action-result";
+import { isRedirect } from "@/features/auth/auth-outcome";
 import {
   clearRegisteredAdClicksFromCookie,
   readRegistrationAdAttribution,
 } from "@/features/acquisition/next/ad-attribution-cookie";
+import { resolveOnboardingIntent } from "@/features/company/next/onboarding-intent";
+import { clearInviteTokenCookie } from "@/features/company/next/invite-token-cookie";
+import { pathWithOnboardingIntent } from "@/features/company/onboarding-intent-url";
+import { buildLocalePath } from "@/i18n/locale-registry";
 
-export async function registerProfileAction(data: RegisterUserData) {
+export async function registerProfileAction(data: RegisterUserData, onboardingIntentValue?: string) {
   const adAttribution = await readRegistrationAdAttribution();
-  const result = await serializeResult(getRegisterUserInteractor().invoke(data, { adAttribution }));
+  const onboardingIntent = await resolveOnboardingIntent(onboardingIntentValue);
+  if (onboardingIntent.status === "invalid") {
+    await clearInviteTokenCookie();
+    if (onboardingIntent.source === "explicit")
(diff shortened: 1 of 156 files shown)


app/[locale]/(protected)/onboarding/wizard/components/step-profile.store.ts
```diff
@@ -21,16 +23,22 @@ export class StepProfileStore extends BaseFormStore<RegisterUserData> {
...
+    this.onboardingIntent = onboardingIntent;
+  };
+
   onSubmit = async (event?: FormEvent<HTMLFormElement>) => {
     event?.preventDefault();
     this.setIsLoading(true);
 
     try {
-      const res = await registerProfileAction(toJS(this.form));
+      const res = await registerProfileAction(toJS(this.form), this.onboardingIntent);
```
Comment (benjiwagner) [inline app/[locale]/(protected)/onboarding/wizard/components/step-profile.store.ts]: why dont we make the onboarding intent part of the form data?

app/[locale]/(protected)/onboarding/wizard/actions.ts
```diff
@@ -1,25 +1,61 @@
...
   clearRegisteredAdClicksFromCookie,
   readRegistrationAdAttribution,
 } from "@/features/acquisition/next/ad-attribution-cookie";
+import { resolveOnboardingIntent } from "@/features/company/next/onboarding-intent";
+import { clearInviteTokenCookie } from "@/features/company/next/invite-token-cookie";
+import { pathWithOnboardingIntent } from "@/features/company/onboarding-intent-url";
+import { buildLocalePath } from "@/i18n/locale-registry";
 
-export async function registerProfileAction(data: RegisterUserData) {
+export async function registerProfileAction(data: RegisterUserData, onboardingIntentValue?: string) {
```
Comment (benjiwagner) [inline app/[locale]/(protected)/onboarding/wizard/actions.ts]: The business logic of the actions should always live within the interactors instead, like we defined in our knowledge base.

Comment (benjiwagner) [reply app/[locale]/(protected)/onboarding/wizard/actions.ts]: And also this is the same for the other action files that we have touchfile actions that we have added or changed in this change set.

features/user/prisma-user.repository.ts
```diff
@@ -512,13 +512,47 @@ export class PrismaUserRepo
...
+  async peekAuthUserCompanyIdUnscoped(userId: string) {
     const authUser = await this.prisma.authUser.findUnique({
       where: { id: userId },
       select: { companyId: true },
     });
+    return authUser?.companyId;
+  }
+
+  @BypassTenantGuard
+  async lockCompanyForRegistrationUnscoped(companyId: string) {
```
benjiwagner: This looks a bit weird because we have done this nowhere else previously. Also we have the transaction decorators, wrapper functions, why don't we use them instead? If there is specific reasoning behind it.

features/user/prisma-user.repository.ts
```diff
@@ -512,13 +512,47 @@ export class PrismaUserRepo
...
+    const company = await this.prisma.company.findUnique({
+      where: { id: authUser.companyId },
+      select: { id: true },
+    });
+
+    return company?.id ?? null;
+  }
+
+  @BypassTenantGuard
+  async lockAuthUserCompanyIdForRegistrationUnscoped(userId: string) {
```
benjiwagner: I don't really understand this method here as well also we don't usually use the query raw Prisma because usually we try to rely on the normal Prisma methods.

features/user/register/register-user.interactor.ts
```diff
@@ -52,18 +52,31 @@ const OutputSchema = z.object({
 });
 export type RegisterUserResult = Data<typeof OutputSchema>;
 
+const RegistrationTargetSchema = z.discriminatedUnion("type", [
+  z.object({ type: z.literal("createCompany") }),
+  z.object({ type: z.literal("invitation"), companyId: z.string().min(1) }),
+  z.object({ type: z.literal("legacyAuthBinding") }),
```
benjiwagner: Can you explain to me why we need this legacy off binding literal?

app/[locale]/(protected)/onboarding/wizard/components/step-profile.store.ts
Comment (benjiwagner) [reply app/[locale]/(protected)/onboarding/wizard/components/step-profile.store.ts]: Agreed. `onboardingIntent` now lives inside `RegisterOnboardingProfileData` and `StepProfileStore.form`. The store submits one form object, and a focused store test proves initialization and forwarding through the action boundary.

app/[locale]/(protected)/onboarding/wizard/actions.ts
Comment (benjiwagner) [reply app/[locale]/(protected)/onboarding/wizard/actions.ts]: Agreed. Intent validation, target selection, identity matching, redirect preservation, cookie cleanup, and workspace-choice routing now live in dedicated interactors. The wizard action only reads request attribution, invokes `RegisterOnboardingProfileInteractor`, serializes/localizes the result, refreshes, and redirects.

Comment (benjiwagner) [reply app/[locale]/(protected)/onboarding/wizard/actions.ts]: Applied across the other touched actions as well. Workspace choice now uses `ChooseWorkspaceOnboardingInteractor`; signup, password recovery, reset, and verification behavior lives in the corresponding auth interactors; invitation sign-out lives in `SignOutInteractor`. The actions retain framework-only adaptation.

features/user/prisma-user.repository.ts
benjiwagner: Good catch. I removed the custom company advisory-lock method. Registration now uses `runInTransaction(..., { companyId })`, which takes the same shared company lock as operator deletion, and checks company existence through `PrismaCompanyRepo` inside that transaction.

features/user/prisma-user.repository.ts
benjiwagner: The remaining raw query is limited to `SELECT ... FROM AuthUser ... FOR UPDATE`, now named `findAuthUserCompanyIdForUpdateUnscoped`. Prisma 7 has no pessimistic row-lock API. This lock serializes simultaneous create-workspace submissions before either can write, without a fake update that would mutate timestamps. The database race test now exercises that exact case and proves one workspace is created.

features/user/register/register-user.interactor.ts
benjiwagner: This is rollout compatibility, not a state created by the new flow. Before this PR, the Better Auth create hook could set `AuthUser.companyId` before tenant-user registration, so a user already mid-invitation at deploy can have a valid company binding but no `User` row. The fallback lets that user finish after locking and rechecking both records and never creates a workspace. I renamed it to `existingAuthUserCompanyBinding` so the literal describes the data we can prove rather than assuming its provenance.

PR #155, opened 2026-09-05
feat(data-view): save and group list views as tabs

benjiwagner:
## Summary

List views (contacts, organizations, deals, services, tasks, inbox, the company tables and the operator console) gain personal **saved views**: a named snapshot of filters, search, sort, page size, layout, grouping, column order, widths and hidden columns, shown as tabs attached to the top bar. Every change is saved into the active view as you make it; the built-in **All** tab keeps its own per-user state the same way. Views are private to the user who created them. Grouping widens from single-select custom columns to relations, users and date buckets, in the table as collapsible group rows and on the board as columns ordered by option index, backed by one grouped count query instead of three per group. The filter accordion is replaced by a command palette that supports several filters per field with the operator editable afterwards and an explicit remove control per applied filter. Saved filter presets are removed; a preset is recreated as a view in seconds.

Linear: [CUS-315](https://linear.app/customermates/issue/CUS-315). Folds in [CUS-248](https://linear.app/customermates/issue/CUS-248).

## Context

Customers asked for pipeline views without realising a grouped card view already was one, because grouping was two clicks into a display popover and nothing about the result could be named or kept. The first cut of this branch added Linear-style sharing with a per-user override layer, Reset and Save. After checking Linear's real behaviour live (it offers Personal / Workspace / Team at creation and keeps an ad-hoc layer with Clear | Save) and weighing it against the AI-assisted filter setup planned next, the owner chose the simpler model in the final commit of this PR: personal views that autosave, no filter row under the tabs, tabs styled like Linear's. The unreleased migration was rewritten in place so the override table and the visibility enum never exist.

## Validation

Run locally on the final head against disposable PostgreSQL 17 with `APP_MODE=cloud`, Node 24:

- Database rebuilt from the edited migration (`migrate deploy`, 43 migrations, `migrate diff` reports no drift against `schema.prisma`).
- `yarn typecheck`, `yarn lint` (max-warnings 0): clean. `yarn vitest run tests/conventions`: 660 passed. `RUN_DATABASE_TESTS=true yarn test`: 5421 passed, 0 failed (json reporter). `yarn openapi:generate`: no drift beyond the deliberate preset removal. `yarn build`: passed.
- In the running app as two seeded users: tabs render with exactly the four rail ids and no Reset, Save, picker or filter row; a filter added on the All tab survives a reload and lands in `P13n`; removing it through the palette persists `[]`; a filter added on a named view lands in that `DataView` row and survives a reload; create, rename and delete a view through the tab menu with its confirmation; Sofia's rail shows only `All` and Max's view id is unreadable by direct URL; the operator table renders the rail with a name-only create overlay. Each rework slice had two adversarial reviewers; confirmed findings carry named regression tests.
- Not verified on the Vercel preview behind sign-in.

## Impact and rollback

- BREAKING for API consumers: `savedFilterPresets` is removed from the GET response of every listed resource in `public/v1/openapi.json`.
- Migration `20260904120000_saved_data_views` creates `DataView`, adds `P13n.activeViewKey` and **drops `P13n.savedFilterPresets`** (contents deleted). Roll back by re-adding that column first, then dropping `activeViewKey` and `DataView`; the header documents the order. The two later migrations add `grouping` and its indexes.
- The branch's preview database ran an earlier version of the same migration folder (it still holds a `DataViewOverride` table and a `visibility` column the code no longer reads); reset that database before trusting the preview.
- A GET never persists; state becomes durable through the debounced autosave only. Demo mode: view writes stay blocked.

Knowledge-base companion: customermates/knowledge-base PR #117, to merge after this one.

---

### Update: Table and Board, operator parity, design alignment, dead-code sweep

The three layouts (Table / Cards / Kanban) are now two, **Table and Board**. The ungrouped card grid is deleted; Board is the card view with a grouping. Choosing Board with no grouping shows a prompt to pick a groupable field or **create a single-select custom column** on the spot (type preselected and locked, applied as the grouping, no record drawer). The **operator console tables** gain the same layout, sort, grouping and column options as the entity lists, grouping users, workspaces and audit entries by status, plan, subscription status and audit source through the operators' own intercepted filters. The view tabs are rebuilt on the design system (ghost-button link, `--selected` wash on the current tab held through hover, dashed `border-input` draft).

Dead and non-operational code left by the shifting requirements is removed: the data-view surface descriptor registry, the store `isKanbanMode`/`groupingColumnId` getters, the `KANBAN_*` aliases, the `viewUnavailable` result field and its toast, `GetResult.groupingColumnId` (dropped from `public/v1/openapi.json`, never populated in API mode), the legacy `sortField`/`sortDir` and JSON-`filters` URL decoders, and the `P13n.groupingColumnId` write input; the filter operator menu moved into the palette directory. A new additive migration `20260905120000_backfill_grouping_descriptor` lifts any legacy `card` + `groupingColumnId` row to a grouping descriptor so the shadow column is no longer read at query time.

Verified in the running app as two seeded users: the two-control segmented picker on every surface; the board prompt and create-single-select flow (a new column becomes the board's columns and autosaves into the personal All-tab state); operator users/workspaces grouped on a board and the audit log grouped as collapsible rows; text, relation, enum and date filter pages; clear-all; per-user isolation. Gates on the merged head: typecheck, `lint`, `tests/conventions`, `RUN_DATABASE_TESTS=true yarn test` (one known operator-count flake, green in isolation), `yarn build`, and an `openapi.json` diff of exactly the `groupingColumnId` removal.

Known cosmetic artifact: clearing a grouping stores `{}` rather than SQL `NULL` in the JSON column, because the tenant-guarded Prisma client serialises the `Prisma.DbNull` sentinel (a no-own-keys instance) to `{}`; both read back through `GroupingSchema` as "no grouping", so behaviour is identical. Not worth a 16-site sentinel refactor.

---

### Update: view-switch loading, Linear-style tabs, layout cards, group rows

Five owner requests, one of which was a real bug.

**The view-switch jump (bug).** `applyView` applied the incoming view optimistically and then refreshed through `refreshResolvedInBackground`, the one path that hard-coded the `"background"` mode, so `dataRequest` stayed `ready` and the surface rendered the previous rows under the incoming columns, grouping and layout for the whole round trip; a switch to a grouped board flashed `BoardGroupingPrompt` on the way. It now enters `refreshing` synchronously and resolves through a visible refresh (`isReady` guarding first load), and bumps `requestGeneration` so a request already in flight when the tab was clicked cannot land and revert the switch. The **All** tab was worse: it is never a chip, so `applyView` applied `{}` and reset filters, sort, grouping, layout and page size to surface defaults until the server re-resolved the stored state. Reads now carry that per-user All state next to the chips (`allState` on `DataViewResultFields` only, so `public/v1/openapi.json` is unchanged), and the store keeps it in `allViewState`. A response computed before a local write but landing after it no longer clobbers the All snapshot or a saved view's remembered state, which matters now that a switch no longer waits for the pending autosave to flush.

**Tabs.** 28px pills, fully rounded, 10px padding, 12px/500, every tab carrying a border and a resting fill so each reads as its own control, active taking `--selected` with a `--border-strong` outline; the new-view control is a 28px transparent round icon button. Measured against Linear live. Rail height stays 48px to match `MassActionsBar`, and the four `global-data-views*` ids, `aria-current` and the roving tabindex are unchanged.

**Layout picker.** The two segmented triggers render as preview cards, each a `LayoutIllustration` drawn from `div`/`span` and theme tokens above its label, the way the widget editor picks a chart style. Still Radix `Tabs`, so the ids, the disabled Board state, its tooltip and the agent-chat `data-state="active"` contract survive. `Common.table.layouts.{table,board}` added to all five bundles; `fr` no longer says "kanban" for the Board layout.

**Group rows.** The header row painted `--selected` rather than its intended muted wash, because `components/ui/table.tsx`'s `has-aria-expanded:bg-selected` won on an expanded group; overridden locally to a uniform faint wash at rest, expanded and hovered. The group checkbox moves ahead of the disclosure button, so it becomes the first flex child of the same `px-3` cell and lands in the same column as the header and record checkboxes with no negative margins; it gets its own `DataView.selectGroup` label.

Verified in the running app (Chromium, light and dark, 1440px and 375px): chips measured at 28px / `9999px` / 10px / 12px-500 with a border and fill on every tab in both themes; a tab switch shows the skeleton for the round trip with no stale rows and no grouping-prompt flash; hiding a column on All, switching to a saved view and back restores exactly that column set; the layout cards render on the deals popover, the mobile sheet and the operator console with the same ids; header, group and record checkboxes all at the same x with the group checkbox centred on the chevron; the operator board grouped by Plan. Gates on this head: `typecheck`, `lint`, `openapi:generate` (empty diff), `vitest run components/data-view core/base tests/conventions` (1277 passed), `RUN_DATABASE_TESTS=true yarn test` (5692 passed; the same known operator-count flake, 11/11 in isolation), `yarn build`.

Follow-up, deliberately not taken here: `has-aria-expanded:bg-selected` in the shared `TableRow` primitive now has no live consumer through the data-view group header, but table cell content comes from per-surface column definitions that may render other `aria-expanded` triggers, so deleting it from the primitive is an unaudited visual change across every table and belongs in its own change.

---

### Update: rail controls, filter palette legibility, appearance ordering

**Rail controls.** `#global-data-views-new` and `#global-data-views-menu` were transparent ghost buttons beside the new bordered pill tabs and read as floating glyphs. Both now take the inactive tab's resting surface, extracted into one exported `VIEW_SURFACE_CLASS` that the tab, the plus and the chevron compose from, so the three cannot drift. Measured live: both controls resolve byte-for-byte to an inactive tab's surface in each theme. This deliberately diverges from Linear, whose add-view control is transparent; the owner asked for the border explicitly.

**The view menu reaches the All tab.** It was gated on a saved view being active, so the All tab offered no way to keep what you were looking at. It now renders there with **Duplicate as new view** and **Copy link** only, built by a separate `allViewMenuItems`: All has no row to rename or delete and no position among the saved views. Duplicating it snapshots the current on-screen state rather than a stored payload, because `submitMeta` already falls back to `createViewFromCurrent` whenever the source view is not found and `ALL_VIEW_KEY` never is. The prefilled name reuses `DataView.views.duplicateName`, so it reads "All copy" in English.

**Filter palette.** The operator was a bare word with no border, fill or chevron, so it read as a caption rather than the control that changes the condition; it is now a field-variant dropdown, 30px and level with the back control, and its accessible name carries the operator (`Change condition: in`) instead of the purpose alone. Applied filters and the field list rendered as one undifferentiated run of rows; applied filters are now bordered 36px cards with their own radius, above plain 30px field rows, under 11px uppercase headings matching the section headings used elsewhere.

The applied card carries **no resting fill**, which is the one place the first implementation went wrong and three reviewers caught independently. Resting it on `--muted` left a selected-versus-resting step of 0.4 percentage points (contrast 1.009 against a field row's 1.129), so the keyboard cursor and hover were invisible on exactly the rows that most need clicking, and it sat a clickable row below its own surface, a treatment the design system reserves for disabled content. Leaving the whole `--selected` wash to the cursor is also the only option whose step is ground-independent: `--input-background` measured fine in the desktop popover but degraded in the mobile drawer, whose ground is `--background` rather than `--popover`. Measured after the fix, light and dark, popover and drawer: the applied card and a field row take identical washes on every surface.

**Appearance popover.** Group by moves above Sort by, the three `<Separator />` rules are gone, and the last section gets its own `Common.table.fields` label so the word "Appearance" no longer renders twice with nothing between it and the overlay title.

The palette regression tests now measure rather than pin strings: they parse `styles/globals.css`, resolve the `:root` and `.dark` token blocks through `var()` indirection, composite each row's resting and selected fill over the popover ground and compare sRGB contrast, so the test fails on the exact defect the reviewers found. Every fix was proved by planting the failure it catches and restoring.

Verified live at `localhost:4000` in both themes: the two rail controls measure identical to an inactive tab; the All tab's menu lists exactly Duplicate and Copy link; an applied filter row moves from transparent with a 13% rim to a 5.5% wash with a 22.4% rim under the cursor while a field row stays a plain 30px row; the operator trigger renders bordered with its chevron and announces its value; and the popover reads Layout, Group By, Sort by, Fields at 274px with zero separators.

Gates on this head, all green first time: `yarn typecheck`, `yarn lint`, `yarn openapi:generate` (empty diff), `yarn vitest run tests/conventions` (664), `yarn vitest run components/data-view core/base` (626), `RUN_DATABASE_TESTS=true yarn test` (5706 passed, 0 failed, 42 database suites executed and the known operator-count flake passed too), and `yarn build`.

---

### Update: tab surface, soft-primary active state, New view pill

**The tabs take the header buttons' surface.** They sat on a `bg-muted/50` wash while `#deals-filter` and its neighbours are `variant="secondary"`, white with `shadow-xs`, so the two rows read as different materials. The shared surface constant now carries that background and shadow; the border already matched at `rgba(0,0,0,0.075)`. Resting tabs keep `text-muted-foreground`, which is what lets the active one carry weight. Measured live: a resting tab and a header button are now identical on background, border and shadow in both themes.

**The active tab carries primary softly**, replacing the neutral `--selected` wash: `border-primary/40 bg-primary/20` with the label in a new `--primary-soft-foreground`. Plain `text-primary` on that tint misses the WCAG AA body-text floor in **both** themes, at 4.09:1 light and 3.95:1 dark, so the token is derived per theme from that theme's own `--primary` and `--foreground` (85/15 in oklab) and clears the floor at 5.62:1 and 5.01:1 while keeping the hue. A solid `bg-primary` fill was considered and rejected, since `#deals-add` is already solid primary in the same region.

Two composition traps worth recording, both caught by resolving the real `cn()` rather than reasoning about it. tailwind-merge groups by modifier, so `text-primary-soft-foreground` displaces neither `hover:bg-accent` nor `hover:text-foreground` from the surface: the active state needs both hover utilities restated or a hovered active tab reverts to accent fill and foreground text. And the existing `toContain("text-foreground")` assertion would have passed green through the substring in `hover:text-foreground`, so it is replaced by an explicit `not.toContain("hover:text-foreground")`.

**The create control is the dashed New view pill**, not a round plus. `DataView.views.createTitle` is already "New view" and the draft ghost rendered exactly that string in exactly that pill, so building the trigger naively would have shown two identical dashed pills side by side while the overlay is open. The trigger now carries `data-view-draft` itself and the separate ghost is deleted; verified live that the marker is absent at rest, present exactly once while drafting on the trigger element, and absent again after cancel. Its visible text is its accessible name, so the duplicate `aria-label` is gone, and the pill opts out of the surface's new shadow with `shadow-none`. The view-actions control keeps its chevron.

Gates on this head, all green: `yarn typecheck`, `yarn lint`, `yarn openapi:generate` (empty diff), `yarn vitest run tests/conventions` (664), `yarn vitest run components/data-view core/base` (630), `RUN_DATABASE_TESTS=true yarn test` (5710 passed, 0 failed, 43 database suites executed), and `yarn build` in a copy-on-write clone so the running dev server was untouched. Both adversarial reviewers returned zero findings.

---

### Update: filter chips, rail rhythm, inbox parity

**Applied filters are chips again.** Each took a full-width row, so a short filter claimed the whole palette width and the zone read as more of the field list below it. They return to the chip the filter bar used before this branch: an interactive badge that sizes to its content, wraps, and packs left, rendered outside the command list so that list is purely the fields you can add. The chip keeps the badge's primary tint but takes `--primary-soft-foreground` for its label, because `bg-primary/20` with plain `text-primary` at eleven pixels misses the AA body-text floor, the same defect fixed on the active tab. Measured live: a chip is 113x22 sized to content against a 580x32 field row.

**The rail sits flush under the top bar.** It carried ten pixels of top padding while the entity detail summary rail sits flush at the header edge; it now uses that rail's rhythm. Because the nav centred its children, zeroing the top padding alone would have dropped the actions control eight pixels below the tabs, so the nav starts its controls at the top edge instead. Measured after: tab top and header bottom both 64, actions control top also 64.

**The inbox joins the top bar like every other list.** It mounted the rail inside its 380px thread column and explicitly opted out of `joinsTopBar`, making it the one surface whose views did not span the page. A small client surface now owns the rail and the view sync above both columns, and `InboxList` is just the list. Measured at 1440px: the rail is 1174px across the content area where the thread column is 379px. This reverses a deliberate decision that `inbox-page-views.test.ts` pinned, at the owner's request; the test now pins the new intent.

Gates on this head: `yarn typecheck`, `yarn lint`, `yarn openapi:generate` (empty diff), 1389 tests across `components/data-view`, `core/base`, `tests/conventions` and the inbox, `RUN_DATABASE_TESTS=true yarn test` (6048 passed, 0 failed), and `yarn build` in a copy-on-write clone. Each new assertion was proved load bearing by planting the failure it catches.

Also in this range: a merge of `origin/main`, which brought the durable message drafts work. Its new `FilterFieldKey.draft` had to be declared in the three filter registries this branch made total over `FilterFieldKey` as part of CUS-248, and `@unipile/sdk` moved 2.32.0 to 2.41.0, so a `yarn install` is needed after pulling.

---

### Update: the joined top-bar boundary, and demo pipelines in the seed

**The divider above the view tabs is gone.** The top bar dropped its bottom border through a `useLayoutEffect`, which never runs on the server, so the server markup shipped the header border and the joined strip's border together and the reader saw two dividers until hydration removed one. The record page's pinned-field strip had the same flash. Every strip that joins the top bar now marks itself with `data-joins-top-bar` and the shell drops the header border in CSS, so the decision travels in the markup. The context flag, its hook and the prop that carried it are deleted, which is a net 88 lines. Measured: the deals list and a deal record compute a zero header border with the strip owning the single boundary, while the dashboard, which joins nothing, keeps its own.

**The demo now opens on pipelines.** The workspace could always run one, by grouping a list on a single-select column, but the demo opened on a flat table with a single saved view, which is why the "you have no pipelines" feedback kept arriving. The seed gives the demo user two views on each of the six workspace surfaces, so the tab row is itself the demonstration:

| Surface | Board | Filtered table |
| --- | --- | --- |
| Contacts | Lead pipeline, grouped on sales pipeline | In play |
| Organizations | Accounts by type | Direct customers |
| Deals | Sales pipeline, grouped on status | Open deals (existing) |
| Services | Catalogue by pricing | Hardware |
| Tasks | Delivery board, grouped on status | High priority |
| Inbox | none, the surface declares no groupable field | Drafts, Unread |

Verified against the seeded database rather than from the fixtures: the contact lead pipeline fills six stages (10/4/4/4/4/4), deals fill four with per-stage counts, totals and weighted forecast, tasks fill four of six, organizations two, services three, and the inbox narrows 25 threads to 1 draft and 11 unread. The single draft comes from the messaging seed on main; more drafts there would make that view richer.

Gates: `yarn typecheck`, `yarn lint`, `RUN_DATABASE_TESTS=true yarn test` (6048 passed, 0 failed), and the affected suites. Each new assertion was proved load bearing by planting the failure it catches.

---

### Final end-to-end pass

Driven against a freshly reset and reseeded local database, on `40f1d516`.

**Journeys, all green.** View lifecycle on deals: create from the New view pill, rename through Edit (prefilled), duplicate (prefilled "… copy"), Move left reorders, Copy link, and Delete confirms by name then falls back to All. Filter palette: pick a field, read the bordered operator control (`aria-label="Change condition: in"`), pick a value, return to root and see the applied chip render as a 108px badge against a 342px field row, then remove it with its X. Appearance popover: Layout, Group By, Sort by, Fields at 288px with zero separators, two 84px illustrated layout cards, layout switch board to table (10 columns, 10 rows, 4 collapsible group rows) and a column toggle 10 to 9 to 10. Group rows: a 2% wash, all three checkboxes at the same x, the group checkbox centred on its chevron, `Select group Open`. Inbox: rail spans the page outside the thread column and filters 25 threads to 1 draft. Operator console: same rail, same section order, same layout cards, no separators. Both themes and 375px checked; the header boundary computes zero wherever a strip joins it.

**One defect found and fixed.** Every board card hydrated with a mismatched `aria-describedby`: the server rendered `DndDescribedBy-0`, the client `-1`, `-2` or `-3`, because dnd-kit names its context from a module counter that restarts in the server process. React reported it and said it would not patch it up, so each draggable card's accessible description pointed at the wrong node for the life of the page. The context now carries a `useId`, and the server markup ships one stable id with no counter value anywhere. It predates this branch, since the board never passed an id, but the saved-view work and the demo seed put a board on screen far more often. Covered by a test that fails when the id is removed.

**One false alarm worth recording**, so the next reader does not chase it: `tests/conventions/runtime-console.test.ts` failed once during this pass on `chk2.tmp.ts`, a scratch script I had created in the repository root while investigating. It is the convention test doing its job on a file that no longer exists; the tree is clean and the test passes.

**One thing checked and cleared.** Several `P13n.activeViewKey` rows appeared after the reset without an obvious cause, which would have been serious if a read were persisting state. It is not: clearing every key and then visiting contacts, organizations and tasks without touching a tab left all sixteen rows null. The values came from my own tab clicks during verification.

Gates on this head: `yarn typecheck`, `yarn lint`, `yarn openapi:generate` (empty diff), 1737 tests across `components/data-view`, `core/base`, `tests/conventions`, `app/components` and `components/entity-detail`, `RUN_DATABASE_TESTS=true yarn test` (6049 passed, 0 failed), and `yarn build`.


core/base/grouping/clear-grouping.ts
```diff
@@ -0,0 +1,20 @@
+import type { AppPrismaClient } from "@/prisma/db";
+
+import { Prisma } from "@/generated/prisma";
+
+export type ClearGroupingArgs = {
+  columnId: string;
+  companyId: string;
+};
+
+export async function clearGroupingForDeletedColumn(
```
benjiwagner, line 10: This looks quite complicated. Why don't we inline this in the respective occurences

PR #157, opened 2026-09-06
feat(agent-chat): show live waiting feedback and normalize durable tool input

benjiwagner:
## Summary

Give the AI chat honest, live feedback while the model is thinking, and repair a durable-workflow defect that silently stripped tool input validation and schema defaults. Three related changes, developed and verified together:

1. **Live waiting feedback.** From send until the first token the chat showed only animated dots, which reads as a hang on long questions that call no tools. It now shows a truthful phase label driven by real workflow lifecycle events (`Starting`, `Working on your request`, `Preparing the next action`), plus elapsed seconds once a wait passes five. There are no fabricated stages, no fake percentages, no raw reasoning, and no extra model call to narrate progress. The `progress` event is deliberately non-durable: it never persists as an assistant message, never enters a later prompt, and is dropped once the turn is completed, failed or stop-requested.

2. **Durable tool input normalization.** Exporting a tool to the durable workflow reconstructed its JSON schema with `jsonSchema(shell.inputSchema)`, which drops the runtime validator that `providerSafeSchema` attaches. JSON Schema defaults are annotations, not applied values, so Zod defaults, coercions, transforms and refinements never ran. A valid model call that omitted an optional field (for example `mcpPage`, which carries `.default(1)`) reached stricter downstream code as `undefined` and failed validation. `normalizeAgentAiToolInput` now runs the authoritative validator, and a per-`toolCallId` resolver guarantees that the approval check and the execution see the same normalized value, computed exactly once. Invalid input is settled as a tool failure and is neither approved nor executed. A reused `toolCallId` carrying different input is refused rather than executed twice.

3. **Interrupted turn reconciliation.** A failed workflow could leave the chat stuck on "Reconnecting" or "Stopping" with a credit reservation still held. The workflow catch block now reconciles every failure rather than only the first, `recordAgentTurnExternalRun` is fenced on `runId` plus a null `externalRunId`, reservation release is guarded on `providerStartedAt: null`, and the user sees an explicit `role="alert"` notice telling them changes may already have been applied.

Every production gateway round now also asserts `zeroDataRetention: true` and `disallowPromptTraining: true`, pinned by `tests/conventions/agent-model-boundary.test.ts`.

## Context

Linear: CUS-314.

The tool-input defect was found while benchmarking the agent against synthetic CRM fixtures: identical prompts failed on some rounds with pagination validation errors that the models had not caused. Preserved execution outputs confirmed the mechanism. It is not a pagination bug. It affects any tool whose schema relies on defaults, coercions, transforms or refinements, including write tools, which is why the repair lands at the dispatch boundary rather than by hardcoding a default.

The waiting-feedback work came from the same observation: on questions that call no tools the interface gave no signal at all for the entire model round.

The ZDR and no-prompt-training flags began as a requirement of that benchmark and are kept here deliberately. They move the privacy posture from "the team-wide account setting is enabled" to "every request asserts it", which is a stronger and more durable claim. Reviewers should note this constrains provider routing: a serving provider that cannot honour both flags will be refused rather than silently downgraded.

## Validation

Run against this branch rebased onto `main`, in a clean worktree with no benchmark instrumentation present:

- `yarn typecheck` passed.
- `yarn lint` passed with `--max-warnings 0` in 58.67s.
- `yarn i18n:audit` reported advisory findings only, none of them from these keys.
- `yarn test` with `RUN_DATABASE_TESTS=true` against local PostgreSQL: 5287 passed, 3 skipped, across 593 files.
- `yarn build` passed: compiled successfully in 45s, TypeScript and route generation clean, public OpenAPI unchanged.

All five locales carry the five new keys with full parity and real translations. New and extended test coverage includes: normalize-once across the approval and execution paths, `toolCallId` identity conflict refusal, panel tool refinements and transforms before command emission, bounded validation failure that executes no mutation, a real-tenant interactor read that succeeds without pagination errors, nine database-level interruption and compensation cases, and store coverage for the uncertain-turn notice and replay.

Behaviour was additionally exercised end to end during the benchmark that produced this work: 72 local episodes across five model configurations over 12 synthetic CRM scenarios, on synthetic fixtures only, with no real records touched and no outbound messages sent. First visible feedback measured at a median of 6.6 ms across 63 samples.

## Impact and rollback

User-visible: the chat replaces bare dots with a phase label and, after five seconds, an elapsed counter; an interrupted turn now shows an explicit notice instead of appearing to hang.

Behavioural: tool calls are validated and normalized before approval and execution, so calls that previously failed downstream now succeed with their schema defaults applied, and malformed calls now fail cleanly without executing. Every gateway round asserts zero data retention and no prompt training.

Risk: the gateway flag change is the only item that alters outbound provider behaviour. If a serving provider cannot honour both flags the round is refused rather than silently downgraded, which is the intended failure mode but would surface as an agent turn failure rather than a degraded answer.

Rollback: revert the merge commit. The change is additive, adds no migration, and touches no schema. Reverting restores the previous `gateway: { only: [servingProvider] }` shape, the bare typing dots, and the unnormalized dispatch path.

### DIFF
--- a/app/components/agent-chat/__tests__/agent-chat-interrupted-notice.test.ts
+++ b/app/components/agent-chat/__tests__/agent-chat-interrupted-notice.test.ts
@@ -1,0 +1,61 @@
+import { createElement } from "react";
+import { renderToStaticMarkup } from "react-dom/server";
+import { createTranslator } from "next-intl";
+import type * as IntlModule from "next-intl";
+import { describe, expect, it, vi } from "vitest";
+
+import de from "@/i18n/locales/de.json";
+import en from "@/i18n/locales/en.json";
+import es from "@/i18n/locales/es.json";
+import fr from "@/i18n/locales/fr.json";
+import itLocale from "@/i18n/locales/it.json";
+
+const context = vi.hoisted(() => ({ translator: (key: string) => key }));
+
+vi.mock("next-intl", async (importOriginal) => ({
+  ...(await importOriginal<typeof IntlModule>()),
+  useTranslations: () => context.translator,
+}));
+vi.mock("@/core/stores/root-store.provider", () => ({
+  useRootStore: () => ({ agentChatStore: {} }),
+}));
+vi.mock("@/core/utils/use-copy-to-clipboard", () => ({
+  useCopyToClipboard: () => vi.fn(),
+}));
+vi.mock("@/components/entity-terminology/use-entity-terminology", () => ({
+  useEntityTerminology: () => ({ plural: (entity: string) => entity }),
+}));
+vi.mock("@/components/ai-elements/message", () => ({
+  MessageResponse: () => null,
+}));
+
+import { AgentChatItemView } from "../agent-chat-items";
+
+describe("Interrupted agent turn notice", () => {
+  it.each([
+    ["de", de],
+    ["en", en],
+    ["es", es],
+    ["fr", fr],
+    ["it", itLocale],
+  ] as const)("renders a localized alert without a retry action in %s", (locale, messages) => {
+    const translator = createTranslator({ locale, messages });
+    context.translator = translator as unknown as (key: string) => string;
+    const markup = renderToStaticMarkup(
+      createElement(AgentChatItemView, {
+        item: {
+          kind: "turn_interrupted",
+          id: "notice-1",
+          messageId: "user-message-1",
+        },
+      }),
+    );
+
+    expect(markup).toContain('role="alert"');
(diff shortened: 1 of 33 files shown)


benjiwagner:
Folded into #145 as commit 24096e05, per the decision to review the whole agent-chat change as one pull request. The waiting-feedback and durable tool-input work is unchanged; its render site moved into the routines refactor's AgentConversationLog component and the progress indicator was ported across, which a clean merge would otherwise have dropped silently. Branch feat/agent-chat-progress-feedback is left in place for reference.

PR #166, opened 2026-09-08
fix: harden production error paths

benjiwagner:
## Summary

- Defer ambiguous email-delete burst events before calling the provider, preserving adopted rows until durable retry can reconcile them.
- Treat only exact retryable Unipile webhook failures (429 and `500 api/internal_error`) as durable retries, suppress intermediate noise, and emit one Sentry event when the tenth attempt exhausts.
- Give operator workspace deletion the existing bulk-write transaction budget while keeping the operation atomic.
- Return a dedicated, localized privacy-acknowledgement validation error from the contact form.

## Context

Tracks [CUS-328](https://linear.app/customermates/issue/CUS-328/harden-actionable-production-sentry-error-paths) and addresses Sentry groups CUSTOMERMATES-67, CUSTOMERMATES-5K, CUSTOMERMATES-42, CUSTOMERMATES-66, and CUSTOMERMATES-5P. CUSTOMERMATES-J is the client mirror of the same failed workspace-deletion request as CUSTOMERMATES-66, so this change keeps that client signal intact for correlation. It intentionally does not add broad client, transport, or browser-error suppression.

PR #145 also edits `core/validation/validation.types.ts`; this PR adds one enum value there, which is the only file overlap.

## Validation

- `yarn lint`
- `yarn typecheck`
- `RUN_DATABASE_TESTS=true yarn test` — 683 files passed, 1 skipped; 6,169 tests passed, 2 skipped
- `yarn build`
- Focused durable verification — 52 relevant tests passed, including the full operator-deletion database suite
- Real local PostgreSQL 17 retry test — 11 concurrent failure writes returned every atomic attempt count from 1 through 11; the row remained retryable with its error intact
- Real local PostgreSQL 17 timeout A/B — the same workspace with 1,000 child rows was blocked on its exact company advisory lock in both runs. The former 5,000 ms budget failed atomically after 5,532 ms with all data retained and no audit; the real current path succeeded after 5,642 ms with all target rows removed and exactly one audit
- Local browser E2E on `/de/contact` — unchecked privacy submission returned HTTP 200 with the exact German field and toast message, no Next.js error dialog, and no console errors
- Fresh Vercel preview for `ec2e879a` — `/de/contact` returned HTTP 200 and contained the expected German heading and privacy text
- Fresh GitHub checks — CI, PostgreSQL 16, rendered-page E2E, CodeQL, repository policy, and Vercel all passed
- Independent webhook, operator, and holistic reviews; no unresolved findings remain

## Impact and rollback

No schema or data migration. Webhook events remain durable and nonterminal for the existing hourly reprocessor; intermediate expected failures stay quiet, while final retry exhaustion is reported once. Workspace deletion remains atomic, and ordinary non-burst email deletions retain their existing behavior. Roll back by reverting this PR.

### DIFF
--- a/core/validation/validation.types.ts
+++ b/core/validation/validation.types.ts
@@ -32,6 +32,7 @@ export enum CustomErrorCode {
   emailLinkColourInvalid = "emailLinkColourInvalid",
   legalNoticeNotDelivered = "legalNoticeNotDelivered",
   termsNotAgreed = "termsNotAgreed",
+  privacyAcknowledgementRequired = "privacyAcknowledgementRequired",
   customColumnNotFound = "customColumnNotFound",
   customColumnIdNotFound = "customColumnIdNotFound",
   customColumnTypeMismatch = "customColumnTypeMismatch",
--- a/ee/messaging/ingest/reprocess-stuck-webhook-events.interactor.ts
+++ b/ee/messaging/ingest/reprocess-stuck-webhook-events.interactor.ts
@@ -1,11 +1,10 @@
-import type { WebhookEventRepo } from "../webhooks/webhook-event.repo";
+import { WEBHOOK_REPROCESS_MAX_ATTEMPTS, type WebhookEventRepo } from "../webhooks/webhook-event.repo";
 import type { ProcessUnipileWebhookInteractor } from "../webhooks/process-unipile-webhook.interactor";
 
 import { SystemInteractor } from "@/core/decorators/system-interactor.decorator";
 
 const REPROCESS_MIN_AGE_MS = 5 * 60 * 1000;
 const REPROCESS_MAX_AGE_DAYS = 30;
-const REPROCESS_MAX_ATTEMPTS = 10;
 const REPROCESS_BATCH_LIMIT = 500;
 
 @SystemInteractor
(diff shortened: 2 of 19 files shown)


benjiwagner:
Live provider E2E is now complete against this exact head (`ec2e879a7182288fcafb42c282d5e21e3fa71790`):

- Provisioned two controlled IMAP accounts through structured Unipile Custom Auth; both reached running/unlocked state, signed `account.add` ingestion completed, folders/backfill settled, and the authenticated local Connected Accounts UI showed both accounts.
- Scoped one V2 webhook endpoint to exactly those accounts and proved the path gate: wrong route/method 404, unsigned request 401, oversized request 413, valid signed request 200 with a processed durable event.
- Proved SMTP -> receiver IMAP -> signed `email.new` webhook -> local message -> authenticated local Inbox UI.
- Deterministically seeded the `>10` delete burst with eleven signed no-op events on the real fixture account, then performed one real external IMAP Inbox-to-Archive move. After the eight-second grace, the real event remained nonterminal at attempt 1 and the original local row was unchanged.
- Waited the real five-minute eligibility age and invoked the local cron route. The event processed with its error fields cleared, while the same local message/thread/RFC Message-ID reconciled to the provider's new message id and Archive folder.
- Performed a fresh real Inbox-to-Trash IMAP move after the burst window. Its event processed on attempt 0 and only the target local row was removed.
- Removed all 13 synthetic provider messages and matching local fixture rows. The two reusable controlled accounts remain; the exact-scope endpoint record retains its email events but is disabled, and the ephemeral app/gate/tunnel were stopped.

The durable setup and regression contract is in [knowledge-base PR #133](https://github.com/customermates/knowledge-base/pull/133), which should merge after this product PR. No credential values, controlled addresses, live provider ids, or ephemeral URL were recorded.

PR #172, opened 2026-09-13
feat(routines): audit-log routine create, edit, pause and delete

benjiwagner:
## Summary

Adds `routine.created`, `routine.updated` and `routine.deleted` domain events and emits them from the user-driven routine write paths, so routine configuration changes appear in the company audit log the way role, webhook and custom-column changes already do.

Routines shipped in #145 with no audit coverage at all: there was no `ROUTINE_*` member in `DomainEvent`, and none of the routine interactors imported `EventService`. Since `EventService.publish` is the only production writer of `AuditLog`, no routine create, edit, delete or administrator pause left a trace anywhere in the workspace audit trail.

## Context

`features/role/upsert-role.interactor.ts` and `delete-role.interactor.ts` are the golden reference and this change follows them exactly: create publishes the DTO, update publishes `{ routine, changes }` with `calculateChanges`, delete publishes the deleted DTO.

Covered write paths (all of them route through these three interactors, so the UI server actions, the `manage_routines` MCP tool and the authenticated MCP-over-HTTP endpoint are all covered by the same change):

| Path | Event |
| --- | --- |
| `UpsertRoutineInteractor` (create) | `routine.created` |
| `UpsertRoutineInteractor` (edit) | `routine.updated` with the change set |
| `PauseRoutineInteractor` (admin pause) | `routine.updated` with the change set |
| `DeleteRoutineInteractor` | `routine.deleted` |

Deliberately **not** emitting, and why:

- `ReconcileRoutineRunsInteractor` (auto-disable after three consecutive failures) is `@SystemInteractor`, which wraps `invoke` in `runWithoutTenant`. `EventService` extends `UserAccessor`, whose `this.user` throws with no tenant, and `AuditLog.userId` is non-nullable so `createAuditLog` drops a null-actor row (`features/event/event.service.ts:114`). There is no actor to attribute a cron action to.
- The three actorless callers of `ReleaseOwnerRoutinesInteractor`: the operator console (`ee/operator/update-operator-user-status.interactor.ts:35`) acts cross-tenant, and the two lifecycle sweeps (`ee/lifecycle/deactivate-trial-users-and-send-notice.interactor.ts:36`, `deactivate-users-after-subscription-grace-period.interactor.ts:32`) run with no actor at all. Same reason.
- `RunRoutineNowInteractor`: a manual run is already recorded in `RoutineRun` with `executedByUserId`.
- The routine cascade when a teammate is deactivated, for the reason below.

### The release cascade, deliberately not audited

Deactivating a teammate disables every routine they own. I built the audit for that cascade, following `delete-organization.interactor.ts`, which publishes an event per affected record. A database test against the real deactivation path showed it was the wrong thing to ship, so it is reverted.

`routineDto` (`ee/routines/prisma-routine.repository.ts`) derives presentation state: when a routine's owner is not active it reports `enabled: false` and `disabledReason: "ownerUnavailable"` whatever the stored columns say. The admin flow sets the user inactive *before* the release runs, so by the time the routines are disabled the DTO already presented them as disabled, and the only remaining difference was `nextRunAt`. Per-routine rows would have recorded a meaningless timestamp diff while implying a state change that, in the product's own model, had already happened.

The cause is audited: `user.updated` carries the status change, and a routine's disabled state is a function of it. `features/user/__tests__/admin-deactivation-routine-audit.database.test.ts` pins that contract against a real database, so the decision is enforced rather than remembered.

### Audit change-field labels

The activity timeline renders each changed field through `useCanonicalColumnLabel`, which falls back to a humanized English key. Ten routine fields resolved in neither catalog, so a German user saw `Name · Owner · Prompt +7`, and "Prompt" contradicted the form, which calls that field "Instructions". `AuditLogModal.fields` now carries all ten in five locales, reusing each locale's own wording (`Anweisungen`, `Auslöser`, `Inhaber`), and `AUDIT_FIELD_KEYS` in `tests/conventions/i18n-key-resolution.test.ts` grows to match. It now reads `Name · Inhaber · Anweisungen +7`.

No webhook surface is added. `WEBHOOK_EVENTS` is a separate hand-maintained catalog and routine events are intentionally absent from it, which also means a routine edit cannot trigger a routine: `createRoutineRuns` returns before any routine lookup for an event outside `WebhookEventSchema.options`.

The events map to `null` in `AUDIT_EVENT_ENTITY_TYPE` because routines have no `EntityType`. That matches role, webhook and custom column: the rows are persisted and visible on the company audit-log page and in the dashboard activity feed, but never on a record detail timeline.

No migration. `AuditLog` is generic (`event String`, `eventData Json`, `entityId`).

## Validation

Run in an isolated worktree against a disposable local PostgreSQL 17 container (`yarn db:provision`, port 38589), `APP_MODE=cloud`, with the synthetic seed applied.

- `yarn typecheck` — pass
- `yarn vitest run tests/conventions` — 85 files, 699 tests pass
- `RUN_DATABASE_TESTS=true npx vitest run --maxWorkers=4` — **730 files passed, 1 skipped; 6787 tests passed, 2 skipped; exit 0**
- `yarn build` — pass
- `yarn openapi:generate && yarn raw-docs:generate` — tree stays clean, no spec drift

### Exercised in the running app

Signed in as a seeded system-role admin and driven through the real forms and confirmation dialogs, not through interactors: edited a routine, deleted one, created one, and paused another owner's routine as administrator. All four wrote the expected rows, attributed to the acting user, with exact change sets (`name` alone for the rename; `enabled` plus `disabledReason` for the pause) and no spurious `nextRunAt` or `updatedAt` churn.

Verified rendering on every surface: the company audit log page, the detail modal, the event filter (all three options present, derived from the enum), the dashboard activity timeline, and the whole set again in German. The server's own event log printed `[event] routine.created listeners=0 webhooks=0 routines=0`, which is the no-fan-out and no-self-triggering claim confirmed at runtime. Hitting the per-user routine cap produced a refusal that published nothing.

### Tests

- `ee/routines/__tests__/routine-audit-events.test.ts` — 12 tests. Each event and payload shape, the prompt diff surviving into the change set, the empty change set on a no-op save, and four negative cases that must publish nothing.
- `ee/routines/__tests__/routine-audit-log.database.test.ts` — 5 tests through DI against real PostgreSQL, asserting actual `AuditLog` rows, including that the `routine.deleted` row outlives the routine and its cascade-deleted run history.
- `features/mcp-tools/__tests__/routine-mcp.database.test.ts` — one added test drives create, update and delete through `manageRoutinesTool.execute` and asserts the three audit rows. The MCP surface is now proven by execution rather than inferred from shared wiring.
- `features/user/__tests__/admin-deactivation-routine-audit.database.test.ts` — one added test driving a real deactivation against PostgreSQL: the routines end up disabled with `ownerUnavailable`, and the audit trail holds exactly the `user.updated` row.

The emitting behaviour was proven to discriminate by temporarily removing the emit and watching the relevant test fail.

Existing test changes are constructor-arity fixes plus one added repo mock method. No assertion was loosened or removed.

## Impact and rollback

User-visible: routine changes appear in Company → Audit logs, in the detail modal and in the dashboard activity timeline, fully localized in all five application locales including the changed-field labels. The event filter picks them up automatically because its options derive from the `DomainEvent` enum. Deactivating a teammate is unchanged: it disables that teammate's routines and records the cause as `user.updated`.

`AuditLog.eventData` now stores the full `RoutineDto`, including the routine prompt (up to 5000 characters), for every routine change. This matches how role and custom-column events store their full DTO, and the prompt diff is the point of the audit row, but it is worth noting that `AuditLog` has no retention path.

Rollback: revert the commit. No migration, no data backfill, no environment change. Audit rows already written remain readable; their labels resolve from the locale bundles, which the revert would also remove, so a reverted deployment would render the raw key for any row written in the interim.


benjiwagner:

PR #171, opened 2026-09-13
feat(agent-chat): open records on their page through navigate

benjiwagner:
## Summary

Removes the hosted assistant's `open_record` interface tool and lets `navigate` open one existing record's page by `entity` and `recordId`, always at `/{type}/{id}` and never through the `?open=` drawer deep link. The drawer is now reached only through the list page's Add control, which the assistant highlights.

- `navigate` accepts either a routable `targetId` or `entity` plus `recordId` (UUID); the runtime validator refuses `new`, paths, URLs, unknown entities and mixed forms, and the client store resolves the page path itself from its own entity-to-segment map, so no free-form path ever crosses the panel channel.
- `open_record` is gone from the tool catalog, the panel command names, the respond-to-command interactor, the client stores, the activity mapper and the system prompt; the prompt now states that records never open in the drawer.
- Docs (EN and DE): CRM Records, Dashboard, Inbox, Profile, Search and Assistant pages describe the detail URL for existing records, the `?open={entityType}:new` deep link for adding only, and search results opening the detail page (which is what the code does).
- `scripts/agent-eval.ts` gains the scenario "open the deal Nova Expansion", asserting exactly one `navigate` command with `{ entity: "deal", recordId }` and no `?open=` anywhere in the stream.

## Context

Owner decision from the Mate alignment audit: the assistant only calls MCP tools or navigates, highlights and tours the interface; existing records are never opened in the left-hand drawer, which the product's own click paths already route to the detail page. Stacked on #170 (benchmark harness).

## Validation

- `yarn typecheck`, `yarn lint`, `yarn vitest run tests/conventions`, full `yarn test` with `RUN_DATABASE_TESTS=true` on the sandbox database, `yarn build`: green locally.
- New and updated unit tests: navigate validator (record form accepted, drawer/path/URL/mixed forms refused, `open_record` absent), client store (record opens at `/deals/<id>`, no `?open=` for any entity, blocked navigation propagated), panel command mapping, Google-wire schema rows and rewrite counts (one tool fewer).
- The paid `agent:eval` scenario is written but not yet executed here: the sandbox has no `AI_GATEWAY_API_KEY`.

## Impact and rollback

Conversations whose history contains an `open_record` activity still render through the `interface.navigate` mapping. Revert the commit to restore the tool; no data or schema changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #170, opened 2026-09-13
feat(agent-benchmark): add the local quality, cost and latency harness

benjiwagner:
## Summary

Adds `scripts/agent-benchmark/`, a local-only quality, cost and latency benchmark for the hosted Assistant, plus the minimal runtime extensions it needs: an opt-in benchmark model overlay (`LOCAL_AGENT_BENCHMARK=true`, refused on Vercel), per-entry reasoning effort and Google thinking level passed through to the Gateway, and 22 runtime-verified pricing pins in the snapshot.

- Harness: 24 ported cases (S1-H12, N13-N24) plus 12 new complex CRM cases (churn ranking, plan-my-week, committed vs weighted forecast, data-quality audit, schema design, audit-log month-over-month, grounded renewal email with an injected note, routine draft, ambiguous target, cross-workspace refusal, German two-turn), deterministic oracles against the database, two arm-blind LLM judges, a campaign ledger with a hard USD cap enforced before every episode, fresh fixture namespace per episode with the actor-turn-count invariant, and a report with pass rate, pass^3, cost per successful task, cache shares, TTFT and wall percentiles, exact sign tests with Holm correction, the statistical floor, and the deterministic selection rule from the plan.
- Runtime: `model-catalog.ts` gains `reasoningEffort`/`thinkingLevel` on entries and a guarded `bench:` overlay; `agent-turn.ts` forwards them as `reasoning` and `providerOptions.vertex.thinkingConfig`; `refresh-agent-model-pricing.ts` pins every benchmark endpoint.
- Self-tests: the database suite seeds all 36 cases, refuses a reused namespace, catches a planted wrong answer, and checks the C27 oracle in both directions; the report test exercises the selection rule on synthetic artifacts.

## Context

Part of the Mate quality and cost programme (plan in the owner's session). Credits are a pass-through of provider cost, so the default model and reasoning level need a measured decision: pass^3, cost per successful task and latency per arm, on the real runtime and a real database. The v4 study (CUS-314) never measured the shipped Flash-Lite and its Gemini failures were harness artefacts; this harness fixes those lessons (fresh namespace per episode, uniformly-failing checks listed, cache-locality noted, floor stated).

## Validation

- `yarn typecheck`, `yarn lint` (0 warnings), `yarn vitest run tests/conventions` (699 tests), full `yarn test` with `RUN_DATABASE_TESTS=true` and the sandbox database, `yarn build`: all green locally on the sandbox worktree.
[...]

PR #173, opened 2026-09-13
feat(agent-chat): runtime v2 with tool-set routing, prompt v2 and small-model schemas

benjiwagner:
## Summary

Runtime v2 for the hosted assistant, flag-gated so the benchmark can A/B it against the shipped runtime on one model (`AGENT_RUNTIME_VARIANT=current|v2`, default `v2`; `AGENT_RUNTIME_DISABLE=<knob,...>` for ablations; flags travel in the workflow payload).

- **Tool-set routing.** Records, workspace, docs, custom-field, interface and support tools are always in the model's list; messaging, social, widgets, webhooks, routines and admin sets load from the request wording, the current page, the sets the conversation already used, or an explicit `load_toolset` call, applied through `prepareStep` `activeTools`. The ChatGPT deep-research `search`/`fetch` pair leaves the hosted catalog. A records question now ships 55 KB of tool definitions instead of 152 KB, and the conservative initial context drops from 170 KB to 73 KB.
- **Prompt v2.** Static rules first, the user, date and language line last (cacheable prefix); one approval protocol; CRM data invariants shared with the MCP server instructions; explicit verification, clarification and confidentiality rules; interface paragraph only on the chat surface; routine trigger guide trimmed to the fired event.
- **Approvals per owner decision.** Routine `pause` and `run_now` and mailbox thread moves run without approval, routine `delete` stays gated, and a hosted guard refuses a routine `create` that leaves `enabled` unstated so the assistant drafts unless asked to activate. Docs reads inherit the turn locale when the model omits it.
- **Tool results and schemas for small models.** Totals before items in `list_records`, `list_users` and `manage_webhooks` list; list defaults 25; `pageSize` a plain 1-100 integer rounded up to a supported size; `sortDescriptor` described; filter syntax derived from the operator enum (all 17 operators); per-action keys in every multiplexed tool's `action` description; shorter social and Sales Navigator descriptions; `list_ui_targets` answers an unmatched query with an explicit miss and id prefixes.
- **Budget and continuation.** Worst-case pricing floor 2 bytes per provider token (balanced reserve 4 credits per round, 16 ahead, was 8 and 32); the `prepareStep` budget check includes the system prompt; admission measures the provider-specific, surface-filtered catalog; the continuation checkpoint carries a bounded numeric digest (total, page, item count, sums) of successful reads; `caching: "auto"` on the Gateway request; a turn settled from modelled cost raises a workflow warning.
- **Activity copy.** `manage_routines` gets its own kinds and consequences in five locales; approval cards show a dedicated label instead of the running label.

## Context

Owner decisions from the Mate alignment audit (see the plan in the owner's session): only MCP tools and navigate/highlight/tour, approvals as listed above, and a runtime that lets a small model reason within a bounded context. Stacked on #171 (interface surface) which is stacked on #170 (benchmark harness). CI workflows trigger only for pull requests targeting `main`, so this PR shows checks once its base merges and GitHub retargets it; the full local gate set is the evidence until then.

## Validation

- `yarn typecheck`, `yarn lint` (0 warnings, via the pre-commit hook on every commit), `yarn vitest run tests/conventions`, full `yarn test` with `RUN_DATABASE_TESTS=true` on the sandbox database, `yarn build`: green locally.
- `scripts/agent-benchmark/measure-context.ts` on this tree: records question 152,574 B full catalog, 55,203 B routed (28 of 51 tools); prompt v2 9,437 B; reservations balanced 4/16 credits, fast 1/4.
- New tests: toolset routing (lexicon EN/DE, page route, prior activities, load_toolset, malformed calls), runtime flags, prompt v2 (cacheable prefix, rules, surface, trimmed trigger guide), hosted guard and locale injection, page-size rounding and complete filter operators, continuation digest (numbers only, never text or ids), reservation floor, routines activity copy in all locales, approval label split; gating snapshot updated to the owner's decisions.
- Not run here: the paid `agent:eval` and the benchmark A/B (the sandbox has no `AI_GATEWAY_API_KEY`).

## Impact and rollback

Behavioural changes for every hosted turn: fewer tools per round, a new prompt, the approval changes above, and lower reservations. `AGENT_RUNTIME_VARIANT=current` restores the previous catalog, prompt and checkpoint shape without a deploy of old code; the approval policy, schema fixes and reservation floor are not flag-gated (they are corrections). Revert the branch to roll back entirely.


benjiwagner:
Superseded by pull request create failed: GraphQL: Head sha can't be blank, Base sha can't be blank, No commits between main and feat/mate-alignment, Head ref must be a branch (createPullRequest), which carries this branch's commits unchanged as one pull request against main.

benjiwagner:
Correction: the superseding pull request is https://github.com/customermates/customermates/pull/179 (the earlier comment captured an error message instead of the link).

PR #178, opened 2026-09-13
feat(billing): audit the connected accounts a billing change removes

benjiwagner:
## Summary

A connected account destroyed for billing reasons left no trace anywhere. `DeleteAccountForBillingService` calls Unipile's `removeAccount`, which is irreversible at the provider, then soft-deletes the local row to `status: deleted`. `ConnectedAccount` has no cause, no actor and no `deletedAt`, and the service published nothing, so nothing in the product could tell a workspace owner why their accounts disappeared.

The same account deleted by its owner **does** publish `CONNECTED_ACCOUNT_DELETED` (`ee/messaging/connect/delete-connected-account.interactor.ts:39-46`). This closes that asymmetry at the single choke point every billing-driven removal passes through.

## Context

Four causes reach `deleteForBillingOrThrow`, and they are now distinguishable rather than merged:

| Cause | Caller | `removalReason` |
| --- | --- | --- |
| Plan change detected against the provider | `delete-accounts-for-plan.interactor.ts` | `planDowngrade` |
| Trial ended more than 24h ago | `delete-connected-accounts-for-expired-trials.interactor.ts` | `trialExpired` |
| Subscription unpaid or expired past the grace window | same interactor | `subscriptionLapsed` |
| Owner deactivated | `delete-connected-accounts-for-inactive-owners.interactor.ts` | `ownerInactive` |

That interactor previously merged its two id lists into one `Set`. It now walks them as two labelled groups while preserving the de-duplication, and an id in both lists is deleted once as `trialExpired`.

**Attribution.** The row is a system publish carrying `{ systemCompanyId: account.companyId, systemUserId: account.userId }`, the pattern already used by `send-legal-document-notices.interactor.ts:163` and `process-account-add-webhook.interactor.ts:94`. Both columns are non-null on `ConnectedAccount`. The alternative was no row at all: `AuditLog.userId` is non-nullable and `features/event/event.service.ts:114` drops any event whose actor resolves to null. The account owner did not perform this deletion, which is exactly why `removalReason` is on the payload: without it the row would be indistinguishable from a deletion they made themselves, which would be worse than no row.

**No new `DomainEvent`.** `CONNECTED_ACCOUNT_DELETED` already exists, so none of the enum gates fire, the entity-name extractor already handles it, and `connected_account.deleted` is absent from `WEBHOOK_EVENTS`, so there is no webhook delivery and no routine trigger.

### Deliberately not included

The `Subscription` row itself is not audited, and that remains a real gap after this change. Recorded here so the reasoning and the exact site are not lost.

What this PR now captures: a plan change, trial expiry or lapse **that actually removed accounts**, per account, with its cause. What is still unrecorded: a plan change that removed nothing, every status transition (`active` to `past_due` to `unpaid` to `expired` — the one that deactivates every user and disables their routines after the three-day grace window), seat and quantity changes, and the previous plan, which is overwritten in place by each upsert and so exists nowhere queryable. `AgentUsageEvent.planSnapshot` is the only durable trace and no code path surfaces it.

It does not belong in `AuditLog`: `AuditLog.userId` is non-nullable so the webhook path — renewal, dunning, cancellation, expiry, i.e. most billing events — could never write a row, and `AuditLog` cascades on user and company deletion, which contradicts the separate retention the DPA declares for billing records.

The shape that does work is a small dedicated table with a nullable actor and no tenant cascade, holding previous and next plan/status/quantity plus a source discriminator. The code is ready for it: `ee/subscription/subscription.service.ts:128` reads `existing` immediately before `:131` writes the new state, so before and after are both in hand at a single choke point that every webhook and every Refresh passes through. Registration (`features/user/prisma-user.repository.ts:348`) writes the initial row; the three operator-console writers already keep their own `previous`/`next` trail in `OperatorAuditEvent`.

Separately and unrelated to auditing: `RefreshSubscriptionInteractor` is gated `@TenantInteractor({ resource: Resource.company, action: Action.readOwn })` yet overwrites plan, status and quantity and can trigger permanent deletion of connected accounts. A read-scoped permission performing a destructive write looks like an authorization bug rather than a design choice.

## Validation

Run in an isolated worktree against a disposable local PostgreSQL 17 container (`yarn db:provision`, port 26654), `APP_MODE=cloud`, synthetic seed applied.

- `yarn typecheck` — pass
- `yarn vitest run tests/conventions` — 85 files, 699 tests pass
- `RUN_DATABASE_TESTS=true npx vitest run --maxWorkers=3` — **728 files passed, 1 skipped; 6774 tests passed, 3 skipped; exit 0**
- `yarn build` — pass
- `yarn openapi:generate && yarn raw-docs:generate` — tree stays clean, no spec drift

### Exercised in the running app

Drove a real downgrade of the seeded workspace from `pro` through `DeleteAccountsForPlanInteractor` against the dev database, with only the external Unipile and email calls stubbed. It removed the five newest accounts and kept the oldest, which is correct for one included account per user, and wrote five audit rows attributed to the account owner with `removalReason: "planDowngrade"`. Before this change that was zero rows.

Then checked the surfaces: the five removals appear in Company → Audit logs as "Account Disconnected" with the account name and actor, and the dashboard activity timeline renders the expanded detail in German as `Anzeigename` / `Anbieter` / `Entfernungsgrund: Tarif herabgestuft`.

### Tests

- `ee/messaging/connect/__tests__/billing-account-removal-audit.database.test.ts` — 4 tests through DI against real PostgreSQL: a downgrade records every account it removed with owner and cause; only the overage is removed and only that is audited; a plan that still covers everyone writes nothing; a repeated downgrade neither re-deletes nor re-audits.
- `ee/messaging/connect/__tests__/delete-account-for-billing.service.test.ts` — extended with the exact publish shape, the already-deleted skip publishing nothing, a failed provider call leaving no row, and all four reasons surviving into the payload.

Proven to discriminate: removing the publish made the database test fail, and restoring it made it pass.

The three changed `toHaveBeenCalledWith` assertions in existing tests are **stronger** than before, not merely unbroken: each now pins the cause as well as the id, and the de-duplication test now pins which cause wins.

## Impact and rollback

User-visible: billing-driven account removals now appear in the company audit log and the activity timeline, localized in all five application locales including the reason value. Nothing else changes; the deletion behaviour itself is untouched.

`deleteForBillingOrThrow` gains a required second argument. It has three callers, all in this diff.

**Known limitation.** The publish is the last step, after the irreversible provider call and the local soft-delete. If the audit insert itself fails, the account is gone and no row is written, and the already-deleted early return means the next sweep skips it. That ordering is still the best of the three available: publishing earlier would risk a row for an account that is not actually gone. The window is a single `auditLog.create` with no fan-out, and the equivalent pre-existing window (a throw between the provider call and the local mark) is strictly larger.

Rollback: revert the commit. No migration, no backfill, no environment change. Rows already written remain readable, though their reason value would fall back to the raw code once the locale entries are gone.


benjiwagner:

PR #180, opened 2026-09-13
fix(billing): require company update to refresh a subscription

benjiwagner:
## Summary

`RefreshSubscriptionInteractor` was gated `@TenantInteractor({ resource: Resource.company, action: Action.readOwn })` — a **read** permission — while it overwrites the local `Subscription` row (plan, status, quantity, trial dates) from the provider and, when the synced plan differs from the stored one, invokes `DeleteAccountsForPlanInteractor`, which calls Unipile `removeAccount` and **permanently destroys other users' connected accounts**.

Every role in the product satisfied that gate. `features/role/prisma-role.repository.ts:111,121-124` grants `readOwn` and `readAll` on any manage-only resource regardless of the toggle, so `company: canManage = no` still passed. The seeded "Sales Manager" and "Customer Success" roles hold exactly those. It was not a gate.

The UI matched: `showRefresh` checked only `plan !== enterprise && status !== trial`, with no permission test, so the button was rendered for those members.

## Context

`Action.update` is the correct permission and makes the `company` resource uniform: reads at `readOwn`, writes at `update`. It matches `CreateCheckoutSessionInteractor` (`create-checkout-session.interactor.ts:32`), `UpdateCompanySettingsInteractor` (`update-company-settings.interactor.ts:43`), and the billing authority `features/auth/subscription-recovery.ts` already defined independently.

The UI gate uses `can(Resource.company, Action.update)`, deliberately **not** `canManage`. `canManage` requires create **and** update **and** delete (`user.store.ts:59-62`), so it would hide the control from users the server authorizes. System-role admins are unaffected: `hasPermission` short-circuits on `isSystemRole` on both sides, which matters because the seeded Admin role carries `permissions: []`.

### The same hole on the read side

`GetSubscriptionInteractor` is `readOwn` and returns the Lemon Squeezy `customerPortalUrl`. `app/layout.tsx` loads it on every authenticated page and ships it into the client store, so every member received a billing-management link in the RSC payload, hidden only by a client-side check in `SubscribeManageButton`. That is the same pattern this PR fixes for Refresh, so it is fixed here too: the link is withheld unless the caller can act on billing, which also avoids a provider HTTP call whose result they could not use. Plan, status and seats are still returned, because reading the subscription is legitimately a read permission.

The predicate now has one definition, `canManageBilling(user)` in `features/auth/subscription-recovery.ts`, used by both the recovery path and the subscription getter.

### Not changed

`SubscribeManageButton` still uses the stricter `canManage`. It is a presentation choice over data the server now withholds anyway, and changing it is outside this fix.

## Validation

Run in an isolated worktree against a disposable local PostgreSQL 17 container, `APP_MODE=cloud`, synthetic seed applied.

- `yarn typecheck` — pass
- `RUN_DATABASE_TESTS=true npx vitest run --maxWorkers=3` — **729 files passed, 1 skipped; 6771 tests passed, 3 skipped; exit 0**
- `yarn build` — pass
- `yarn openapi:generate && yarn raw-docs:generate` — no drift

### Proven at runtime, before and after

The server action was forged with each user's real session cookie against the running app, because hiding a button proves nothing about the gate. `Next-Action: 001e4dad…` posted to `/en/company/subscription`:

| gate | read-only member (Sales Manager) | admin |
| --- | --- | --- |
| old `readOwn` | **got past authorization** into billing logic, reaching `Subscription does not have a LemonSqueezy ID` | got past |
| new `update` | `ForbiddenError: Access denied. Required permissions: update on company`, `code: permissionDenied` | still gets past |

In the seeded workspace the run then failed on a missing Lemon Squeezy id. In any real workspace that has one, the member's request would have continued into `updateSubscriptionOrThrow` and, on a plan difference, into the account deletion.

### Tests

- `refresh-subscription.interactor.test.ts` — a read-only member is refused; a member with `company:update` is allowed. Both fail if the decorator is reverted.
- `get-subscription.interactor.test.ts` (new) — the portal link is withheld from a read-only member and the provider is not called; plan and seats are still returned to them; a billing manager and a system admin both receive it. The first fails if the new check is removed.

## Impact and rollback

A member whose role lacks `company: canManage` can no longer press Refresh or receive the billing portal link. That is the intent. Nobody who could legitimately manage billing loses anything: system admins bypass, and `canManage = yes` grants `update`. The subscription-expired recovery flow is unaffected — it calls `createCheckoutSessionAction`, already `{company, update}`, and `resolveSubscriptionRecoveryPath` already routed users without `company:update` to `administratorRequired`.

Rollback: revert the commit. No migration, no data change.


benjiwagner:

PR #181, opened 2026-09-13
feat(audit): record routine and billing-driven changes, and gate subscription refresh

benjiwagner:
## Summary

Consolidates #172, #178 and #180 into one branch. Three related gaps in the same area: what the audit log records about routines, what it records when billing destroys a connected account, and who is allowed to trigger that destruction.

| | Gap | Now |
| --- | --- | --- |
| **Routines** | No `ROUTINE_*` event existed; no routine create, edit, delete or admin pause left any trace | `routine.created` / `updated` / `deleted` from every user-reachable write path |
| **Billing removals** | An account destroyed at Unipile for billing reasons left no cause, no actor, no `deletedAt` | `connected_account.deleted` with the actual cause, from the single choke point |
| **Refresh gate** | `readOwn` let any member overwrite billing state and permanently delete other people's accounts | requires `company: update`, and the portal link is no longer handed to read-only members |

### Documentation alignment (commit 7)

The root README named audit logging four times, three defensively, which read as though the licence were fencing it off when it does the opposite. It now says it twice, neutrally. The Community Build enumeration stays where it is authoritative, in `ee/LICENSE.md:38` and `ee/README.md:12`, rather than being restated (and subtly reworded) in the root README.

Along the way this corrected claims the code contradicts:

- An unchanged company-settings or member save **does** write an audit entry. `isNoOpUpdate` only skips a publish when the payload carries a `changes` key, and `COMPANY_UPDATED`/`USER_UPDATED` carry a snapshot without one. My own earlier wording in this PR said the opposite.
- Inbox events are denylisted, not unpublished, so "parts of the product publish none" was wrong about its own first bullet.
- The audit screen has no export action, but the same rows are readable through the activity timeline, its REST endpoint and `get_activities`. A flat "there is no export" overstated it.
- One API key per client gives **no** audit granularity: `AuditLog` has no key column and the row names the owning user.
- UI labels corrected against shipped i18n: Record changes, Account & calendar, Selection unavailable, element id.
- Two adjacent lines the new audit section contradicted within the same file: "no admin panel that bypasses scope" (the cloud operator console does, under allowlist and its own log) and the roles line that omitted six resources including `auditLog`.

Every edit was verified against code and adversarially re-checked before applying. English and German move together.

**Deliberately not changed:** `content/legal/{en,de}/privacy.mdx`. Its cascade-deletion clause presupposes no production company-deletion feature exists, but the cloud operator console does hard-delete a company. That is a live GDPR notice, so the correction is yours to approve rather than mine to apply.

## Context

Scope is this request plus this pull request; no Linear issue. Three commits, one per concern, so they stay reviewable separately.

**Routines.** `EventService.publish` is the only writer of `AuditLog`, and no routine interactor imported it. The three events are published from `UpsertRoutineInteractor`, `PauseRoutineInteractor` and `DeleteRoutineInteractor`, following the role interactors. The UI server actions, the `manage_routines` MCP tool and the MCP-over-HTTP endpoint all route through those, so one change covers every path. The events map to `null` in `AUDIT_EVENT_ENTITY_TYPE` (routines have no `EntityType`), matching role, webhook and custom column. Ten routine change-fields that resolved in neither label catalog are now localized, so the activity timeline stops rendering humanized English in every language.

**Billing removals.** The same account deleted by its owner already published `CONNECTED_ACCOUNT_DELETED`; the billing-driven path published nothing. It now does, attributed with `systemCompanyId`/`systemUserId` from the account — the pattern the legal-notice cron uses — and carries which of the four causes applied: plan downgrade, ended trial, lapsed subscription, deactivated owner. No new `DomainEvent`, so no enum gate fires and there is no webhook or routine fan-out.

**Refresh gate.** `RefreshSubscriptionInteractor` was gated `readOwn` while overwriting plan, status, quantity and trial dates and, on a plan difference, invoking the account deletion. Every role in the product satisfied that gate, and the button was rendered for them. It now requires `Action.update`, matching `CreateCheckoutSessionInteractor` and the billing authority `features/auth/subscription-recovery.ts` already defined. The same hole on the read side is closed too: `GetSubscriptionInteractor` was handing the Lemon Squeezy portal link to every member at `readOwn`, hidden only client-side.

### Deliberately not included

The `Subscription` row itself is not audited. `AuditLog.userId` is non-nullable so the webhook path — renewal, dunning, cancellation, expiry — could never write a row, and `AuditLog` cascades on user and company deletion, contradicting the separate retention the DPA declares for billing records. The reasoning and the exact site for a future `SubscriptionChange` table (`ee/subscription/subscription.service.ts:128`, where before and after are both in hand) are recorded in #178.

## Validation

Isolated worktree, disposable PostgreSQL 17, `APP_MODE=cloud`, synthetic seed.

- `yarn typecheck` — pass
- `RUN_DATABASE_TESTS=true npx vitest run --maxWorkers=3` — **733 files passed, 1 skipped; 6797 tests passed, 3 skipped; exit 0**
- `yarn build` — pass
- `yarn openapi:generate && yarn raw-docs:generate` — no drift

**Merge integrity.** Each branch was combined with `git merge --squash` (a real three-way merge, not a commit replay). Every line each branch added to the eight shared files was then asserted present in the combined tree: **0 missing across 16 file/branch pairs**. The only conflict was two parallel import lines, resolved by keeping both.

### Exercised end to end on one pristine database

Re-provisioned and re-seeded from scratch: 161 audit rows, **zero** of the event types this branch adds. Then, through the real UI as a seeded admin: renamed a routine, deleted one, created one, paused another owner's routine. Then a real `pro` downgrade through `DeleteAccountsForPlanInteractor` with only Unipile and email stubbed. Then the refresh action forged with each user's real session cookie.

Result: **170 rows — 4 routine, 5 removal**, and:

| | read-only member (Sales Manager) | admin |
| --- | --- | --- |
| refresh action | `ForbiddenError: Access denied. Required permissions: update on company` | past the gate |

The downgrade removed the five newest accounts and kept the oldest, correct for one included account per user, each with `removalReason: planDowngrade`. All nine rows render in Company → Audit logs, verified in German: `Routine erstellt / aktualisiert / gelöscht` and `Konto getrennt`, with the activity timeline expanding to `Entfernungsgrund: Tarif herabgestuft`.

Every new behaviour was proven to discriminate by removing it and watching the relevant test fail.

## Impact and rollback

Routine and billing-removal changes now appear in the audit log and activity timeline, localized in all five locales. A member whose role lacks `company: canManage` can no longer refresh the subscription or receive the billing portal link; that is the intent, and nobody who could legitimately manage billing loses anything, since system admins bypass and `canManage = yes` grants `update`.

`deleteForBillingOrThrow` gains a required second argument (three callers, all in this diff). `AuditLog.eventData` stores the full routine DTO including the prompt, matching role and custom column, and `AuditLog` has no retention path.

Known limitation, carried from #178: the removal event publishes after the irreversible provider call, so a failed audit insert leaves the account gone with no row, and the already-deleted guard makes the next sweep skip it. That ordering is still the best available; publishing earlier would risk a row for an account that is not gone.

Rollback: revert the commits. No migration, no backfill, no environment change.

Supersedes #172, #178 and #180.


ee/subscription/get-subscription.interactor.ts
```diff
@@ -62,7 +63,11 @@ export class GetSubscriptionInteractor extends AuthenticatedInteractor<void, Sub
 
     let customerPortalUrl: string | null = null;
 
-    if (subscription.lemonSqueezyId && subscription.plan !== SubscriptionPlanEnum.enterprise) {
+    if (
+      canManageBilling(this.user) &&
```
benjiwagner: this checks we have added in this PR seem to be duplicate to the decorator on the interactors for the permission. doublecheck the codebase and also the changeset

benjiwagner: You were right that something was duplicated, though not the decorator.

The decorator gates *entry* to this interactor on `company: readOwn`. The portal link needs the stricter `company: update`, and a class decorator can't express that because it guards the whole call, not one field of the response. So the two checks answer different questions and neither is redundant.

What I did duplicate is the permission *mechanism*. The `canManageBilling` helper I added was a fourth hand-written copy of "system role, or permissions contains resource+action" — that logic already lives in `tenant-interactor.decorator.ts:50-58`, `UserService.hasPermissionForUser` and `BaseRepository.hasPermission`.

Fixed in e47751dc: the interactor now asks `UserService.hasPermission(company, update)`, the same way `DeleteCustomColumnInteractor` does for its per-entity check, and I reverted the extraction so `subscription-recovery.ts` is byte-identical to main again. This PR now adds no permission logic of its own.

I checked the rest of the changeset for the same smell. The only other permission expression is `subscription-view.tsx` using `can(company, update)` to hide the Refresh button — that mirrors the server gate deliberately so the UI doesn't offer a control the server refuses, which is the existing pattern next door in `SubscribeManageButton`.

Full suite 733 files / 6797 tests green, and the portal test still fails if the check is removed.

benjiwagner: Swept the complete changeset for the same class of mistake, four lenses (production helpers, UI/i18n, tests, wiring) with each finding adversarially verified. 16 candidates, 14 refuted, 2 real — both now fixed in 475cc68b.

**Permissions:** only 3 production lines in the whole diff touch permissions, and each uses a canonical mechanism (`@TenantInteractor`, `UserService.hasPermission`, `userStore.can`). No hand-rolled copies remain.

**Real one 1 — a user-visible defect I introduced.** `RoutineDto` carries both `ownerUserId` and the hydrated `owner` object, and the snapshot expander filters neither, so every `routine.created` / `routine.deleted` entry rendered **two rows both labelled "Owner"**, one a bare UUID. Confirmed against real rows in the database. `IGNORED_FIELDS` already handles this exact shape for `roleId`, so `ownerUserId` joins it and the label I added is dropped from all five bundles plus the allowlist. Net: less code than before.

**Real one 2.** Three tests in the audit suite re-staged refusal scenarios the ownership suite already owns, adding only "nothing was published". Moved that assertion onto the tests that stage the scenario, deleted the copies. Proven still to discriminate: making delete publish on its conflict path fails the ownership test.

**Deliberately not changed:** the event-service stub appears in 4 test files, but there is no shared helper in `tests/helpers/` and 6+ untouched suites (role, webhook, custom-column, tasks, organizations) define it locally the same way — and `conventions.md` prefers explicit local code over abstractions that only reduce repetition. `account-removal-reason.ts` follows `routine-schedule-preset.ts`; the emit shapes mirror the role and connected-account publishers exactly.

733 files / 6794 tests green, lint and build clean.

ee/subscription/refresh-subscription.interactor.ts
```diff
@@ -10,7 +10,7 @@ export abstract class RefreshSubscriptionRepo {
   abstract getSubscriptionOrThrow(): Promise<{ lemonSqueezyId: string | null; plan: SubscriptionPlan }>;
 }
 
-@TenantInteractor({ resource: Resource.company, action: Action.readOwn })
+@TenantInteractor({ resource: Resource.company, action: Action.update })
```
benjiwagner, line 13: why is this now an update action?

ee/subscription/get-subscription.interactor.ts
```diff
@@ -62,7 +64,11 @@ export class GetSubscriptionInteractor extends AuthenticatedInteractor<void, Sub
 
     let customerPortalUrl: string | null = null;
 
-    if (subscription.lemonSqueezyId && subscription.plan !== SubscriptionPlanEnum.enterprise) {
+    const mayManageBilling =
+      subscription.plan !== SubscriptionPlanEnum.enterprise &&
+      (await this.userService.hasPermission(Resource.company, Action.update));
```
benjiwagner, line 69: Isn't this duplicate because we already have the action and the respective interactor?

README.md
```diff
@@ -49,7 +49,7 @@ Docs entry points:
 - n8n workflows and automation support
 - MCP support for agent tooling and structured tool calling
 - Unified inbox for email, LinkedIn, WhatsApp, Instagram, and Telegram (Cloud, from the Pro plan)
-- Audit logging on every plan, self-hosted included; Single Sign-On and Whitelabeling on the Enterprise plan (self-hosted Enterprise deployments by agreement)
+- Audit logging; Single Sign-On and Whitelabeling on the Enterprise plan (self-hosted Enterprise deployments by agreement)
```
benjiwagner: Isnt auditlogging on every plan?

README.md
```diff
@@ -164,10 +164,10 @@ Start here: [customermates.com/docs](https://customermates.com/docs)
 
 Customermates uses an open-core licensing model.
 
-The Community Edition—all first-party code outside `ee/`, including `features/audit-log/`—is licensed under [AGPL-3.0-only](./LICENSE). Audit logging is therefore free for production self-hosting.
+The Community Edition (all first-party code outside `ee/`, including `features/audit-log/`) is licensed under [AGPL-3.0-only](./LICENSE).
```
benjiwagner: why is auditlog mentioned here specifically again. Auditlogging should be part of the normal apgl licence like the rest

README.md
benjiwagner: Right, and my edit made that ambiguous. Fixed in 656792db.

I had joined it to the Enterprise clause with a semicolon:

> - Audit logging; Single Sign-On and Whitelabeling on the Enterprise plan (...)

which reads as though audit logging were on the Enterprise plan too. It is its own bullet now, with no plan qualifier, like `Role-based access control for teams`. The comparison table two sections down still carries the per-edition answer: `| Audit log | Included | Included |`.

To confirm the underlying fact: it is on every plan and on self-hosted. There is no audit entitlement in `core/commercial/plan-catalog.ts`, `ee/subscription/entitlements.ts` has none either, the nav entry at `app/components/navigation/workspace-sections.ts:29` is not `cloudOnly`, and the page is gated only by the `auditLog` role permission.

README.md
benjiwagner: Agreed, and done in 656792db.

The line now reads:

> The Community Edition (all first-party code outside `ee/`) is licensed under [AGPL-3.0-only](./LICENSE).

`features/audit-log/` is outside `ee/`, so it was already covered. Naming it was exactly what made the licence section look like audit logging is a special case that needs defending, which is what prompted this whole thread.

One thing that needed changing with it: `tests/conventions/open-core-license.test.ts` pinned the literal string "including `features/audit-log/`" in the README, so it would have gone red. I dropped that assertion and left the two structural ones, which are the real guard and do not depend on wording:

```ts
expect(existsSync(join(REPO_ROOT, "features/audit-log/prisma-audit-log.repository.ts"))).toBe(true);
expect(existsSync(join(REPO_ROOT, "ee/audit-log"))).toBe(false);
```

Location is what determines the licence, so asserting the location is strictly better than asserting a sentence about it.

Worth flagging separately: `ee/LICENSE.md:18` still ends "Audit logging is not an Enterprise Feature." I left that deliberately. It is not an exclusion, it is the opposite: the Enterprise Feature definition has a catch-all for anything "expressly identified as Enterprise or Cloud-only in the Applicable Documentation", and Applicable Documentation is these two READMEs. That sentence is what stops a future pricing-table edit silently pulling audit logging into the commercial set. Happy to remove it too if you would rather, but it is doing real work where it is.

ee/subscription/refresh-subscription.interactor.ts
benjiwagner: Because it writes. `Action.readOwn` was wrong for it.

`RefreshSubscriptionInteractor.invoke` calls `subscriptionService.updateSubscriptionOrThrow`, which writes the `Subscription` row, and when the plan came back different it then calls `deleteAccountsForPlan`, which **deletes connected accounts** (`refresh-subscription.interactor.ts:33-36`). Under the old gate any member with plain read access to the company could trigger that.

I proved it at runtime rather than by reading, since the gate is a decorator and easy to reason about wrongly. I extracted the server action id from the client bundle and POSTed it with two real session cookies:

- read-only member, old `readOwn` gate: request passes the guard and reaches the billing logic
- same member, new `update` gate: `ForbiddenError: Access denied. Required permissions: update on company`
- billing manager, new gate: still succeeds

`update` on `company` is the permission that already guards company mutations elsewhere, so this puts it with its peers rather than inventing a new one.

ee/subscription/get-subscription.interactor.ts
benjiwagner: Not duplicate, and it is a fair thing to ask twice because the two checks look identical at a glance.

The decorator on this interactor gates **entry** on `company: readOwn`, because any member may read the subscription to see the plan, seat count and trial end. The in-body check gates **one field of the response**, `customerPortalUrl`, on the stricter `company: update`, because that link opens LemonSqueezy where someone can change the card, the plan or cancel.

A class decorator cannot express that. It guards the whole call, so raising it to `update` would hide plan and seat count from ordinary members, and leaving it at `readOwn` would hand everyone the billing portal. Different questions, different scopes.

What genuinely was duplicated is the thing you caught the first time: I had hand-rolled a `canManageBilling` helper that reimplemented "system role, or permissions contains resource+action". That is gone. The line now calls `UserService.hasPermission(company, update)`, the same canonical mechanism `DeleteCustomColumnInteractor` uses for its per-entity check, so this PR adds no permission logic of its own.

If you would rather have no in-body check at all, the alternative is a separate `GetBillingPortalUrlInteractor` decorated `{company, update}`, with the client fetching it after the subscription. That is arguably cleaner, at the cost of a second round trip and a changed response shape. Say the word and I will do it, otherwise I will leave it as is.

The test pins the behaviour either way: removing the check fails `get-subscription.interactor.test.ts`, which asserts a read-only member gets `customerPortalUrl: null` and that LemonSqueezy is never called for them.

